PR #3, opened 2015-06-02
YML output fix

Zrubi:


marmarek:
Is this the final one?

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

Zrubi:
Really hope so :)

PR #6, opened 2015-08-30
Allow defining and using custom Storage types

kalkin:
I'm currently reimplementing my Qubes R2 lvm thin storage patchset on current head (in hope to get it upstream some day). I ran in to the following issue: There is no way to model: "this specific Xen based vm is backed up by LVM/Btrfs/Whatever" instead of files.  At the current moment Qubes is hardcoded to use the `QubesXenVmStorage` class on Linux and the `QubesWniVmStorage` on windows. 

### Proposal
I humbly propose saving the storage class responsible for the vm as one of the
attributes in the config file `/var/lib/qubes/qubes.xml`. (This is similar to
how I solved it for my lvm patcheset for  the R2 version and it proved to be a
reasonable solution.) 

### Implementation

The attribute `storage_class` saves which python class is responsible for storage of the specific vm. Given a class `o` the content of `storage_class` would be `o.__module__ + '.' + o.__class__.__name__ `(see `dump(o)` in `core/storage/__init__py`). 

By saving the full module name and class name it should be possible for third-party to provide own storage module which do not need to be repackaged with `core-admin`, so they could be just installed in the default python site-packages path. (see `load(string)` in in `core/storage/__init__py`

`default['storage_types']` are thought to be used as aliases for the default provided storage classes, i.e: when used from command line as argument

I also added a dummy storage and some test cases.

I think adding some way to have multiple storage engines/classes will greatly enhance Qubes.There were some ideas about a btrfs snapshot storage for vms on the ML, there are also zfs qubes users which could benefit from a zfs storage class and last but not least my lvm thin patchset saves me a ton of ssd disk space and time waisted waiting for qvm-create & qvm-clone.


marmarek:
I'll review/comment on it later, as our top priority for now is R3.0.
But generally something like this is a good idea. Probably needs some
careful implementation (for example to guard that template and its
children VMs use compatible storage classes).

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

kalkin:
@marmarek 

Thank you for your  fast response :smile: 

> Probably needs some careful implementation (for example to guard that template and its children VMs 
use compatible storage classes).

I think you can handle it by implementing in  _get_rootdev the following algorithm:`
```python
if self.vm.template.storage_class != self.vm.storage_class:
   storage_template = self.vm.template.storage_class(self.vm)
   return storage_template._get_rootdev()    
```

This way you can have a TemplateVm based on storage engine A and an AppVm based on the storage engine B.

kalkin:
After a few days passed by i noticed a few flaws in the proposed design. There is no way to configure the storage or pass any parameters like VolumeGroup or ThinPool name. Also missing is a way to configure different pools using different storage adapters i.e: An lvm based storage which uses `/dev/vg1`, while the qubes-tests are run on `/dev/vg2/`

Such data could be read/stored in the `/etc/qubes/storage.conf` like this:
```INI
[default] ; pool name
type=xen ; the default xen storage
; class = qubes.storage.xen.QubesXenVmStorage ; class always overwrites type

[test-dummy]
type=dummy

[test-lvm]
class=qubes.storage.lvm.ThinStorage ; The storage adapter
vg=qubes_dom0                       ; The LVM VolumenGroup
thin_pool=pool00                    ; The Thin pool (do not confuse with qubes storage pool name!)


; To use our own storage adapter, you need just to specify the module path and
; class name
; [pool-b]
; class = foo.bar.MyStorage
; 

```

@marmarek when you have some free time have a look on the working draft [here](https://github.com/kalkin/qubes-core-admin/tree/storage-pools)

kalkin:
So whats the state on this? I really would like to finish up this soon. :)

marmarek:
Generally it looks good. But this is only one part of the picture. The missing parts:
* handling AppVMs stored in different poll than its template (the code from comment, or simply preventing such situations)
* setting the storage poll (qvm-create option?)
* changing the storage poll (I have no problem with not supporting this operation, but needs to be explicitly prevented)
* how would `qvm-clone` work when the target storage poll is different than the source? Same as above - can be not supported, but needs explicit note.

@woju can you also take a look at it?

kalkin:
@marmarek  Thank you for your response, sorry it took some time to respond, but
this going to be a big one.

You address the right missing parts, at the end you have to decide how flexible
and powerful solution you want. There is a lot of what could be possible, but it
would introduce big changes to the storage layer (which IMHO are appropriate).

I tried before to be very conservative with my proposal, because I did not want
to overwhelm you. After our last "talk" here, I worked further on the storage and
lvm and tripped once again over how much custom logic is in the storage classes.
My LVM Patchset was just an ugly hack, where the `LvmStorage` inherited
from  `XenStorage` to avoid code duplication.  Also it had a lot of assumption
about how XenStorage and QubesVMs behave. There are a lot of

```python
if self.vm.is_template()
```
in the code .
So here is my proposal for the storage "api".

## Proposal
 We will be needing to save the pool_name and perhaps some specific pool data
 for each device separately in the `qubes.xml`.

* `get_pool(pool_name, vm)`
* `StoragePool.create_storage(size)`
* `StoragePool.clone()`
* `StoragePool.import(storage)` imports from another pool
* `Storage.remove()`
* `Storage.export()` receives data from another pool for populating storage
* `Storage.size()`
* `Storage.create_snapshot(commitable=[True|False])`
* `Storage.commit_snapshot()` can be used for TemplateVM to write back changes
* `Storage.remove_snapshot()`
* `Storage.serialize()` use to save data like path and pool name about the Storage in qubes.xml
* `Storage.deserialize()`
* `Storage.get_config()` get the libvirt xml data

## How would it work
### TemplateVM
```python
def create(root_pool_name, private_pool_name):
  self.template_img = get_pool(root_pool_name, vm).create_storage("20G")
  self.private_img = get_pool(private_pool_name, vm).create_storage("2G")
def start(volatile_pool_name):
  self.root_img = template_img.create_snapshot(commitable=True)
  self.volatile_img = get_pool(volatile_pool_name, vm).get_storage("2G")
  self.update_libvirt_domain()
def shutdown():
  self.root_img.commit_snapshot()
  self.volatile_img.remove()
def clone(vm, root_pool_name, private_pool_name):
  if root_pool_name == vm.template_img.pool_name:
    # Use the same pool
    self.template_img = get_pool(root_pool_name, self).clone(vm.template_img)
  else:
    # Use different pools so data needs to be imported
    root_pool = get_pool(root_pool_name, vm)
    self.template_img = migrate_storage(self.template.template_img, root_pool)
  ...
```
### AppVm
```python
def create(private_pool_name):
  self.private_img = get_pool(private_pool_name, vm).clone(self.template.private_img)
def start(volatile_pool_name):
  self.root_img = self.template.template_img.create_snapshot(commitabe=False)
  self.volatile_img = get_pool(volatile_pool_name, vm).get_swap("2G")
  self.update_libvirt_domain()
def shutdown():
  self.root_img.remove_snapshot()
  self.volatile_img.remove()
```

## The Missing Parts

Allow me to change the order of your question, so somethings will be clearer
> setting the storage poll (qvm-create option?)

```shell
qvm-create -p root:POOL_NAME -p private:POOL_NAME[:MORE:STORAGE:SPECIFIC:OPTIONS] -p swap:POOL_NAME
```

> handling AppVMs stored in different poll than its template (the code from comment, or simply preventing such situations)


`Parent` is a template vm. Both of his devices are provided by the storage pool `foo`. You do this:
```shell
qvm-create Child -t Parent -p private:bar -p swap:bar
```

 Child.root_img is
 `self.template.template_img.create_snapshot(commitabe=False)`. So the `Parent`
 pool is used for the root snapshot, while the the private storage should be
 handled by the pool `bar`.  Something like this:


```python
def migrate_storage(fromStorage, toPool):
  size = fromStorage.size()
  toStorage = toPool.create_storage(size)
  return toPool.import(toStorage.export())

```

The import/export functions would be some equivalent of doing:
```shell
mount /dev/mapper/foo /mnt/foo
mount /dev/mapper/bar /mnt/bar
cp -ar /mnt/foo /mnt/bar
umount /dev/mapper/foo /dev/mapper/bar
```

> how would qvm-clone work when the target storage poll is different than the
source? Same as above - can be not supported, but needs explicit note.

same as above

> changing the storage poll (I have no problem with not supporting this operation,
but needs to be explicitly prevented)

```
qvm-prefs MyAppVm -s private bar
```
Behind the scene the data would be migrate with the migrate_storage function.

## Benefits

* The Storage code should become smaller and leaner.
* All the storage manipulations would be driven by the VM class. This would allow to develop
  other VMs with different templating/storage logic without rewriting the whole
  storage layer.
* Simpler to test

## Things to be discussed
There're few things about which I'm not sure yet:

1. How the import/export should be implemented so it's nice to use it from
python, and it alos cleans up and unmounts everything after itself.  

2. At the moment there is some logic which depending if the vm is a TemplateVM
or an AppVm creates a root cow image inside the volatile image.  This is a
pretty specific hack for the file based storage solution, perhaps it would be
better to use volatile image only for swap and move the logic of creating a r/w
image for merging with the snapshot function of the `Storage`. I.e would LVM and
Btrfs not need any logic to have some write buffer in the volatile image.

3. Overkill?? :)

marmarek:
On Tue, Oct 06, 2015 at 04:15:40PM -0700, kalkin wrote:
> So here is my proposal for the storage "api".

General info: in R4.0 we're going to rewrite most of Qubes*Vm classes
code. The work on this is pretty advanced (see woju/qubes-core-admin,
core3-devel branch, "qubes" directory). All major changes should be based on that code (to
not duplicate the work). So this should be either reusable (possibly
with _minor_ adjustments), or totally based on "core3". I'm not going to
accept major change now, just to redo/undo it later in R4.0.

I'm not saying that this proposal breaks this assumption, but I'm also
not saying it doesn't. Currently storage classes in "core3" are similar
to those in R3.0.

> ## Proposal
>  We will be needing to save the pool_name and perhaps some specific pool data
>  for each device separately in the `qubes.xml`.
> 
> * `get_pool(pool_name, vm)`
> * `StoragePool.create_storage(size)`
> * `StoragePool.clone()`

`Storage.clone()`?

> * `StoragePool.import(storage)` imports from another pool
> * `Storage.remove()`
> * `Storage.export()` receives data from another pool for populating storage
> * `Storage.size()`

also:
* `Storage.resize()`?

> * `Storage.create_snapshot(commitable=[True|False])`
> * `Storage.commit_snapshot()` can be used for TemplateVM to write back changes
> * `Storage.remove_snapshot()`
> * `Storage.serialize()` use to save data like path and pool name about the Storage in qubes.xml
> * `Storage.deserialize()`

probably instead of `deserialize`, `Storage.serialize()` should be
simply usable for object constructor.

> * `Storage.get_config()` get the libvirt xml data

(...)

Generally the most common use case for template is:
1. start
2. shutdown (committing the changes)

AppVMs can be started at any point (either before starting template, or
after). Can be multiple of them

It should be optimized for this workflow. This means:
1. In most common use case, template changes will be committed. So
commit operation should be cheap (revert can be expensive).
2. Multiple AppVMs may be running at the same time. Having separate
snapshot (in LVM terminology) for each of them is waste of resources
(both time and space). Because every block changed by the template would be
written to _every_ snapshot (actually original block, not the changed
one). This is one of the reasons why we have two layers of dm-snapshot.

This may mean that we need to adjust that API, but may also mean that
the API is ok, just the LVM implementation should take appropriate steps
(don't map 1:1 create_snapshot -> lvcreate -s). I'm for adjusting the
API to not mislead the reader.

Regarding two layers of dm-snapshot, there is one more reason for it.
Untrusted Storage Domain. In short, we need access (in AppVM!) to
original root fs from the template (as template wrote it), to be able to
verify it. This isn't currently implemented, but we're probably going to
return to this idea (in the not so near future).

Details in original architecture specification:
https://www.qubes-os.org/attachment/wiki/QubesArchitecture/arch-spec-0.3.pdf

If you haven't seen this already, I recommend reading also
https://www.qubes-os.org/doc/TemplateImplementation/

> The import/export functions would be some equivalent of doing:
> ```shell
> mount /dev/mapper/foo /mnt/foo
> mount /dev/mapper/bar /mnt/bar
> cp -ar /mnt/foo /mnt/bar
> umount /dev/mapper/foo /dev/mapper/bar
> ```

Definitely not `mount`. Dom0 shouldn't parse any VM filesystem. But `dd`
probably is ok.

> ## Benefits
> 
> * The Storage code should become smaller and leaner.
> * All the storage manipulations would be driven by the VM class. This would allow to develop
>   other VMs with different templating/storage logic without rewriting the whole
>   storage layer.
> * Simpler to test

Makes sense.

> ## Things to be discussed
> There're few things about which I'm not sure yet:
> 
> 1. How the import/export should be implemented so it's nice to use it from
> python, and it alos cleans up and unmounts everything after itself.  
> 
> 2. At the moment there is some logic which depending if the vm is a TemplateVM
> or an AppVm creates a root cow image inside the volatile image.  This is a
> pretty specific hack for the file based storage solution, perhaps it would be
> better to use volatile image only for swap and move the logic of creating a r/w
> image for merging with the snapshot function of the `Storage`. I.e would LVM and
> Btrfs not need any logic to have some write buffer in the volatile image.

Above explained why it still makes sense to have two layers of
dm-snapshot (or equivalent). Also I don't think that `volatile.img`
should be logically renamed to `swap.img`. There are other systems, not
only Linux. Also some templates may want to use it for other things
(/tmp?). It would be better to treat `volatile.img` as a blackbox,
handled by VM. Yes, this means moving the current mkswap somewhere to
the VM (probably initramfs).

> 3. Overkill?? :)

We've evaluated migrating to libvirt storage pools (to take advantage of
already implemented volumes handling - mostly LVM). But it doesn't solve
all the problems, it isn't possible to replace all that storage classes
with just libvirt. Your solution seems to be much better in this
context. And one day we would be able to add LibvirtStoragePool if we
want to.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

kalkin:
> General info: in R4.0 we're going to rewrite most of Qubes*Vm classes
code. The work on this is pretty advanced (see woju/qubes-core-admin,
core3-devel branch, "qubes" directory). All major changes should be based on that code (to
not duplicate the work). So this should be either reusable (possibly
with _minor_ adjustments), or totally based on "core3". I'm not going to
accept major change now, just to redo/undo it later in R4.0.

Well this is almost the same reason i delayed the implementation for so long. I
wanted to talk this stuff through with you so I do not need to reimplement it
_again_. I have seen that woju does a rework. An VM classes overhaul is really
needed.

I think some changes may be merged into R3 (like the StoragePool logic), while
others could be implemented as part of R4. I would love to collaborate with woju
on merging and introducing this changes to R4. It mostly depends on when do you
target to relase R4 (After reading the VersionScheme I'm not really smarter, are
we talking end of year or next summer? Will be the next Release 3.1 or 4.0?) and
how long would R2 be still supported (assuming i use f21 templates). I plan to migrate to next Qubes Version when I have a working LVM storage solution, so my workflows wont degrade.

Another possibility would be partly introduce the changes like StoragePool to
3.0 while the whole storage logic would be implemented similar to my proposal,
but wrapped by the current Storage classes. This would keep the storage logic
for R3 inside the storage code without introducing big changes to the QubesVm
classes. This also would allow to battle test the proposed architecture.

> `Storage.clone()`?

Nop `StoragePool.clone()`. My gut is telling me it should be a responsibility of
the StoragePool not the Storage. But you can convince me otherwise :). 

> `Storage.resize()`?

Definitely. 

> probably instead of `deserialize`, `Storage.serialize()` should be simply
usable for object constructor.

I thought about using serialiaze/deserialize with the QubesVm config attributes
`['func']`  and `[save]`. So we can save any configuration data about the
storage used.

> 1. In most common use case, template changes will be committed. So commit
operation should be cheap (revert can be expensive). 

> 2. Multiple AppVMs may be running at the same time. Having separate snapshot
(in LVM terminology) for each of them is waste of resources (both time and
space). Because every block changed by the template would be written to
_every_ snapshot (actually original block, not the changed one). This is one of
the reasons why we have two layers of dm-snapshot.

No it is not. It's pretty cheap. Correct me if I am wrong, but I think you are
talking about the old LVM snapshots and not the Thin Pool snapshots. Or did I
misunderstand something?

> Regarding two layers of dm-snapshot, there is one more reason for it. Untrusted
Storage Domain. In short, we need access (in AppVM!) to original root fs from
the template (as template wrote it), to be able to verify it. This isn't
currently implemented, but we're probably going to return to this idea (in the
not so near future).

Well this would work with `verify(self.template.template_img)`. We could let the
StoragePool implement the `verify` function or the Storage itself.

> Above explained why it still makes sense to have two layers of
dm-snapshot (or equivalent). 

Yes it's a great solution for the file based storage, but snapshots in LVM using
Thin or with Btrfs are very cheap. I do not propose to kill this feature. I just
proposed not to have the root.cow of the AppVM inside the volatile image. But i
guess if you have a custom logic in your QubesVM class, which is used for
formating the image as swap , there is no reason, to not have some custom logic
in AppVM which puts the root.cow image inside the volatile.img


> Also I don't think that `volatile.img` should be logically renamed to
`swap.img`. There are other systems, not only Linux. Also some templates may
want to use it for other things (/tmp?). It would be better to treat
`volatile.img` as a blackbox, handled by VM. Yes, this means moving the current
mkswap somewhere to the VM (probably initramfs).

kalkin:
> Also I don't think that `volatile.img` should be logically renamed to
`swap.img`. There are other systems, not only Linux. Also some templates may
want to use it for other things (/tmp?). It would be better to treat
`volatile.img` as a blackbox, handled by VM. Yes, this means moving the current
mkswap somewhere to the VM (probably initramfs).

FULL ACK

woju:
Hi,

I've read through your proposal. First of all, thank you, it mostly makes sense.

The core3 effort does not touch storage scheme very much, but there are some minor corrections, such as moving all the functions that manipulate files to the class (only last week I moved `du` reimplementation to `qubes.storage` from old `qubes.qubesutils`) and dumbing down `dir_path` property. Those changes are probably not that hard to rebase.

You should also be aware of two long-term plans. First, nearer, is that in liveusb we have hacked together hybrid storage, which holds templatevms on the stick (or cd) and appvms in RAM (currently by mounting tmpfs on correct directory). Such configuration (or comparable) should remain a possibility.

Second, in the unspecified future we plan to move whole storage to separate domain and keep Xen block backends there. In theory that should not affect much of the storage, and because the plan is not that specific, there is no need to create API for it just now.

Some specific comments:
- API should provide two functions to get size of the image/storage, one to get size available to VM ("total disk size") and the second for getting size that is really taken from physical disk (taking into account sparse files, unallocated extents, compression, whatever). Currently they are called `get_*_sz` and `get_*_utilization`, however feel free to change the names, because the last one is very long.
- Maybe the property that specifies the pool should be set per-template (and all appvms should refer to their respective templates). That would be convenient way to disallow cross-pool dependencies. If user wanted some domain in different pool, the template can be cloned to that pool.
- With the move to setuptools, storage classes will be an `entry_point`, which means programmers will be able to provide their own classes without "registering" them or dropping .py files into our package. Just installing their package will be sufficient. See QubesOS/qubes-issues#1238.

marmarek:
On Wed, Oct 07, 2015 at 04:01:36AM -0700, Wojtek Porczyk wrote:
> - Maybe the property that specifies the pool should be set per-template (and all appvms should refer to their respective templates). That would be convenient way to disallow cross-pool dependencies. If user wanted some domain in different pool, the template can be cloned to that pool.

If implementing VM on different storage pool than its template is easy
(and apparently is), it wouldn't be necessary to block such
functionality.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

marmarek:
On Wed, Oct 07, 2015 at 03:10:27AM -0700, kalkin wrote:
> I think some changes may be merged into R3 (like the StoragePool logic), while
> others could be implemented as part of R4. I would love to collaborate with woju
> on merging and introducing this changes to R4. It mostly depends on when do you
> target to relase R4 (After reading the VersionScheme I'm not really smarter, are
> we talking end of year or next summer? Will be the next Release 3.1 or 4.0?) and
> how long would R2 be still supported (assuming i use f21 templates). I plan to migrate to next Qubes Version when I have a working LVM storage solution, so my workflows wont degrade.

R2 - no EOL date set yet, but at least to the end of the year.
next release - R3.1. Most features are complete, hopefully first rc at
the end of Oct/begging of Nov.
R4.0 - first rc at the end of the year (we hope...).

> Another possibility would be partly introduce the changes like StoragePool to
> 3.0 while the whole storage logic would be implemented similar to my proposal,
> but wrapped by the current Storage classes. 

No, since 3.0 is released, only bugfixes are accepted. New functionality
should go into 3.1 or 4.0.

> > `Storage.clone()`?
> 
> Nop `StoragePool.clone()`. My gut is telling me it should be a responsibility of
> the StoragePool not the Storage. But you can convince me otherwise :). 

It's storage volume you're cloning not the storage pool. In case of
files or btrfs I believe it would be enough to have just Storage object
to clone, not the whole pool. Probably also in case of LVM (Storage
object needs VG name anyway).
If some particular implementation needs access to StoragePool object
to perform clone operation, it can have such reference internally, I
don't care. But at the API level it seems more logical to have `clone()`
in Storage.

> > probably instead of `deserialize`, `Storage.serialize()` should be simply
> usable for object constructor.
> 
> I thought about using serialiaze/deserialize with the QubesVm config attributes
> `['func']`  and `[save]`. So we can save any configuration data about the
> storage used.

There would be no problem with creating some wrapper just for R3.1
('save'/'func'). But we should aim at clear design for 4.0.

> No it is not. It's pretty cheap. Correct me if I am wrong, but I think you are
> talking about the old LVM snapshots and not the Thin Pool snapshots. Or did I
> misunderstand something?

Ah, indeed, I'm talking about old snapshots.

> > Regarding two layers of dm-snapshot, there is one more reason for it. Untrusted
> Storage Domain. In short, we need access (in AppVM!) to original root fs from
> the template (as template wrote it), to be able to verify it. This isn't
> currently implemented, but we're probably going to return to this idea (in the
> not so near future).
> 
> Well this would work with `verify(self.template.template_img)`. We could let the
> StoragePool implement the `verify` function or the Storage itself.

No, it wont work. It's about _Untrusted_ Storage Domain. The VM needs to
be able to verify root fs, not storage domain itself, as it is ...
untrusted.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

kalkin:
> The core3 effort does not touch storage scheme very much, but there are some minor corrections, such as moving all the functions that manipulate files to the class (only last week I moved du reimplementation to qubes.storage from old qubes.qubesutils) and dumbing down dir_path property. Those changes are probably not that hard to rebase.

You are talking about your own branch for _core4_ not the main repo right? Or did i confuse something?

> You should also be aware of two long-term plans. First, nearer, is that in liveusb we have hacked together hybrid storage, which holds templatevms on the stick (or cd) and appvms in RAM (currently by mounting tmpfs on correct directory). Such configuration (or comparable) should remain a possibility.

Just to make sure i understand it right: The AppVM volatile.img is using a mounted tmpfs. So one of the disks of the AppVM (in this case the volatile.img) is backed by a tmpfs based StoragePool. This is how i would see it.

> Second, in the unspecified future we plan to move whole storage to separate domain and keep Xen block backends there. In theory that should not affect much of the storage, and because the plan is not that specific, there is no need to create API for it just now.

I think this would only benefit from storage refactoring.

> API should provide two functions to get size of the image/storage, one to get size available to VM ("total disk size") and the second for getting size that is really taken from physical disk (taking into account sparse files, unallocated extents, compression, whatever). Currently they are called get_*_sz and get_*_utilization, however feel free to change the names, because the last one is very long.

We need a new issue for such stuff where we would only specify the storage rewrite to be merged with @woju R4 branch. Should i open it? I do not want to step on any toes and also not sure if it should go in to the issues project or here (because pull requests and commit msg mentions would reference  this repo). 

> With the move to setuptools, storage classes will be an entry_point, which means programmers will be able to provide their own classes without "registering" them or dropping .py files into our package. Just installing their package will be sufficient. See QubesOS/qubes-issues#1238.

I'm completly on board with that. You can already create your own StoragePool implementations with the proposed patchset. See [above](https://github.com/QubesOS/qubes-core-admin/pull/6#issuecomment-137431790).

>> On Wed, Oct 07, 2015 at 04:01:36AM -0700, Wojtek Porczyk wrote:
>> - Maybe the property that specifies the pool should be set per-template (and all appvms should refer to their respective templates). That would be convenient way to disallow cross-pool dependencies. If user wanted some domain in different pool, the template can be cloned to that pool.

> If implementing VM on different storage pool than its template is easy
(and apparently is), it wouldn't be necessary to block such
functionality.

Yeah i am with @marmarek on that. This should not be an issue (i think).

>> Another possibility would be partly introduce the changes like StoragePool to
 3.0 while the whole storage logic would be implemented similar to my proposal,
 but wrapped by the current Storage classes.
> No, since 3.0 is released, only bugfixes are accepted. New functionality
should go into 3.1 or 4.0.

Yeah, right. I meant 3.1. 

Now back to the main topic: **This pull request**
I would propose the following steps to integrate the StoragePool patchset in R3.1.
1. I would rebase and retest on the current master(?) branch and get it hopefully merged with upstream.
2. I would extend if I may the QubesVm to save the used storage pool for each device. The changes should not be that big. 
3. Extend the `qvm-create` tools so you can pass StoragePool parameter for each device.

#### Open Questions
1. Where is the code handling liveusb? I think i would need to implement the tmpfs StoragePool to not break it.
2. The windows wni storage. Is it dead code? Or do you use in you own proprietary stuff? AFAIK it's not used windows based vms?

marmarek:
On Thu, Oct 08, 2015 at 01:23:44PM -0700, kalkin wrote:
> > The core3 effort does not touch storage scheme very much, but there are some minor corrections, such as moving all the functions that manipulate files to the class (only last week I moved du reimplementation to qubes.storage from old qubes.qubesutils) and dumbing down dir_path property. Those changes are probably not that hard to rebase.
> 
> You are talking about your own branch for _core4_ not the main repo right? Or did i confuse something?

Code names here are unfortunately misleading.
R1 use "core1"
R2 use "core2"
R3.0 use "core2"
R4.0 use "core3"

So, that was about R4.0, with "core3".

> > You should also be aware of two long-term plans. First, nearer, is that in liveusb we have hacked together hybrid storage, which holds templatevms on the stick (or cd) and appvms in RAM (currently by mounting tmpfs on correct directory). Such configuration (or comparable) should remain a possibility.
> 
> Just to make sure i understand it right: The AppVM volatile.img is using a mounted tmpfs. So one of the disks of the AppVM (in this case the volatile.img) is backed by a tmpfs based StoragePool. This is how i would see it.

+1

> > Second, in the unspecified future we plan to move whole storage to separate domain and keep Xen block backends there. In theory that should not affect much of the storage, and because the plan is not that specific, there is no need to create API for it just now.
> 
> I think this would only benefit from storage refactoring.

Yes.

> > API should provide two functions to get size of the image/storage, one to get size available to VM ("total disk size") and the second for getting size that is really taken from physical disk (taking into account sparse files, unallocated extents, compression, whatever). Currently they are called get_*_sz and get_*_utilization, however feel free to change the names, because the last one is very long.
> 
> We need a new issue for such stuff where we would only specify the storage rewrite to be merged with @woju R4 branch. Should i open it? I do not want to step on any toes and also not sure if it should go in to the issues project or here (because pull requests and commit msg mentions would reference  this repo). 

While at it - generally I think the better place for _discussion_ is
qubes-devel mailing list. And github issues for tracking current outcome
of the discussion (design document, updated with new ideas). Then when
finally agreed on the design and got implementation, the code should go
as a pull request for appropriate repository (obviously), referencing
appropriate issue number and design document should go to the
documentation repo (copy&paste). 

I don't think pull request is good place for discussing design. We
should move it to more appropriate place. If not moving all to
qubes-devel, at least move it to issues...


That said, the last one here:

Recently I've talked with @woju about the design here, for R4.0, but
somehow applies also here. The main concept is to have StorageManager
class, which would take care of collecting all the needed devices
together for a given VM (root_img, private_img, volatile_img,
modules_img). And that StorageManager would use StorageVolume objects as
you've designed them (`Storage` in your design)

So it would be:
QubesVm:
  .storage -> (Some)StorageManager
                .root_img -> (Some)StorageVolume
                .private_img -> (Some)StorageVolume
               ...

Then StorageManager will take care of all actions there - calling
everything required while creating VM, before starting it, after
shutting down etc.

So your classes would not replace current vm.storage property, but would
be attached to that object.

For further discussion about 4.0 design, please at least create new
issue and paste there updated proposal :)

> Now back to the main topic: **This pull request**
> I would propose the following steps to integrate the StoragePool patchset in R3.1.
> 1. I would rebase and retest on the current master(?) branch and get it hopefully merged with upstream.

Ok, indeed the change is not that big.

> 2. I would extend if I may the QubesVm to save the used storage pool for each device. The changes should not be that big. 

I don't think it is a good idea, because that change will be effectively
reverted in R4.0. Because storage pools will be moved one level down.

That would mean that having VM in different storage pool than its
template needs to be solved using something like you've already
proposed.

> 3. Extend the `qvm-create` tools so you can pass StoragePool parameter for each device.
> 
> #### Open Questions
> 1. Where is the code handling liveusb? I think i would need to implement the tmpfs StoragePool to not break it.

In `qubes-installer-qubes-os` repo - `live` directory and
`conf/liveusb.ks` file.

> 2. The windows wni storage. Is it dead code? Or do you use in you own proprietary stuff? AFAIK it's not used windows based vms? 

Currently it is dead code. We may return to this in some far future, but
for now IMHO we can simply add there some placeholder (README.txt?)
saying that it can be found in git history. And remove the file.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

kalkin:
I rebased on master and rerun tests. Seems fine to me

marmarek:
What about *any* way to make use of it? At least `qvm-create` option...

@woju can you take a look at it?  For me it looks good, but just to make sure I haven't missed anything...

kalkin:
@marmarek 
Oh i misunderstood you. I though the option should go in to R4. No problem i will add that

marmarek:
I don't see the point in merging this in R3.1, without being able to use it in R3.1.

kalkin:
@marmarek 
FULL ACK. Like i said i have misunderstood you. I will add this feature to qvm-create this week

woju:
On Mon, Oct 12, 2015 at 11:29:56AM -0700, Marek Marczykowski-Górecki wrote:
> @woju can you take a look at it?  For me it looks good, but just to
> make sure I haven't missed anything...

Looks good... However the tests aren't testing anything useful. Better
that than nothing though, just put TODO in comments.

We will most certainly reap that config file, but it is OK for now.

kalkin:
> However the tests aren't testing anything useful. Better that than nothing though, just put TODO in comments.

There were very helpful during development for making sure the behaviour is as expected. 

Also because i was not sure if the WNI storage is used I tried to develop them as if you would test them on a platform which has WNI support. I.e [here](https://github.com/kalkin/qubes-core-admin/commit/fa7b726af73cb27aecdde6e887faafa97b96da00#diff-71aca2a27b65ceafa30a0016dfe9ba55R63)

I'm also not a huge fan of using names like `TC_00` and enumerating tests, but it looked to me like a convention you agreed upon.  

> We will most certainly reap that config file, but it is OK for now.

Where would be the pools options saved? I liked the idea that you are able just to open your editor and add a storage. Of course in future you need something like `qvm-storage`.

woju:
On Tue, Oct 13, 2015 at 01:18:50AM -0700, kalkin wrote:
> > However the tests aren't testing anything useful. Better that than nothing though, just put TODO in comments.
> 
> There were very helpful during development for making sure the behaviour is as expected. 
> 
> Also because i was not sure if the WNI storage is used I tried to develop them as if you would test them on a platform which has WNI support. I.e [here](https://github.com/kalkin/qubes-core-admin/commit/fa7b726af73cb27aecdde6e887faafa97b96da00#diff-71aca2a27b65ceafa30a0016dfe9ba55R63)
> 
> I'm also not a huge fan of using names like `TC_00` and enumerating tests, but it looked to me like a convention you agreed upon.  

This is needed for correct ordering of the tests, which are sorted
lexicaly. When testing, it is OK to use other facilities that are
already tested. This way we ensure that the first test to fail it the
one that signals the problem, not something below it.

> > We will most certainly reap that config file, but it is OK for now.
> 
> Where would be the pools options saved? I liked the idea that you are
> able just to open your editor and add a storage. Of course in future
> you need something like `qvm-storage`.

As XML node under root <qubes> node (for the pools) or XML node under
the respective domain (for per-domain images or so). This will be
similar to respective <labels> and <devices> node. Maybe there will be
some extra properties like global default pool or so, but I'm not sure,
just didn't think about it enough.

I thought about something like this:

<qubes>
  <storage>
    <pool id="pool-1" class="LVMStoragePool">...</pool>
    <pool id="pool-2" class="FileStoragePool">...</pool>
    ...
  </storage>

  <domain id="..." class="AppVM">
    <storage>
      <volume pool="pool-1">...</volume>
    </storage>
  </domain>
</qubes>

What attrs and nodes go inside <pool/> and <volume/> is probably
dependent on pool class.


As the design of class hierarchy and which object is a attribute of
which, after discussion with @marmarek (as he pointed above) I thought
about something like this:

app = qubes.Qubes()

app.storage will be mapping of the pools by their id, each one will be
an object of a class that inherits from qubes.storage.StoragePool. For
example app.storage[0] will be qubes.storage.file.FileStoragePool. The
app.storage object will be some simple class like qubes.vm.DeviceManager
or something.

vm = app.domains[...]

vm.storage will be object of a child of qubes.storage.StorageManager.
Excact class will be dependent of the class of vm, because different
classes have different volumes.

Then there will be vm.storage.root_img, vm.storage.private_img,
vm.storage.volatile_img, which will be instances of class that will
inherit from qubes.storage.StorageVolume (like
qubes.storage.file.FileStorageVolume), exact class being dependent on
respective StoragePool.

There may also be some provision for ad-hoc attached devices like
cdroms, windows tools install media for w32 HVMs etc.


I hope the above makes sense.

kalkin:
Hi,

sorry I kind of neglected this because of some private issues, but it's still on my TODO list.

woju:
On Thu, Oct 29, 2015 at 04:56:55AM -0700, kalkin wrote:
> sorry I kind of neglected this because of some private issues, but it's still on my TODO list.

No problem, core3 is also delayed about 1 month, so take your time.

kalkin:
I just pushed the first draft. 
`XenStorage` uses by default `/var/lib/qubes/` as it's vmdir, but you can specify others to use by adding to the `/etc/qubes/storage.conf` the following:
```config
[pool2]
type = xen
dir = /mnt/bar
```

kalkin:
I do not want to rush you, but I need some feedback, if this is ok or I should change something, especially if you want to merge it in to R3.1

core-modules/000QubesVm.py
```diff
@@ -345,7 +346,14 @@ def __init__(self, **kwargs):
                 self.services['qubes-update-check'] = False
 
         # Initialize VM image storage class
-        self.storage = defaults["storage_class"](self)
+        self.storage = get_pool(self.pool_name, self).getStorage()
+        self.dir_path = self.storage.vmdir
+        self.icon_path = os.path.join(self.storage.vmdir, 'icon.png')
+        self.conf_file = os.path.join(self.storage.vmdir, self.name + '.conf')
+        self.appmenus_templates_dir = os.path.join(self.storage.vmdir, 'apps.templates')
+        self.appmenus_icons_dir = os.path.join(self.storage.vmdir, 'apps.icons')
```
marmarek: I think those two lines should be moved to the appmenus plugin. Maybe change that to some properties with getter calculating the path at access time (to not worry about self.dir_path not set yet).

core/storage/xen.py
```diff
@@ -249,3 +270,67 @@ def prepare_for_vm_startup(self, verbose):
...
+                string (str) absolute path to the directory where the vm files
+                             are stored
+        """
+        vm_type = type(vm)
+
+        if vm_type in [QubesAppVm, QubesHVm]:
+            subdir = 'appvms'
+        elif vm_type in [QubesTemplateVm, QubesTemplateHVm]:
+            subdir = 'vm-templates'
+        elif vm_type in [QubesAdminVm, QubesNetVm, QubesProxyVm]:
```
marmarek: This can be replaced by `issubclass(vm_type, QubesNetVm)`

core/storage/xen.py
```diff
@@ -249,3 +270,67 @@ def prepare_for_vm_startup(self, verbose):
...
+            subdir = 'servicevms'
+        else:
+            raise QubesException(str(vm_type) + ' unknown vm type')
```
marmarek: Not sure about this one. Maybe better default to some value? 'appvms'? 'other'?
How would that work for example with QubesDisposableVm?

marmarek:
Sorry, haven't noticed the update. I've made few minor comments, but generally I'm ok with this for R3.1.

core/storage/xen.py
kalkin: Added `QubesDisposableVm` handling see: 1934f06869fc280892db593321e90737c9bbb1e3

core/storage/xen.py
kalkin: Fixed it see: 9eee00c6d790c9513c7c92681f73987d2beeb537

core-modules/000QubesVm.py
kalkin: @marmarek
I tried that. I extended the [qubes-core-appmenus.py#L53-L65](https://github.com/QubesOS/qubes-core-admin-linux/blob/master/appmenus-scripts/qubes-core-appmenus.py#L53-L65) to access `self.storage.vmdir` inside the lambda. The `self.get_attrs_config()` in `QubesVm` is executed, before the `self.storage` is initialised. To initialise the `self.storage` you need to access the `pool_name` variable which is set by executing get_attrs_config().  

So we need to initialise the storage before we initialise appmenus.

kalkin:
I rebased the whole patchset on the tag v3.1.3. 

@marmarek I fixed two of the three issues you mentioned. I'm not sure how to solve the third one:
- Leave it as a hack, because it will be rewritten in R4.
- Refactor the whole storage initialisation to be more like [qubes-core-appmenus.py](https://github.com/QubesOS/qubes-core-admin-linux/blob/master/appmenus-scripts/qubes-core-appmenus.py). This way the storage hook could be initialised after `pool_name` is set, but before the appmenu hook is run.
- Any other ideas?

kalkin:
I renamed the `dir` attribute to `dir_path` and `type` to `driver`. Using attribute names which also are builtin Python functions is a joyfull expierience:
*  `assert dir is None` is always passes
* `self.dir = dir` if `dir is None` then `self.dir` is assigned the builtin function

core-modules/000QubesVm.py
marmarek: I was thinking about changing `self.appmenus_templates_dir` (in qubes-core-admin-linux) and others to more like:
```python
def get_appmenus_templates_dir(self):
    return os.path.join(self.storage.vmdir, 'apps.templates')
QubesVM.appmenus_templates_dir = property(get_appmenus_templates_dir)
```
Instead of setting them in `get_attrs_config`.
This way, you don't need `self.storage` at `self.get_attrs_config` time

kalkin:
Moved the code to `core-admin-linux`. Here is the pull request QubesOS/qubes-core-admin-linux/pull/3

marmarek:
It's all ok. Will push the merge shortly.

PR #8, opened 2015-10-24
Support more VM types in qvm-grow-root

nihilifer:
QubesOS/qubes-issues#1268


marmarek:
Generally ok, I've added few improvements for `resize2fs` call (will push soon). Additionally I think it worth making it clear that VM would be started - this may be unexpected. Something like `qvm-grow-root --allow-start`, and without that option, it would refuse operation when VM startup is required. Or some other interface, but I would rather avoid interactive prompt.

PR #10, opened 2015-11-24
libvirtError in _update_libvirt_domain is no longer silently swallowed.

v6ak:
I had some issue with fstrim and the missing else had caused the code to continue and fail later with a non-descriptive error message. This commit makes the error message more descriptive and helpful.

Should I pull this patch also to master?

Maybe I will be able to provide at least a reproduction case for the original failure, but the roots are quite old. In mycase, the xxx-dvm VM partially existed. That is, there was a directory in /var/lib/qubes/appvms (this was reported clearly by the qvm-trim-template), there was a file in /etc/libvirt/libxl/ (a hint to this fact was reported after this patch), but qvm-remove did not know anything about the VM.

### DIFF
--- a/core-modules/000QubesVm.py
+++ b/core-modules/000QubesVm.py
@@ -729,6 +729,8 @@ class QubesVm(object):
                 raise QubesException("HVM domains not supported on this "
                                      "machine. Check BIOS settings for "
                                      "VT-x/AMD-V extensions.")
+            else:
+                raise e
         self.uuid = uuid.UUID(bytes=self._libvirt_domain.UUID())
 
     @property


marmarek:
:+1: for this fix
And yes, the patch first needs to go into master. If it applies cleanly (probably yes), don't worry about it.
Reproduction case would be also useful, so we could have automated test for that.

v6ak:
OK, added a pull request for master. I was unsure about that and https://www.qubes-os.org/doc/source-code/ did not provide any answer…

I am not sure how far can I go in a reproduction case, but this might uncover multiple robustness issues. What I know is:

1. There was also some leftover VM from an unsuccessful fstrim attempt in the past. (Maybe from time of 3.0 RC1.)
2. I had libvirt/libxl out-of-sync with Qubes VM database. That is, Qubes Manager (and qvm-remove) did not know about the colliding VM, but libvirt/libxl did know about it.

Probably both can be reproduced manually. I, however, am not sure how to reproduce the odd state without creating it on purpose. Maybe by crashing something during the process of template fstrim.

Some thoughts on robustness:

* I am not sure how libvirt/libxl can get out-of-sync with Qubes VM DB. It would be, however, great, if it is handled more gracefully. Note that I actually did not know about /etc/libvirt/libxl, I have found it by something like `sudo find / -xdev -print0 | sudo xargs -0 grep <UUID of the offending VM>`. Maybe at least qvm-remove could remove the leftover VM files from libxl/libvirt.
* When a `<name>-fstrim` VM exists from the past, it should be probably also handled more gracefully. Is removing the VM too controversial?

marmarek:
>I am not sure how libvirt/libxl can get out-of-sync with Qubes VM DB. It would be, however, great, if it is handled more gracefully. Note that I actually did not know about /etc/libvirt/libxl, I have found it by something like sudo find / -xdev -print0 | sudo xargs -0 grep <UUID of the offending VM>. Maybe at least qvm-remove could remove the leftover VM files from libxl/libvirt.

Yes, it should be handled more gracefully. `qvm-remove` (actually underlying API to removing VM) already has a code to remove it, so probably you've got there with some other code path. 

> When a <name>-fstrim VM exists from the past, it should be probably also handled more gracefully. Is removing the VM too controversial?

Not sure about that. IMHO failing with a clear message that user needs to remove that VM first is a safer choice.

v6ak:
> `qvm-remove` (actually underlying API to removing VM) already has a code to remove it, so probably you've got there with some other code path.

OK. I am not sure how to reproduce the state without manually introducing that, though.

> IMHO failing with a clear message that user needs to remove that VM first is a safer choice.

OK, sounds good. The pull request makes the error message clear. (That is, a NullPointerException-like message is turned into duplicate VM name message. Still not entirely nice, but at least clear.) The only unclear thing is how to remove the VM. I did not know that libxl/libvirt has its own stored database of VMs and the qvm-remove command was not able to handle it.

So, maybe the only remaining task for this issue is making qvm-remove able to remove leftover files even if the VM appears removed.

marmarek:
This commit is already backported to release3.0 branch.

PR #13, opened 2015-12-07
Add pool support to qvm-clone

kalkin:
Added the option `-P` to select storage pool. I also removed unnecessary imports and made `qvm-clone` PEP8 conform. I hope this is ok.


kalkin:
Is there any way to get this simple change in Qubes R3.1? This would have the benefit of creating easy VM images which are stored on some other harddrive/partion.

PR #15, opened 2015-12-22
Move xen storage tests to own file

kalkin:
This is especially helpful if you want to run general storage tests + your own storage tests, without waiting for Xen storage based tests.


marmarek:
It should be named storagexen or sth like this, because those are only storage tests, not overall Xen tests (which we may have in the future).

kalkin:
@marmarek Good idea, I will do it after I'm back from the holidays and 32C3. Perhaps we will run into each other there :)

marmarek:
On Tue, Dec 22, 2015 at 04:05:27PM -0800, kalkin wrote:
> Perhaps we will run into each other there :) 

Most likely :)

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

kalkin:
Renamed to storage_xen.py

marmarek:
Thanks! Will merge it shortly (hopefully today...).

marmarek:
Odd, I think I've already merged it, specially I see the commit here:
https://github.com/QubesOS/qubes-core-admin/commit/fdd618b420f312f55c57fe47e0fd39ea3c0939ab
But for some reason github didn't noticed that.

marmarek:
Hmm, the commit has different ID, but looks pretty the same. Maybe I've done cherry-pick instead of merge... So, closing manually now.

PR #20, opened 2016-03-08
qvm-backup/qvm-backup-restore: Allow reading pass phrase from file or stdin

jgriffiths:
Passing -p/--passphrase-file uses the first line of the file passed,
or stdin if '-' is given. This works with piped input too.

Fixes 1523.

### DIFF
--- a/qvm-tools/qvm-backup
+++ b/qvm-tools/qvm-backup
@@ -30,6 +30,7 @@ import qubes.backup
 import os
 import sys
 import getpass
+from locale import getpreferredencoding
 
 def print_progress(progress):
     print >> sys.stderr, "\r-> Backing up files: {0}%...".format (progress),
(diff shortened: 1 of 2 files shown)


marmarek:
I think this is already handled in https://github.com/marmarek/qubes-core-admin/pull/11 it looks like a duplicated ticket (1523, 1653)...
But the option with a file, not only stdin may be a better choice :)

jgriffiths:
Looks like I need to track your repos as well as here, gah!

Rudd-O:
First line?  Why only first line?  Randomly generated data piped into -p could give extremely low entropy based on where the newline falls.

jgriffiths:
The password must be able to be typed in for feature parity, raw binary data is not supported.

If you want to create a random large passphrase then hex or base64 encode it before passing it in.

Rudd-O:
Doesn't that reduce entropy?  Is it a bad idea to introduce a specific parameter for random data read from a file-like entity?

jgriffiths:
Entropy is not affected by the encoding used to represent it, it simply takes more bytes to encode a given amount of it using a less efficient encoding. If you want to encrypt with a binary file, pass it though a binary to text filter using stdin.

PR #21, opened 2016-03-11
Don't try to provide debug files where none are generated

jgriffiths:
Comments as per the parallel submit to qubes-artwork.


marmarek:
Ok.

marmarek:
Please note that currently that python scripts are put in `/usr/lib64/python2.7/site-packages`. But this actually will change soon, so this PR is ok.

jgriffiths:
I've changed this so it should be mergable as it - when the python scripts are moved it can be set noarch then per the comments.

PR #23, opened 2016-03-13
implemented dom0 qubes.GetRandomizedTime

adrelanos:
Required for fixing 'sys-whonix doesn't connect to Tor after system suspend'.

https://github.com/QubesOS/qubes-issues/issues/1764

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,8 +76,10 @@ endif
 	cp qubes-rpc-policy/qubes.NotifyUpdates.policy $(DESTDIR)/etc/qubes-rpc/policy/qubes.NotifyUpdates
 	cp qubes-rpc-policy/qubes.NotifyTools.policy $(DESTDIR)/etc/qubes-rpc/policy/qubes.NotifyTools
 	cp qubes-rpc-policy/qubes.GetImageRGBA.policy $(DESTDIR)/etc/qubes-rpc/policy/qubes.GetImageRGBA
+	cp qubes-rpc-policy/qubes.GetRandomizedTime.policy $(DESTDIR)/etc/qubes-rpc/policy/qubes.GetTime
 	cp qubes-rpc/qubes.NotifyUpdates $(DESTDIR)/etc/qubes-rpc/
 	cp qubes-rpc/qubes.NotifyTools $(DESTDIR)/etc/qubes-rpc/
+	cp qubes-rpc/qubes.GetRandomizedTime $(DESTDIR)/etc/qubes-rpc/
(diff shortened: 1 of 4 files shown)


qubes-rpc/qubes.GetRandomizedTime
```diff
@@ -0,0 +1,80 @@
...
+
+## Get a random 0 or 1.
+## Will use this to decide to use plus or minus.
+##
+## Thanks to
+## http://linux.byexamples.com/archives/128/generating-random-numbers/
+ZERO_OR_ONE="$(( 0+($(od -An -N2 -i /dev/random) )%(0+2) ))"
+
+## Create a random number between 0 and 180.
+DELAY="$(( $(od -An -N2 -i /dev/random)%(180-0+1) ))"
```
marmarek: Why use different methods of getting random number? `shuf` usage below seems much more readable. And according to the second answer in https://stackoverflow.com/questions/22887891/how-can-i-get-a-random-dev-random-number-between-0-and-999999999-in-bash, using modulo doesn't provide uniform distribution.

marmarek:
Other than the above comment, looks good!

qubes-rpc/qubes.GetRandomizedTime
adrelanos: No more reason to do so. (No longer excluding +/- 5.) Much more readable indeed! Fixed that.

(If it looks good, I'll also fix that in bootclockrandomization.)

Makefile
```diff
@@ -76,8 +76,10 @@ endif
 	cp qubes-rpc-policy/qubes.NotifyUpdates.policy $(DESTDIR)/etc/qubes-rpc/policy/qubes.NotifyUpdates
 	cp qubes-rpc-policy/qubes.NotifyTools.policy $(DESTDIR)/etc/qubes-rpc/policy/qubes.NotifyTools
 	cp qubes-rpc-policy/qubes.GetImageRGBA.policy $(DESTDIR)/etc/qubes-rpc/policy/qubes.GetImageRGBA
+	cp qubes-rpc-policy/qubes.GetRandomizedTime.policy $(DESTDIR)/etc/qubes-rpc/policy/qubes.GetTime
```
marmarek: Wrong target filename.

marmarek:
You need to add new files to `rpm_spec/core-dom0.spec`

Makefile
adrelanos: Fixed.

adrelanos:
Just now added. Please have a look.

PR #29, opened 2016-04-27
Fixed #1930: Delete autostart Symlink when deleting a VM

MarioGeckler:
This patch fixes Issue #1930.
When deleting a VM the autostart Symlink is deleted if it exists.

### DIFF
--- a/core-modules/000QubesVm.py
+++ b/core-modules/000QubesVm.py
@@ -1412,6 +1412,11 @@ class QubesVm(object):
                     e.get_error_code())
                 raise
 
+        if os.path.exists("/etc/systemd/system/multi-user.target.wants/qubes-vm@" + self.name + ".service") == True:
+            retcode = subprocess.call(["sudo", "rm", "/etc/systemd/system/multi-user.target.wants/qubes-vm@" + self.name + ".service"])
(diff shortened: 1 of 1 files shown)


core-modules/000QubesVm.py
```diff
@@ -1412,6 +1412,11 @@ def remove_from_disk(self):
                     e.get_error_code())
                 raise
 
+        if os.path.exists("/etc/systemd/system/multi-user.target.wants/qubes-vm@" + self.name + ".service") == True:
```
marmarek: `== True` is unnecessary.

core-modules/000QubesVm.py
```diff
@@ -1412,6 +1412,11 @@ def remove_from_disk(self):
...
+            retcode = subprocess.call(["sudo", "rm", "/etc/systemd/system/multi-user.target.wants/qubes-vm@" + self.name + ".service"])
```
marmarek: Take a look at handling disable autostart - use systemctl for consistency.

MarioGeckler:
Changed remove symlink to sysctl
removed unnecesary comparison with True

marmarek:
Thanks!

PR #30, opened 2016-05-14
Tool to create bug reports.

Jeeppler:
A bug report is a collection of system information and log files for a specific qube.

Please review the file before merging.

### DIFF
--- a/qubes-bug-report
+++ b/qubes-bug-report
@@ -1,0 +1,213 @@
+#!/usr/bin/env python3
+
+import subprocess as sub
+import shlex
+import argparse
+import time
+import sys
+
+from os.path import expanduser
+
+#the term qube refers to a qubes vm
+
+def is_program_installed_in_qube( qube_name, command ):
+	is_installed = True
+
+	try:
+		shell_cmd = "qvm-run " + shlex.quote( qube_name ) + " --pass-io --no-color-output 'command -v " + command + "' &> /dev/null"
+		sub.check_call( shell_cmd, shell = True )
+
+	except sub.CalledProcessError:
+		is_installed = False
+
+	return is_installed
+
+
+#this function requires virsh 
+#domstate only works for Xen domU (guests)
+def is_qube_running( qube_name ):
+	runs = False
+
+	out = sub.check_output([ "virsh", "-c", "xen:///", "domstate", shlex.quote( qube_name ) ])
+	out = out.decode('utf-8').replace('\n', '')
+
+	if 'running' == out:
+		runs = True
+
+	return runs
+
+
+def get_qube_packages( qube_name ):
+	content = "## Qubes Packages\n\n"
+
+	#a qube can have more than one package manager installed (only one is functional)
+	pkg_cmd = { 'dpkg' : 'dpkg -l qubes-*', 'pacman' : 'pacman -Qs', 'rpm' : 'rpm -qa qubes-*' }
+
+	if is_qube_running( qube_name ):
+
+		for package_manager in pkg_cmd.keys():
+			if is_program_installed_in_qube( qube_name, package_manager ):
+				pkg_list_cmd = pkg_cmd[package_manager]
+
+				try:
+					shell_cmd = "qvm-run " + shlex.quote( qube_name ) + " --pass-io --no-color-output '" +  pkg_list_cmd + "' 2> /dev/null"
+					out = sub.check_output( shell_cmd, shell = True )
+					out = out.decode('utf-8')
+					content = content + "### Package Manager: " + package_manager + "\n\n"
+					content = content + wrap_code( out )
+				except sub.CalledProcessError:
+					True #do nothing
+
+	else:
+		content = content + "**No packages listed, because Qube " + qube_name + " was not running**\n\n"	
+
+	return content		
+
+def get_dom0_packages():
+	content = "## Dom0 Packages\n\n"
+	out = sub.check_output([ "rpm", "-qa", "qubes-*" ])
+	out = out.decode('utf-8')
+	content = content + wrap_code( out )
+	
+	return content
+
+
+def wrap_code( text ):
+	code = "~~~\n" + text + "~~~\n\n"
+	
+	return code
+
+
+def get_log_file_content( qube_name ):
+	content = "## Log Files\n\n"
+	qubes_os_log = "/var/log/qubes/"
+	ext = ".log"
+
+	log_prefix = [ "guid", "pacat", "qubesdb", "qrexec" ]
+
+	for prefix in log_prefix:
+		log_file = prefix + "." + qube_name + ext
+		content = content + "### Log File: " + log_file + "\n\n"
+		content = content + wrap_code( get_log_file( qubes_os_log + log_file ) )
+
+	return content
+
+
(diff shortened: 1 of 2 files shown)


qubes-bug-report
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env python3
+
+import subprocess as sub
```
marmarek: Please use just `import subprocess` - module aliases makes it harder to read the script

marmarek:
Hmm, as looking at it, there is a lot of `qvm-*` tools usage. Maybe better use qubes API, instead of separate processes? But that would mean downgrading to python2 (at least in current Qubes version)...

Then the code would look like:
```python
import qubes.qubes

qc = qubes.qubes.QubesVmCollection()
qc.lock_db_for_reading()
qc.load()
qc.unlock_db()

vm = qc.get_vm_by_name("somename")
# just run a command:
exitcode = vm.run("command", wait=True)
# run and get subprocess.Popen object (for stdin/out):
popen = vm.run("command", passio_popen=True)
```

@woju what do you think? Does it worth right now? Or maybe only when porting this tool to Qubes 4.0?

qubes-bug-report
```diff
@@ -0,0 +1,213 @@
...
+from os.path import expanduser
+
+#the term qube refers to a qubes vm
+
+def is_program_installed_in_qube( qube_name, command ):
+	is_installed = True
+
+	try:
+		shell_cmd = "qvm-run " + shlex.quote( qube_name ) + " --pass-io --no-color-output 'command -v " + command + "' &> /dev/null"
+		sub.check_call( shell_cmd, shell = True )
```
marmarek: Better not use `shell=True` and properly split command args array (place `'command -v ...'` as a single argument).

qubes-bug-report
```diff
@@ -0,0 +1,213 @@
...
+
+	if is_qube_running( qube_name ):
+
+		for package_manager in pkg_cmd.keys():
+			if is_program_installed_in_qube( qube_name, package_manager ):
+				pkg_list_cmd = pkg_cmd[package_manager]
+
+				try:
+					shell_cmd = "qvm-run " + shlex.quote( qube_name ) + " --pass-io --no-color-output '" +  pkg_list_cmd + "' 2> /dev/null"
+					out = sub.check_output( shell_cmd, shell = True )
```
marmarek: Same here (`shell`).

qubes-bug-report
```diff
@@ -0,0 +1,213 @@
...
+		runs = True
+
+	return runs
+
+
+def get_qube_packages( qube_name ):
+	content = "## Qubes Packages\n\n"
+
+	#a qube can have more than one package manager installed (only one is functional)
+	pkg_cmd = { 'dpkg' : 'dpkg -l qubes-*', 'pacman' : 'pacman -Qs', 'rpm' : 'rpm -qa qubes-*' }
```
marmarek: Does pacman support listing only `qubes-*` packages?

qubes-bug-report
```diff
@@ -0,0 +1,213 @@
...
+				pkg_list_cmd = pkg_cmd[package_manager]
+
+				try:
+					shell_cmd = "qvm-run " + shlex.quote( qube_name ) + " --pass-io --no-color-output '" +  pkg_list_cmd + "' 2> /dev/null"
+					out = sub.check_output( shell_cmd, shell = True )
+					out = out.decode('utf-8')
+					content = content + "### Package Manager: " + package_manager + "\n\n"
+					content = content + wrap_code( out )
+				except sub.CalledProcessError:
+					True #do nothing
```
marmarek: `pass`

qubes-bug-report
Jeeppler: Yes, it does:

~~~
% pacman -Qs qubes 
local/qubes-db-vm 3.1.2-3
    QubesDB libs and daemon service.
local/qubes-libvchan-xen 3.1.0-2
    The Qubes core libraries for installation inside a Qubes Dom0 and VM.
local/qubes-vm-core 3.1.12-2
    The Qubes core files for installation inside a Qubes VM.
local/qubes-vm-gui 3.1.4-5
    The Qubes GUI Agent for AppVMs
local/qubes-vm-utils 3.1.7-5
    Common Linux files for Qubes VM.
local/qubes-vm-xen 4.6.0-12
    Xen is a virtual machine monitor
~~~

but `qubes-*` does not work it should be just `qubes`

qubes-bug-report
Jeeppler: The qvm-run script writes always to the console, to avoid that I used the `/dev/null` workaround, which did not work properly with the normal `check_output`.

qubes-bug-report
```diff
@@ -0,0 +1,213 @@
...
+	for prefix in log_prefix:
+		log_file = prefix + "." + qube_name + ext
+		content = content + "### Log File: " + log_file + "\n\n"
+		content = content + wrap_code( get_log_file( qubes_os_log + log_file ) )
+
+	return content
+
+
+def get_qube_prefs( qube_name ):
+	qube_prefs = sub.check_output([ "qvm-prefs", shlex.quote( qube_name ) ])
```
marmarek: When using args array (instead of `shell=True`), quoting isn't necessary.

qubes-bug-report
```diff
@@ -0,0 +1,213 @@
...
+
+	title = "Bug report: " + qube_name
+
+	content = get_qube_prefs( qube_name )
+	content = content + get_dom0_packages()
+	content = content + get_log_file_content( qube_name )
+	content = content + get_qube_packages( qube_name )
+	
+
+	report = template.format( **locals() )
```
marmarek: Use explicit map here, `locals()` would be hard to debug in the future.

qubes-bug-report
```diff
@@ -0,0 +1,213 @@
...
+
+def qube_exist( qube_name ):
+	exists = True
+	
+	try:
+		#calls: qvm-check --quiet vmanme
+		sub.check_call([ "qvm-check", "--quiet", shlex.quote( qube_name ) ])
+
+	except sub.CalledProcessError:
+		exists = False
```
marmarek: `return False`

qubes-bug-report
```diff
@@ -0,0 +1,213 @@
...
+
+	return exists
```
marmarek: `return True`

qubes-bug-report
marmarek: Pass `stderr=open(os.devnull, 'w')`.

qubes-bug-report
Jeeppler: I never want to break the control flow in the middle of a function. I use a flag instead and return from the function only at the end.

Jeeppler:
@marmarek can you review this version of the qubes-bug-report tool?

marmarek:
Generally looks good now. There are few minor code style details, like `x += y` instead of `x = x + y`, but it may stay as it is. But please convert indentation to 4 spaces instead of tabs (https://www.qubes-os.org/doc/coding-style/#tocAnchor-1-1-2).

Jeeppler:
@marmarek can you please review the script again?

marmarek:
Thanks!

PR #31, opened 2016-06-21
qvm-ls --raw-data option added

GammaSQ:
The intention is to have an option that returns raw data without formatting or unnecessary data, mainly so it can easily be parsed by other bash-scripts. Therefore, specified VM-ordering should not be changed and all fields are returned in order as well.

Main drawback: I'm using the internal key-strings, so eg. "last backup" has to be passed as "--raw_data last\ backup"

### DIFF
--- a/qvm-tools/qvm-ls
+++ b/qvm-tools/qvm-ls
@@ -150,37 +150,41 @@ def main():
     cpu_usages = None
 
     if arguments.raw_data:
-        if ''
-    fields_to_display = ["name", "on", "state", "updbl", "type", "template", "netvm", "label" ]
+        fields_to_display = arguments.raw_data
+        if 'cpu' in arguments.raw_data:
+            qhost = QubesHost()
+            (measure_time, cpu_usages) = qhost.measure_cpu_usage(qvm_collection)
+    else:
+        fields_to_display = ["name", "on", "state", "updbl", "type", "template", "netvm", "label" ]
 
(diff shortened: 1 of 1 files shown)


qvm-tools/qvm-ls
```diff
@@ -24,7 +24,7 @@
 from qubes.qubes import QubesVmCollection
 from qubes.qubes import QubesHost
 from qubes.qubes import QubesException
-from optparse import OptionParser
+from argparse import ArgumentParser
```
marmarek: For the next time: please make such changes as a separate commit - one for changing library but without any functional change, then actual change. It will make review much easier.

GammaSQ: The Problem with OptionParser is that it doesn't allow the 'nargs="+"'-option. I wanted to have as many fields specified with one command as required. With OptionParser, 'nargs=1' is the only meaningful option => for n fields I need n commands.

marmarek: I'm perfectly fine with this change. It's just about having it in separate commit.

marmarek:
Merged with a minor fix (https://github.com/QubesOS/qubes-core-admin/commit/84677fa70bbc716abf223e639e0bf6f7771465ba)

marmarek:
(for the next time) take a look at https://www.qubes-os.org/doc/code-signing/
We're going to enforce this rule soon (will be announced).

PR #32, opened 2016-06-24
Add VM state options

ttasket:
New options: --running, --paused and --template


marmarek:
(for the next time) take a look at https://www.qubes-os.org/doc/code-signing/
We're going to enforce this rule soon (will be announced).

PR #34, opened 2016-06-29
Make qfile-daemon-dvm more flexible

rustybird:
qfile-daemon-dvm currently assumes you want to start a dispVM, run exactly one qrexec service, and tear down the dispVM when that service is done. It also does not tell you the name of the dispVM while it's running.

I often wanted to do something like start a dispVM, run a service, then e.g. attach a block device, then run another service etc. The PR allows this in a minimalistic fashion: If the action is LAUNCH instead of qubes.SomeService, then just start the dispVM, write its name to stdout, and quit. If the action is FINISH, then tear down the named dispVM. (Uppercase words, because you can already use DEFAULT in the username argument to request that the service is run as the default user.)

### DIFF
--- a/dispvm/qfile-daemon-dvm
+++ b/dispvm/qfile-daemon-dvm
@@ -171,6 +171,10 @@ class QfileDaemonDvm:
 
 def main():
     exec_index = sys.argv[1]
+    if exec_index == "FINISH":
+        QfileDaemonDvm.finish_disposable(sys.argv[2])
+        return
+
     src_vmname = sys.argv[2]
     user = sys.argv[3]
     # accessed directly by get_dvm()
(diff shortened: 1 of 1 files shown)


marmarek:
There is a way to do this even now - call `qubes.VMShell`, print DispVM name from inside. For example take a look here:
https://github.com/QubesOS/qubes-core-admin/blob/master/tests/basic.py#L738-L742

This will be also much easier in Qubes 4.0:
https://github.com/QubesOS/qubes-core-admin/blob/core3-devel/qubes/vm/dispvm.py#L78-L106

Anyway I'll merge your change, as it seems to be useful just now in 3.2.

rustybird:
> There is a way to do this even now - call qubes.VMShell, print DispVM name from inside.

Ah, hadn't thought of this.

> Anyway I'll merge your change, as it seems to be useful just now in 3.2.

Thanks! I'm using it in split-dm-crypt.

rustybird:
> There is a way to do this even now - call qubes.VMShell, print DispVM name from inside. For example take a look here:
> https://github.com/QubesOS/qubes-core-admin/blob/master/tests/basic.py#L738-L742

Not a great pattern in dom0 though; e.g. it looks like the dispVM could mess with the tester's head a little by lying that its name is "test-cls-vm1", weirdly making the condition at https://github.com/QubesOS/qubes-core-admin/blob/master/tests/basic.py#L767-L768 fail. (Can you send my $100,000 bug bounty in bitcoins please)

marmarek:
Generally we don't care about VMs being compromised on tests. Yes, it may mess with test results, so be it. Tests systems are frequently reinstalled.
But indeed, not the best idea for other use cases.

PR #40, opened 2016-07-13
qvm-ls hide stacktrace if qubes.xml is missing

kalkin:


kalkin:

PR #41, opened 2016-07-13
Remove obsolete options from .pylintrc

kalkin:
This options are removed in current pylint versions. This prevents pylint from
barfing all over in vim/syntastic.


kalkin:

PR #42, opened 2016-07-13
Add LVM thin pool driver to storage

kalkin:
Currently I use an addition binary called `qubes-lvm` which is executed with
`sudo(8)` to achieve root to be able manipulate lvm volumes.

### DIFF
--- a/doc/manpages/qvm-clone.rst
+++ b/doc/manpages/qvm-clone.rst
@@ -17,7 +17,7 @@ Options
 .. option:: -P POOL
 
     Pool to use for the new domain. All volumes besides snapshots volumes are
-    imported in to the specified POOL. ~HIS IS WHAT YOU WANT TO USE NORMALLY.
+    imported in to the specified POOL. THIS IS WHAT YOU WANT TO USE NORMALLY.
 
 .. option:: --pool=POOL:VOLUME, -p POOL:VOLUME
 
--- a/qubes/storage/__init__.py
+++ b/qubes/storage/__init__.py
@@ -310,7 +310,7 @@ class Storage(object):
 
     def create(self):
         ''' Creates volumes on disk '''
-        old_umask = os.umask(002)
+        old_umask = os.umask(0o002)
 
         for volume in self.vm.volumes.values():
             self.get_pool(volume).create(volume)
--- a/qubes/storage/file.py
+++ b/qubes/storage/file.py
@@ -186,7 +186,7 @@ class FilePool(qubes.storage.Pool):
             old_path = volume.path_cow + '.old'
             os.rename(volume.path_cow, old_path)
 
-        old_umask = os.umask(002)
+        old_umask = os.umask(0o002)
         with open(volume.path_cow, 'w') as f_cow:
             f_cow.truncate(volume.size)
         os.umask(old_umask)
--- a/qubes/storage/lvm.py
+++ b/qubes/storage/lvm.py
@@ -34,50 +34,39 @@ class ThinPool(qubes.storage.Pool):
 
     driver = 'lvm_thin'
 
-    def __init__(self, volume_group, thin_pool, **kwargs):
-        super(ThinPool, self).__init__(**kwargs)
+    def __init__(self, volume_group, thin_pool, revisions_to_keep=1, **kwargs):
+        super(ThinPool, self).__init__(revisions_to_keep=revisions_to_keep,
+                                       **kwargs)
         self.volume_group = volume_group
         self.thin_pool = thin_pool
         self._pool_id = "{!s}/{!s}".format(volume_group, thin_pool)
         self.log = logging.getLogger('qube.storage.lvm.%s' % self._pool_id)
-
-    def backup(self, volume):
-        msg = "Expected volume_type 'snap' got {!s}"
-        msg = msg.format(volume.volume_type)
-        assert volume.volume_type == 'snap', msg
-        cmd = ['remove', volume.vid + "-back"]
-        qubes_lvm(cmd, self.log)
-        cmd = ['clone', volume.vid, volume.vid + "-back"]
-        qubes_lvm(cmd, self.log)
-        volume.backups = [volume.vid + "-back"]
-        return volume
 
     def clone(self, source, target):
         cmd = ['clone', source.vid, target.vid]
         qubes_lvm(cmd, self.log)
         return target
 
-    def commit(self, volume):
-        msg = "Expected rw:True & volume_type 'snap' got {!s} & rw:{!s}"
-        msg = msg.format(volume.volume_type, volume.rw)
-        assert volume.volume_type == 'snap' and volume.rw, msg
-        assert volume.path.endswith("-snap")
-        self.backup(volume)
-        cmd = ['remove', volume.vid]
-        qubes_lvm(cmd, self.log)
-        cmd = ['clone', volume.path, volume.vid]
-        qubes_lvm(cmd, self.log)
+    def _commit(self, volume):
+        msg = "Trying to commit {!s}, but it has save_on_stop == False"
+        msg = msg.format(volume)
+        assert volume.save_on_stop, msg
 
-    def _backup(self, volume):
-        msg = "Expected volume_type 'snap' got {!s}"
-        msg = msg.format(volume.volume_type)
-        assert volume.volume_type == 'snap', msg
+        msg = "Trying to commit {!s}, but it has rw == False"
+        msg = msg.format(volume)
+        assert volume.rw, msg
+        assert hasattr(volume, '_vid_snap')
+
         cmd = ['remove', volume.vid + "-back"]
         qubes_lvm(cmd, self.log)
-        cmd = ['clone', volume.vid, volume.vid + "-back"]
+        cmd = ['clone', volume._vid_snap, volume.vid + "-back"]
         qubes_lvm(cmd, self.log)
-        volume.backups = [volume.vid + "-back"]
-        return 'XXX'
+
+        cmd = ['remove', volume.vid]
+        qubes_lvm(cmd, self.log)
+        cmd = ['clone', volume._vid_snap, volume.vid]
+        qubes_lvm(cmd, self.log)
+        cmd = ['remove', volume._vid_snap]
 
     @property
     def config(self):
@@ -188,6 +177,18 @@ class ThinPool(qubes.storage.Pool):
             volume._vid_snap = volume.vid + '-snap'
         return volume
 
+    def revert(self, volume, revision=None):
+        old_path = volume.path + '-back'
+        if not os.path.exists(old_path):
+            msg = "Volume {!s} has no {!s}".format(volume, old_path)
+            raise qubes.storage.StoragePoolException(msg)
+
+        cmd = ['remove', volume.vid]
+        qubes_lvm(cmd, self.log)
+        cmd = ['clone', volume.vid + '-back', volume.vid]
+        qubes_lvm(cmd, self.log)
+        return volume
+
     def _reset(self, volume):
         self.remove(volume)
         self.create(volume)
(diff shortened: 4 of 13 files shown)


kalkin:
I will rebase this PR as soon as #39 is merged

doc/manpages/qvm-clone.rst
```diff
@@ -1,30 +1,39 @@
...
+-------
 
 .. option:: --help, -h
 
     Show this help message and exit
 
+.. option:: -P POOL
+
+    Pool to use for the new domain. All volumes besides snapshots volumes are
+    imported in to the specified POOL. ~HIS IS WHAT YOU WANT TO USE NORMALLY.
```
woju: **T**HIS?

qubes/storage/__init__.py
```diff
@@ -49,58 +51,118 @@ class StoragePoolException(qubes.exc.QubesException):
...
+    def revisions(self):
+        ''' Returns a `dict` containing revision identifiers and paths '''
+        msg = "{!s} has revisions not implemented".format(self.__class__)
+        raise NotImplementedError(msg)
 
-    def __neq__(self, other):
-        return not self.__eq__(other)
+    @property
+    def config(self):
+        ''' return config data for serialization to qubes.xml '''
```
woju: Is it used anywhere else? If not, this should be moved to `__xml__`.

qubes/storage/__init__.py
```diff
@@ -198,37 +305,55 @@ def get_disk_utilization(self):
...
+        ''' Resizes volume a read-writable volume '''
         self.get_pool(volume).resize(volume, size)
 
-    def create(self, source_template=None):
+    def create(self):
         ''' Creates volumes on disk '''
-        if source_template is None and hasattr(self.vm, 'template'):
-            source_template = self.vm.template
-
         old_umask = os.umask(002)
```
woju: `0o002`

qubes/storage/__init__.py
```diff
@@ -320,109 +455,168 @@ class Pool(object):
...
-    def commit_template_changes(self, volume):
-        ''' Update origin device '''
-        raise NotImplementedError(
-            "Pool %s has commit_template_changes() not implemented" %
-            self.name)
+    def commit(self, volume):  # pylint: disable=no-self-use
+        ''' Write the snapshot to disk '''
+        msg = "Got volume_type {!s} when expected 'snap'"
+        msg = msg.format(volume.volume_type)
+        assert volume.volume_type == 'snap', msg
```
woju: Is this method doing something? I see only `assert`. Should it be that way?

woju: And, could you just write:
```python
    assert volume.volume_type == snap, \
        "Got volume_type {} when expected 'snap'".format(volume.volume_type)
```
(here and in other places also)

qubes/storage/__init__.py
```diff
@@ -320,109 +455,168 @@ class Pool(object):
...
+    def _not_implemented(self, method_name):
+        ''' Helper for emitting helpful `NotImplementedError` exceptions '''
+        msg = "Pool driver {!s} has {!s}() not implemented"
+        msg = msg.format(str(self.__class__.__name__), method_name)
+        return NotImplementedError(msg)
+
+
+def _sanitize_config(config):
+    ''' Helper function to convert types to appropriate strings
+    '''  # FIXME: find another solution for serializing basic types
```
woju: Have you looked at `qubes.property.bool`?

qubes/storage/file.py
```diff
@@ -22,134 +22,169 @@
...
-
-        return volume
-
     @property
     def config(self):
         return {
             'name': self.name,
             'dir_path': self.dir_path,
             'driver': FilePool.driver,
+            'revisions_to_keep': self.revisions_to_keep
```
woju: Didn't we agree that we drop the file volumes precisely because we don't want to reimplement LVM to the point of keeping old revisions? This was a major argument, with @marmarek's story of people destroying their templates because they try to boot them and then they have no rollback point.

qubes/tools/qubes_lvm.py
```diff
@@ -0,0 +1,272 @@
...
+
+from __future__ import print_function
+
+import datetime
+import logging
+import os
+import subprocess
+import sys
+import time
+import lvm  # pylint: disable=import-error
```
woju: Instead of disabling pylint, I'd like the module to be included in `ci/requirements.txt` and installed during travis build. If it's possible.

qubes/tools/qubes_lvm.py
```diff
@@ -0,0 +1,272 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+''' Manage pools and volumes managed by the 'lvm_thin' driver. '''
```
woju: Functions from this file *probably* should be in `qubes/storage/`, not `qubes/tools/`. Maybe as (non-standard) methods of pool, maybe not. qubes-tools shouldn't touch any external utils, they should be just consumers of the API.

qubes/storage/lvm.py
```diff
@@ -0,0 +1,344 @@
...
+    ''' Return true if pool exists '''
+    cmd = ['pool', pool_id]
+    return qubes_lvm(cmd)
+
+
+def qubes_lvm(cmd, log=logging.getLogger('qube.storage.lvm')):
+    ''' Call :program:`qubes-lvm` to execute an LVM operation '''
+    # TODO Refactor this ones the udev groups gets fixed and we don't need root
+    # for operations on lvm devices
+    cmd = ['sudo', 'qubes-lvm'] + cmd
```
woju: This should be other way around. The individual `qubes-lvm` subcommands should call `sudo lvm ...` if needed, we shouldn't elevate parts of core to root priviledges.

woju: And as I wrote earlier/below, those functions should be probably in this file.

qubes/storage/lvm.py
```diff
@@ -0,0 +1,344 @@
...
+            self._vid_snap = self.vid + '-snap'
+
+    @property
+    def revisions(self):
+        path = self.path + '-back'
+        if os.path.exists(path):
+            seconds = os.path.getctime(path)
+            iso_date = qubes.storage.isodate(seconds).split('.', 1)[0]
+            return {iso_date: path}
+        return {}
```
woju: This assumes number of revisions to be equal to 1.

qubes/tools/qvm_create.py
```diff
@@ -42,15 +42,21 @@
     default='AppVM',
     help='specify the class of the new domain (default: %(default)s)')
 
-parser.add_argument('--property', '--prop', '-p',
+parser.add_argument('--property', '--prop',
```
woju: I object to this change. I think `--property` would be used more often than `--pool`.

qubes/tools/qvm_clone.py
```diff
@@ -0,0 +1,73 @@
...
+
+import sys
+
+from qubes.tools import QubesArgumentParser, SinglePropertyAction
+
+parser = QubesArgumentParser(description=__doc__, vmname_nargs=1)
+parser.add_argument('new_name',
+                    metavar='NEWVM',
+                    action=SinglePropertyAction,
+                    help='name of the domain to create')
```
woju: `SinglePropertyAction` was meant as helper for a situation, where there was more than one way to set a property. For example, in `qvm-create` you can set label with either `--property label=red` or `--label red`. The same for name, technically you don't need to specify the name as positional argument, it is enough to set it with `--property`.

So `'new_name'` is wrong, it should be `'name'`. Then, there should be that `--property` action as in `qvm-create`.
I'd imagine that after doing `.clone_properties`, the properties would then be overriden from the command line.

qubes/storage/file.py
marmarek: On Thu, Jul 21, 2016 at 08:44:11AM -0700, Wojtek Porczyk wrote:
> >      @property
> >      def config(self):
> >          return {
> >              'name': self.name,
> >              'dir_path': self.dir_path,
> >              'driver': FilePool.driver,
> > +            'revisions_to_keep': self.revisions_to_keep
> 
> Didn't we agree that we drop the file volumes precisely because we don't want to reimplement LVM to the point of keeping old revisions? This was a major argument, with @marmarek's story of people destroying their templates because they try to boot them and then they have no rollback point.

No, using LVM thin make it possible to easily, without much overhead,
have more than one revision (so not only the last one which is easy to
destroy by starting the template once).

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

qubes/tools/qvm_clone.py
```diff
@@ -0,0 +1,73 @@
...
+        dst_vm.clone_disk_files(src_vm, pools=args.pools)
+    else:
+        dst_vm.clone_disk_files(src_vm)
+
+#   try:
+    app.save()  # HACK remove_from_disk on exception hangs for some reason
+#   except Exception as e:  # pylint: disable=broad-except
+#       dst_vm.remove_from_disk()
+#       parser.print_error(e)
+#   return 0
```
marmarek: Read the comment - actually the code should be uncommented but do not
work for some reason.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

qubes/storage/__init__.py
kalkin: No it is used as a part of the API.

qubes/tools/qubes_lvm.py
kalkin: I think we don't need it at all. because lvm2 should be installed as dependency by rpm. It was only for my local installation, because my DevVM had no lvm2 installed

qubes/storage/lvm.py
kalkin: > The individual qubes-lvm subcommands should call sudo lvm ... if needed, we shouldn't elevate parts of core to root priviledges

The reason for separating the lvm part in `qubes-lvm` was to not rely on `sudo lv*` commands, but to reuse the lvm2 library. In such case your executable needs to run as root. This executable is `qubes-lvm`. Because at the end everything worked out, I have not migrated yet all the 'lv *' calls in `qubes_lvm`.

In future I need to figure out the udev rules to make the lvm devices accessible from the `qubes` group.

qubes/storage/lvm.py
kalkin: Currently it's hardcoded that lvm can only have 1 revision (as the file store). It would be easy to add, but currently also no tool displays all the revisions.

kalkin:
@woju Can I somehow help you to get it merge/accepted?

kalkin:
@woju I know your aversion against IM, but if it's easier and faster for you we can make an appointment and go through it both, chatting/talking via XMPP or Skype.

kalkin:
Looks like trusty doesn't have a `python-lvm2` package and xenial is not available. So what know? Obviosly it should be possible to build `python-lvm2` yourself and build it's dependencies, but it's overkill. May be we can use somehow Travis CI Docker support?

marmarek:
As @woju already pointed out in https://github.com/QubesOS/qubes-core-admin/pull/42#discussion_r71732559 , add it to `ci/requirements.txt`. It will be installed using pip (as you can see in `.travis.yml`). And please drop the hack you've just pushed.

kalkin:
@marmarek python-lvm2 is not provided by pip.

marmarek:
Ah. I see two options here:
 - provide lvm2 mockup in `test-packages/` (just interface, without actual implementation)
 - configure pylint to ignore this particular error

I'm slightly for the second option, as on non-travis test run it will still be checked that way (pylint will still notice errors in lvm2 API usage).

PR #47, opened 2016-07-26
Move /var/run/shm.id to /var/run/qubes/shm.id

jaspertron:
This allows the gui daemon to be run without root privileges.


jaspertron:
Fixes QubesOS/qubes-issues#2195.
Requires QubesOS/qubes-gui-daemon#3.

PR #50, opened 2016-08-19
Updated manpages to match current tool options

kulinacs:
- Updated existing manpages to match options shown with the tools help message
- Removed qvm-clone-template.rst as it had no match in qvm-tools
- Added qvm-check.rst, qvm-grow-root.rst, qvm-usb.rst


marmarek:
Thanks!

PR #51, opened 2016-08-28
Accumulated bug fixes & improvements

kalkin:


qubes/tools/qvm_start.py
```diff
@@ -133,13 +133,17 @@ def main(args=None):
...
-    except qubes.exc.QubesException as e:
-        parser.error_runtime('Qubes error: {!r}'.format(e))
-
+    else:
+        try:
+            vm.start(
+                preparing_dvm=args.preparing_dvm,
+                start_guid=args.start_guid)
+        except qubes.exc.QubesException as e:
+            parser.error_runtime('Qubes error: {!r}'.format(e))
```
marmarek: Shouldn't `return 1` be here?

qubes/storage/lvm.py
```diff
@@ -227,6 +229,12 @@ def stop(self, volume):
         return volume
 
     def _snapshot(self, volume):
+        try:
+            cmd = ['remove', volume._vid_snap]
+            qubes_lvm(cmd, self.log)
+        except:
```
marmarek: add pylint disable=bare-except

qubes/tools/qvm_start.py
kalkin: @marmarek No it shouldn't. Because it will end up there only if there is no exception thrown during any vm.start() calls.

marmarek: Ah, I see, `parser.error_runtime` calls `exit`.

marmarek:
Now it looks ok.

PR #52, opened 2016-08-28
Add qvm-block extend

kalkin:
- Added support for file
- Added support for lvm

PR #54, opened 2016-09-04
Volume add size property wrapper

kalkin:
Fixes regression introduced by #52


marmarek:
One more thing: `FileVolume.resize` fails for the same reason.
While at it - please add a test for resize method(s).

kalkin:
Tests are missing, I need to fix them first.

PR #56, opened 2016-09-12
Add qubes.signals for forwarding events

kalkin:
Signals are events which need to propagate to the "outer system" like D-Bus,
GuiDomain & Co. The path which an event takes from internals to the GUI looks
currently like this.

```
core-admin -> entry_point.event_receiver -> DbusReceiver -> DBUS-SYSTEM -> GUI
```

When forwarding qubes.events to the "outer system" we have to differentiate
between different kinds of events:

* All the events send before/after/including domain-init and domain-load,
    are only useful for core-admin and should be dropped.
* Events which should be propagated to the `event_receiver` as soon as
    possible, like domain-add/delete/is-fully-usable
* All the 'property-*' after domains and `Qubes()` is loaded need to be
    buffered as an array of events until `Qubes.save()` is called and only then be
    send to the event_receiver.

They word signal is just used to reduce the confusion between to internal core-admin
events and events handled like described above.

### DIFF
--- a/qubes/__init__.py
+++ b/qubes/__init__.py
@@ -41,6 +41,7 @@ import lxml.etree
 import qubes.config
 import qubes.events
 import qubes.exc
+import qubes.signals
 
 __author__ = 'Invisible Things Lab'
 __license__ = 'GPLv2 or later'
@@ -372,8 +373,9 @@ class property(object): # pylint: disable=redefined-builtin,invalid-name
         return bool(value)
 
 
-class PropertyHolder(qubes.events.Emitter):
-    '''Abstract class for holding :py:class:`qubes.property`
+class PropertyHolder(qubes.events.Emitter, qubes.signals.SignalEmitter):
+    '''Abstract class for holding :py:class:`qubes.property`. This class also
+       forwards all the events to the the signal receivers.
 
     Events fired by instances of this class:
 
@@ -630,6 +632,13 @@ class PropertyHolder(qubes.events.Emitter):
             else:
                 # pylint: disable=no-member
                 self.log.fatal(msg)
+
+    def fire_event(self, event, *args, **kwargs):
+        ''' Fires the event as signal and calls the inherited fire_event method
+        '''
+        self.fire_signal(event, *args, **kwargs)
+        return super(PropertyHolder, self).fire_event(event, *args, **kwargs)
+
 
 # pylint: disable=wrong-import-position
 from qubes.vm import VMProperty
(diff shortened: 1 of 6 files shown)


marmarek:
I think this is a bad idea in principle. D-bus signals should be sent as part of normal event handler - yes, before saving actual data in qubes.xml. If you need to reload qubes.xml, then just have another d-bus signal for this - `save` or sth like this. (currently there is no event for this, but there should be one)

Yes, this means that Qubes Manager will receive new label before it actually land in `qubes.xml`, but the same applies to all the handlers - like coloring icons in menu. It may also happen that `save` will never be called - for example because user hit Ctrl-C.
I propose to simply ignore this fact. In the future (4.1) we'll have qubesd daemon and there will be no more need for explicit `save()`. I propose to design an API with this idea in mind and not introduce mechanisms (involving API) which will be unneeded very soon.

kalkin:
> I think this is a bad idea in principle. D-bus signals should be sent as part of normal event handler - yes, before saving actual data in qubes.xml.

Don't confuse the `qubes.signals` with D-Bus signals. The issue is that the `qubes.events` are tight to the current executing process and are highly specific to the `core-admin` internals. I thought it would make much more sense to only notify about real changes in "Qubes state" and to differentiate them from `qubes.events`. I future `qubesd` could use a similar signal system to forward changes in the "Qubes state" to the GuiDomain.

> Yes, this means that Qubes Manager will receive new label before it actually land in qubes.xml, but the same applies to all the handlers - like coloring icons in menu. It may also happen that save will never be called - for example because user hit Ctrl-C.

Currently the dbus services and the ui are extremely brittle.  
There is no sane way checking the consistency besides checking regularly if they are out of sync of qubes.xml, similar to like QubesManager does.

kalkin:
Also if a D-Bus Service is out of sync and the ui shows some thing wrong, there is no obvious way for a user to fix it, by for example killing the UI-App. You would have to kill the D-Bus Service (which of course will be restarted on demand and reinitialise itself with up to date data).

woju:
Sure, but those `signals` are wrong solution to consistency problems. The correct solution would be to write `qubesd` and provide much harder guarantees (like "won't push the event to clients until actually committed the change to disk").

The consistency problems aren't really that big, because for most of the problems simple restart of the software in question is enough to roll back the change (appmenus are one important exception).

kalkin:
Ok I will reduce this to just forwarding the normal events to the Receiver

PR #57, opened 2016-09-19
PropertyHolder forwards events to event proxies

kalkin:


marmarek:
Why not simply add an extension which would register catch-all event handler?

kalkin:
@marmarek Do i see it right, that in such case we do not need any changes at all to `core-admin`? Basically qubes-dbus package just needs to provide the entry point `qubes.ext` with a system-wide handler?

kalkin:
Ok we really don't need any patching for `qubes-core-admin`. The `Extension` API seems to be enough. I kind of missed it :(

marmarek:
> @marmarek Do i see it right, that in such case we do not need any changes at all to core-admin? Basically qubes-dbus package just needs to provide the entry point qubes.ext with a system-wide handler?

Yes exactly.
I think the only missing part is registering catch-all. It's trivial patch, lets assume you need to handle `*` event.

kalkin:
@marmarek May i also add a general `property-set` hook? Currently i can only hook in to specific properties. I added a catch-all, but it's not as near as useful for `qubes-dbus` as a general catch all `property-set`.

marmarek:
what about having catch-all, which itself filter useful events? Event name is always given as an argument.

kalkin:
@marmarek OK

PR #61, opened 2016-09-26
Added qubes-label entry to qubesdb on VM start

kulinacs:
There have been a handful of requests for access to the Qube label
color from within the Qube itself for theming purposes. Adding this
entry to QubesDB will facilitate that. I do not believe this causes
a breach of security, as I feel the label should be treated the same
as the Qube name, which is also set in 000QubesVm.py.

### DIFF
--- a/core-modules/000QubesVm.py
+++ b/core-modules/000QubesVm.py
@@ -1095,6 +1095,7 @@ class QubesVm(object):
             return
 
         self.qdb.write("/name", self.name)
+        self.qdb.write("/qubes-label", self.label.name)
         self.qdb.write("/qubes-vm-type", self.type)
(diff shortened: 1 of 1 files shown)


marmarek:
This needs some more thinking. Exposing VM its own color may help with some attacks (trying to fool the user that window decoration is something else). Not sure how big the problem is.

marmarek:
@rootkovska what do you think about giving VM information of its own label color?

rootkovska:
We discussed it many times over the past years. This is a bad idea, because:
1. Some users might not want to reveal such info to the VM itself,
2. We don't want users to get used to VM-drawn decorations (this is what "theming purposes" would lead to).

PR #62, opened 2016-10-03
Add domain-shutdown event

kalkin:


marmarek:
This is far not the only situation when domain is shut down. For example `vm.kill` is not covered. But more importantly - domain shutting down itself (`sudo poweroff` from inside). For this we need some libvirt hook (I bet I've mentioned it somewhere already, it's available only in libvirt 2.2.0+).

kalkin:
Why does `QubesVM.shutdown` accepts the parameter `force`, but it's never actually used? Should it be removed?

kalkin:
Or may be the `force` paramter should only be used for the events, to differentiate between a kill and normal shutdown?

marmarek:
Implemented as libvirt even handler instead (https://github.com/marmarek/qubes-core-admin/commit/4f29e8ddba506aaaf94f71f581d1ad03ff5960b7)

PR #66, opened 2016-11-04
Core3 misc fixes

marmarek:
This fixes most of QubesOS/qubes-issues#2412, and few more.

### DIFF
--- a/doc/manpages/qubes-prefs.rst
+++ b/doc/manpages/qubes-prefs.rst
@@ -27,10 +27,6 @@ Options
 
    Decrease verbosity.
 
-.. option:: --force-root
-
-   Force to run as root.
-
 .. option:: --unset, --default, --delete, -D
 
    Unset the property. If is has default value, it will be used instead.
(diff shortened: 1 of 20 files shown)


marmarek:
Ok, so I need to recall what was broken by this (apparently intentional) inconsistency and fix that place instead...
For now, I'd rebase this branch skipping this commit (and also handling conflict with `qubes/app.py`). Do you have anything against?

woju:
On Fri, Nov 25, 2016 at 04:56:39AM -0800, Marek Marczykowski-Górecki wrote:
> Ok, so I need to recall what was broken by this (apparently intentional)
> inconsistency and fix that place instead...

I wouldn't call it "inconsistency", since those are really two different use
cases. One is to require an optional argument with default value (like `yes`
for example). The other is to require "some specific number of arguments" (like
`cp` requiring typically 2). The overlap of requiring exactly 1 argument is
a result of not prohibiting from specifying 1 argument.

> For now, I'd rebase this branch skipping this commit (and also handling
> conflict with `qubes/app.py`). Do you have anything against?

Not at all.

PR #70, opened 2016-11-15
Explicitly deny qubes.Filecopy to dom0 in RPC policy

jpouellet:
It already wasn't possible to qubes.Filecopy to dom0 because dom0
had no actual qubes.Filecopy service.

The introduction of this policy only improves the error message when
someone tries.

Fixes https://github.com/QubesOS/qubes-issues/issues/2031

### DIFF
--- a/qubes-rpc-policy/qubes.Filecopy.policy
+++ b/qubes-rpc-policy/qubes.Filecopy.policy
@@ -3,4 +3,5 @@
 
 ## Please use a single # to start your custom comments
 
+$anyvm	dom0	deny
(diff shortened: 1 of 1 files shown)


marmarek:
As explained in QubesOS/qubes-issues#2031 the problem was elsewhere.

jpouellet:
#### Note to self:
fix was: https://github.com/marmarek/qubes-core-admin-linux/commit/981a11cee14a10ad90a3f0c8803f956d102f9076
because: https://stackoverflow.com/questions/2209755/python-operation-vs-is-not

PR #77, opened 2016-11-25
Simplify redundant printing

jpouellet:
Wat...

### DIFF
--- a/qvm-tools/qvm-check
+++ b/qvm-tools/qvm-check
@@ -52,30 +52,30 @@ Specify no state options to check if VM exists"""
     vm = qvm_collection.get_vm_by_name(vmname)
     if vm is None:
         if options.verbose:
-            print >> sys.stdout, "A VM with the name '{0}' does not exist in the system!".format(vmname)
+            print "A VM with the name '{0}' does not exist in the system!".format(vmname)
         exit(1)
 
     elif options.running:
         vm_state = not vm.is_running()
         if options.verbose:
-            print >> sys.stdout, "A VM with the name {0} is {1}running.".format(vmname, "not " * vm_state)
+            print "A VM with the name {0} is {1}running.".format(vmname, "not " * vm_state)
         exit(vm_state)
 
     elif options.paused:
         vm_state = not vm.is_paused()
         if options.verbose:
-            print >> sys.stdout, "A VM with the name {0} is {1}paused.".format(vmname, "not " * vm_state)
+            print "A VM with the name {0} is {1}paused.".format(vmname, "not " * vm_state)
         exit(vm_state)
 
     elif options.template:
         vm_state = not vm.is_template()
         if options.verbose:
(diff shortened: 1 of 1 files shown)


marmarek:
This looks like the intention was to print on stderr...

jpouellet:
Indeed, but does that make sense? Using stdout here seems consistent with other tools to me.

marmarek:
> Indeed, but does that make sense? Using stdout here seems consistent with other tools to me.

Diagnostic messages should go to stderr. And it most qvm-tools it is the case. Yes, not all of them, but it's a problem of those tools. The only case when stdout should be used is when the main purpose of the tool is printing this information (like qvm-ls, qvm-prefs etc) - but still - only in scope of requested information, any extra messages (warnings etc) should go to stderr.

jpouellet:
On Fri, Nov 25, 2016 at 6:50 PM, Marek Marczykowski-Górecki <
notifications@github.com> wrote:

> Indeed, but does that make sense? Using stdout here seems consistent with
> other tools to me.
>
> Diagnostic messages should go to stderr. And it most qvm-tools it is the
> case. Yes, not all of them, but it's a problem of those tools. The only
> case when stdout should be used is when the main purpose of the tool is
> printing this information (like qvm-ls, qvm-prefs etc) - but still - only
> in scope of requested information, any extra messages (warnings etc) should
> go to stderr.
>
I understand that, and it appears to me that the information being printed
here is exactly the output the tool is invoked to obtain.

I could see an argument for using stderr if the intended invocation is from
some script to use this as quick and dirty state-validation before trying
to perform some operation. It is clear to me whether this is the case or
not. The only use without --quiet by default is in mgmt-salt-dom0-qvm, and
that is just a generic wrapper without an opinion (that I see) on
stdout/err.

This is really a tiny issue either way and not something I feel strongly
about (much bigger problems to solve...) just struck me as quite odd.

Shall I switch this to `>> sys.stderr` instead?

jpouellet:
> Diagnostic messages should go to stderr. The only case when stdout should be used is when the main purpose of the tool is printing this information, any extra messages (warnings etc) should go to stderr.

I understand that, and it appears to me that the information being printed
here is exactly the output the tool is invoked to obtain.

I could see an argument for using stderr if the intended invocation is from
some script to use this as quick and dirty state-validation before trying
to perform some operation. It is clear to me whether this is the case or
not. The only use without --quiet by default is in mgmt-salt-dom0-qvm, and
that is just a generic wrapper without an opinion (that I see) on
stdout/err.

This is really a tiny issue either way and not something I feel strongly
about (much bigger problems to solve...) just struck me as quite odd.

Shall I switch this to `>> sys.stderr` instead?

jpouellet:
> Shall I switch this to >> sys.stderr instead?

Upon thinking about the likely invocations of this script some more, I think stderr indeed seems more appropriate.

PR doing so at: https://github.com/QubesOS/qubes-core-admin/pull/80

If you merge one, just close the other.

marmarek:
Ok, merged the other one, thanks!

PR #78, opened 2016-11-27
Add raw_name field for qvm-ls

jpouellet:
Useful to avoid needing to `... | tr -d '<>{}[]='` just to get clean VM names in scripts.

Fixes https://github.com/QubesOS/qubes-issues/issues/2444

Example usage:
```
[user@dom0 ~]$ qvm-ls --raw-data raw_name state | grep fedora
fedora-23|Halted
fedora-23-dvm|Halted
fedora-24-minimal|Halted
fedora-24|Halted
```

### DIFF
--- a/qvm-tools/qvm-ls
+++ b/qvm-tools/qvm-ls
@@ -41,6 +41,8 @@ fields = {
              + ('>' if vm.is_disposablevm() else '')\
              + ('}' if vm.is_netvm() else '')"},
 
+    "name-raw": {"func": "vm.name"},
+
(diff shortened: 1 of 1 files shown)


jpouellet:
@marmarek or should it be `raw-name` instead of `raw_name`?

See https://github.com/QubesOS/qubes-core-admin/pull/79 for list of current fields (we have dash and space already, but no underscore)

jpouellet:
Or maybe `name-raw` so it sorts next to `name`?

marmarek:
`name-raw` would be better.

jpouellet:
Done.

PR #83, opened 2016-12-04
Update qvm-block man page to match help output

andrewdavidwong:


andrewdavidwong:
I could be wrong, but that Travis error doesn't appear to be related to the changes I've submitted.

marmarek:
Yes, something changed in travis system image and broke (almost?) all our components. Fix already included in qubes-builder, but change in every `.travis.yml` is required too (removing that `ln` command).

PR #89, opened 2017-02-21
Core3 py3k fixes

marmarek:
- Most of it is about str/bytes, but there are also other fixes.
- Backups are broken (because of reverted essential part of it).
- The last commit is unrelated - it extends app.add_new_vm function to accept also VM class name, not necessary class itself (save one function call for API users). @woju If you really don't like it, merge without that last commit.


qubes/__init__.py
```diff
@@ -299,6 +299,11 @@ def __repr__(self):
     def __hash__(self):
         return hash(self.__name__)
 
+    def __lt__(self, other):
+        if isinstance(other, property):
+            return self.__name__ < other.__name__
+        else:
+            return self < other
```
marmarek: So, maybe instead of `sorted(..., key=)` in `property_list()`, this operator should be defined (using the algorithm from `property_list()`)?

woju: That may be a good solution.

marmarek:
@woju Addressed your comments. Do you want me to rebase it, or you'll handle the conflict?

marmarek:
Rebased on core3-devel branch.

PR #91, opened 2017-03-14
qvm-shutdown: Do not mutate list while iterating through it

m-v-b:
This commit makes sure that the Python list vms_list is not mutated
while the code is iterating through it. To the best of my knowledge,
this is a problematic operation.

To rectify this issue, a new temporary list is instantiated, and
the VM objects that have shut down are appended to the temporary list,
which is afterwards used to remove the shut-down VM objects from the
vms_list.

### DIFF
--- a/qvm-tools/qvm-shutdown
+++ b/qvm-tools/qvm-shutdown
@@ -101,22 +101,31 @@ def main():
         while len (vms_list):
             if options.verbose:
                 print >> sys.stderr, "Waiting for VMs: ", [vm.name for vm in vms_list]
+
+            shut_down_vms = []
+
             for vm in vms_list:
                 if not vm.is_running():
-                    vms_list.remove (vm)
+                    shut_down_vms.append(vm)
                     continue
+
                 if vm.get_power_state() == "Halting":
                     if vm in halting_vms:
                         vm.force_shutdown()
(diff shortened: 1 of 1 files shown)


marmarek:
Do you have your code signing key uploaded to some keyserver?

m-v-b:
Hello @marmarek,

Not until very recently. I uploaded my public key to the SKS key server pool only a few hours ago. (I also verified afterwards that the Ubuntu key server has my keys. [1]) This is most likely why all of my pull requests have failed their code signature verifications.

Sorry about this! Is there a way to re-trigger the Travis continuous integration tasks?

[1] https://keyserver.ubuntu.com/pks/lookup?op=vindex&search=0x1DF87CE3B3A5DFAF&fingerprint=on

marmarek:
There is no public way to trigger re-check of signature (other than pushing another commits). Just triggered it manually.
As for Travis - please rebase on current release3.2 (or better: master) branch, to include fixed `.travis.yml`.

m-v-b:
Hmm. I made a mistake. I should not re-use the same pull request while rebasing onto master! About to re-push.

m-v-b:
Ah, I forgot the signature... Sorry for the noise.

PR #92, opened 2017-03-14
qubes-core.service: Do not shut down/reboot before all VMs have shut …

m-v-b:
…down

Prior to this commit, the qubes-core.service inherited systemd's default
timeout value, which most of the time caused the dom0 shut down sequence
to proceed even if some VMs were still not fully shut down at the time of
dom0 shut down.

This commit corrects this issue by setting the service stop timeout to
infinity.

### DIFF
--- a/linux/systemd/qubes-core.service
+++ b/linux/systemd/qubes-core.service
@@ -7,6 +7,8 @@ After=qubes-db-dom0.service libvirtd.service xenconsoled.service
 Type=oneshot
 StandardOutput=syslog
 RemainAfterExit=yes
+# Needed to avoid rebooting before all VMs have shut down.
+TimeoutStopSec=180
 ExecStart=/usr/lib/qubes/startup-misc.sh
 ExecStop=/usr/bin/qvm-shutdown -q --all --wait
 # QubesDB daemons stop after 60s timeout in worst case; speed it up, since no


linux/systemd/qubes-core.service
```diff
@@ -7,6 +7,8 @@ After=qubes-db-dom0.service libvirtd.service xenconsoled.service
 Type=oneshot
 StandardOutput=syslog
 RemainAfterExit=yes
+# Needed to avoid rebooting before all VMs have shut down.
+TimeoutStopSec=0
```
marmarek: Not sure is this is a good idea. Maybe better simply increase the timeout over default 90s. For example 180s, or 240s.

m-v-b:
> Not sure is this is a good idea. Maybe better simply increase the timeout over default 90s. For example 180s, or 240s.

After thinking about it a bit more, I agree with you. Preventing the completion of the shutdown/reboot sequence in case this particular service hangs would be undesirable.

I had originally used this change to accommodate my laptop's slow hard disk drive, but I have since switched to a solid state drive a few months ago, so I don't think I can reproduce this issue any longer. I was apparently not the only one who had this issue with slow hard disks. [1] [2]

Perhaps setting this to 180 or 240 seconds would be appropriate, as you have suggested. To err on the safer side in terms of not preventing shutdown/reboot for too long, I think I will pick 180 seconds.

[1] https://groups.google.com/d/msg/qubes-users/UZoK6Cs1Grw/o587K5kfBgAJ
[2] https://github.com/QubesOS/qubes-issues/issues/2245

PR #93, opened 2017-03-14
qvm-shutdown: Allow multiple VMs to be shut down with one invocation

m-v-b:
Prior to this commit, qvm-shutdown had a limitation where only one
VM name could be passed in via the command line for shutting down.

This commit removes the aforementioned limitation by adapting the
code for multiple command line arguments.

### DIFF
--- a/doc/qvm-tools/qvm-shutdown.rst
+++ b/doc/qvm-tools/qvm-shutdown.rst
@@ -6,8 +6,6 @@ NAME
 ====
 qvm-shutdown
 
-:Date:   2012-04-11
-
 SYNOPSIS
 ========
 | qvm-shutdown [options] <vm-name> [vm-name ...]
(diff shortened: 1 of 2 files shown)


jpouellet:
Don't forget about the usage msg, man page, and qvm dom0 tools page in docs

m-v-b:
Ah, good point.

m-v-b:
I forgot to update the date in the manual page, and I have not yet created a pull request for the documentation update after realizing this.

Even though the following may sound trivial, are the dates in the manual pages important? Depending on the answer, I can prepare commits to remove the dates, or just update the one for qvm-shutdown in this pull request.

It is also possible that the main change in this pull request is only a minor improvement and is unwarranted. I will wait for further feedback before continuing.

marmarek:
As you see, we don't update those dates and I don't think that would change anytime soon. So, better remove them completely. Or generate automatically, based on last commit to that file. But probably not worth it.

m-v-b:
Sounds good. I prepared another commit, only for qvm-shutdown to avoid merge conflicts. I will create another pull request for the remaining utilities and yet another one for the `qubes-doc` repository.

PR #95, opened 2017-03-17
Core3 mgmt methods

marmarek:
* Labels
* Storage
* `vm.property.Set`

And related changes.
/cc @woju

### DIFF
--- a/qubes/mgmt.py
+++ b/qubes/mgmt.py
@@ -176,43 +176,7 @@ class QubesMgmt(object):
         assert self.arg in self.dest.property_list()
 
         property_def = self.dest.property_get_def(self.arg)
-        # do not treat type='str' as sufficient validation
-        if isinstance(property_def, qubes.vm.VMProperty):
-            try:
-                untrusted_vmname = untrusted_payload.decode('ascii')
-            except UnicodeDecodeError:
-                raise qubes.exc.QubesValueError
-            qubes.vm.qubesvm.validate_name(self.dest, property_def,
-                untrusted_vmname)
-            try:
-                newvalue = self.app.domains[untrusted_vmname]
-            except KeyError:
-                raise qubes.exc.QubesPropertyValueError(
-                    self.dest, property_def, untrusted_vmname)
-        elif property_def.type is not None and property_def.type is not str:
-            if property_def.type is bool:
-                try:
-                    untrusted_value = untrusted_payload.decode('ascii')
-                except UnicodeDecodeError:
-                    raise qubes.exc.QubesValueError
-                newvalue = qubes.property.bool(None, None, untrusted_value)
-            else:
-                try:
-                    newvalue = property_def.type(untrusted_payload)
-                except ValueError:
-                    raise qubes.exc.QubesValueError
-        else:
-            # other types (including explicitly defined 'str') coerce to str
-            # with limited characters allowed
-            try:
-                untrusted_value = untrusted_payload.decode('ascii')
-            except UnicodeDecodeError:
-                raise qubes.exc.QubesValueError
-            allowed_set = string.printable
-            if not all(x in allowed_set for x in untrusted_value):
-                raise qubes.exc.QubesValueError(
-                    'Invalid characters in property value')
-            newvalue = untrusted_value
+        newvalue = property_def.sanitize(untrusted_newvalue=untrusted_payload)
 
         self.fire_event_for_permission(newvalue=newvalue)
 
--- a/qubes/mgmt.py
+++ b/qubes/mgmt.py
@@ -132,12 +132,14 @@ class QubesMgmt(object):
 
     @asyncio.coroutine
     def vm_list(self, untrusted_payload):
-        assert self.dest.name == 'dom0'
         assert not self.arg
         assert not untrusted_payload
         del untrusted_payload
 
-        domains = self.fire_event_for_filter(self.app.domains)
+        if self.dest.name == 'dom0':
+            domains = self.fire_event_for_filter(self.app.domains)
+        else:
+            domains = self.fire_event_for_filter([self.dest])
 
         return ''.join('{} class={} state={}\n'.format(
                 vm.name,
--- a/qubes/mgmt.py
+++ b/qubes/mgmt.py
@@ -165,14 +165,28 @@ class QubesMgmt(object):
 
         self.fire_event_for_permission()
 
+        property_def = self.dest.property_get_def(self.arg)
+        # explicit list to be sure that it matches protocol spec
+        if isinstance(property_def, qubes.vm.VMProperty):
+            property_type = 'vm'
+        elif property_def.type is int:
+            property_type = 'int'
+        elif property_def.type is bool:
+            property_type = 'bool'
+        elif self.arg == 'label':
+            property_type = 'label'
+        else:
+            property_type = 'str'
+
         try:
             value = getattr(self.dest, self.arg)
         except AttributeError:
-            return 'default=True '
+            return 'default=True type={} '.format(property_type)
         else:
-            return 'default={} {}'.format(
+            return 'default={} type={} {}'.format(
                 str(self.dest.property_is_default(self.arg)),
-                self.repr(value))
+                property_type,
+                str(value) if value is not None else '')
 
     @asyncio.coroutine
     def vm_property_help(self, untrusted_payload):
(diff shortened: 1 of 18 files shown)


qubes/mgmt.py
```diff
@@ -186,11 +234,239 @@ def vm_property_help(self, untrusted_payload):
...
+            'save_on_stop', 'snap_on_start']
+        return ''.join('{}={}\n'.format(key, getattr(volume, key)) for key in
+            volume_properties)
+
+    @asyncio.coroutine
+    @no_payload
+    def vm_volume_listsnapshots(self):
+        assert self.arg in self.dest.volumes.keys()
+
+        self.fire_event_for_permission()
```
woju: Why not ``fire_event_for_filter()``?

qubes/mgmt.py
```diff
@@ -186,11 +234,239 @@ def vm_property_help(self, untrusted_payload):
...
+        self.dest.storage.get_pool(volume).revert(revision)
+        self.app.save()
+
+    @asyncio.coroutine
+    def vm_volume_resize(self, untrusted_payload):
+        assert self.arg in self.dest.volumes.keys()
+        untrusted_size = untrusted_payload.decode('ascii').strip()
+        del untrusted_payload
+        assert untrusted_size.isdigit()  # only digits, forbid '-' too
+        assert len(untrusted_size) <= 20  # limit to about 2^64
```
woju: What's the unit?

qubes/mgmt.py
```diff
@@ -186,11 +234,239 @@ def vm_property_help(self, untrusted_payload):
...
+        pool_name = untrusted_pool_name
+        assert pool_name not in self.app.pools
+
+        driver_parameters = qubes.storage.driver_parameters(self.arg)
+        assert all(key in driver_parameters for key in untrusted_pool_config)
+
+        # option names validated, validation of option values is delegated to
+        #  extension (through events mechanism)
+        self.fire_event_for_permission(name=pool_name,
+            untrusted_pool_config=untrusted_pool_config)
```
woju: I thought we don't pass untrusted values to events. That was my understanding. If the pool config is trusted enough to be passed to events, shouldn't we drop the ``untrusted_`` prefix?

qubes/mgmt.py
```diff
@@ -186,11 +234,239 @@ def vm_property_help(self, untrusted_payload):
...
+            raise qubes.exc.QubesValueError('label already exists')
+
+        untrusted_payload = untrusted_payload.decode('ascii').strip()
+        assert len(untrusted_payload) == 8
+        assert untrusted_payload.startswith('0x')
+        # besides prefix, only hex digits are allowed
+        assert all(x in string.hexdigits for x in untrusted_payload[2:])
+
+        # TODO: try to avoid creating label too similar to existing one?
+        color = untrusted_payload
```
woju:
```python
import math

GAMMA = 2.2
#THRESHOLD = 

def hex_to_float(color, gamma=1.0):
  return tuple((int(color[i:i+2], 0x10) / 255.0) ** gamma for i in range(0, len(color), 2))

untrusted_r, untrusted_g, untrusted_b = hex_to_float(untrusted_color, gamma=GAMMA)
for label in existing_labels:
  r, g, b = hex_to_float(label.color, gamma=GAMMA)
  # approximate the colour distance as Euclidean distance in linear (gamma-expanded) space
  if math.sqrt((r - untrusted_r) ** 2 + (g - untrusted_g) ** 2 + (b - untrusted_b) ** 2) < THRESHOLD:
    raise qubes.exc.QubesValueError('colour too similar to an existing one')
```

Unfortunately the threshold value should be determined empiricaly. Would you like it as patch, after experimentation? Maybe as part of some enterprise-y extension?

qubes/mgmt.py
marmarek: Good idea.

qubes/mgmt.py
marmarek: bytes of course ;)

qubes/mgmt.py
marmarek: Definitely an extension.

qubes/mgmt.py
marmarek: This is after preliminary validation, so indeed dropping `untrusted_` prefix realier is a good idea.
I have no idea how to validate actual pool configuration, but I think we can assume that if one is granted storage pool creation permission, by design is also granted arbitrary file access on the disk. If not - such policy should be implemented by extension handling appropriate mgmt-* event.

qubes/mgmt.py
```diff
@@ -186,11 +234,239 @@ def vm_property_help(self, untrusted_payload):
...
+    def vm_volume_info(self):
+        assert self.arg in self.dest.volumes.keys()
+
+        self.fire_event_for_permission()
+
+        volume = self.dest.volumes[self.arg]
+        # properties defined in API
+        volume_properties = [
+            'pool', 'vid', 'size', 'usage', 'rw', 'internal', 'source',
+            'save_on_stop', 'snap_on_start']
```
marmarek: Those all are read-only properties, see our discussion about builtins.property listing.

qubes/mgmt.py
woju: Sure. In that case, we can file a bug to remember, paste the snippet and remove TODO.

qubes/mgmt.py
```diff
@@ -144,37 +133,96 @@ def vm_list(self, untrusted_payload):
...
+            # with limited characters allowed
+            try:
+                untrusted_value = untrusted_payload.decode('ascii')
+            except UnicodeDecodeError:
+                raise qubes.exc.QubesValueError
+            allowed_set = string.printable
+            if not all(x in allowed_set for x in untrusted_value):
+                raise qubes.exc.QubesValueError(
+                    'Invalid characters in property value')
+            newvalue = untrusted_value
```
marmarek: Not sure about that - I'd like to have all the untrusted-input validation code in one place (file), for easier auditing. If we spread that across multiple files (imagine searching for all qubes.property subclasses), that would be much harder to audit.

Maybe such validator should be _in addition_ to preliminary validation here? But on the other hand, we already have setters which are serving mostly this purpose.

qubes/mgmt.py
woju: This is dangerously close to current filesystem limits. For example ext4 has a maximum volume size limit of 1EiB = 2^60 B. ``len(str(2**60)) == 19``.

qubes/mgmt.py
woju: Setters were not designed for malicious input (only accidentally wrong), so I'm not suggesting using them. But maybe this code can be put there, reviewed for malicious input, and be good enough?

I sympathise with the idea of "all untrusted validation in one place", but that resulted in spaghetti function you just did.

marmarek: Maybe a mixin for qubes.property, defined in this file? And some mechanism to make sure that if validator is even overridden, it's also in this file? Maybe with some pylint-like script?

jpouellet: I have not yet started reading core3 in earnest, so my apologies if this comment is uninformed, but it would indeed be nice to override properties & change how they are validated.

One example use case can be found in @Rudd-O's [qubes-network-server](https://github.com/Rudd-O/qubes-network-server) where it would be really nice to have a clean way to override the validation function for firewall rules. Not allowing properties to be extended cleanly in this case results in the need to duplicate qvm-prefs and qvm-firewall.

Granted, the whole firewall system is changing anyway so this specific case would no longer apply, but IMO this still serves to illustrate that legitimate use cases exist.

(also firewall rules are not exactly untrusted input... but still)

woju: @jpouellet Yes, this is already done. May I point you in those directions:
- https://qubes-core-admin.readthedocs.io/en/latest/qubes.html#properties
- https://github.com/QubesOS/qubes-core-admin/blob/3e0f225/qubes/__init__.py#L130
- https://github.com/QubesOS/qubes-core-admin/blob/3e0f225/qubes/vm/qubesvm.py#L355-L489 with specific focus on `setter=`

The question is, should we provide some additional (coarse) validation to all properties that cannot be overridden with specific intent to eliminate some basic attacks, possibly encoding-oriented. The problem is, when I wrote `setter=`, I thought it would be used from dom0 and we wouldn't supply it with untrusted data (possibly also malformed). If we'd like to repurpose that, we should look at it once again, rename function signatures to `untrusted_*` (possibly enforcing the name as keyword-only argument, since we have python3).

PR #97, opened 2017-04-07
Core3 policy

marmarek:
Implementation of QubesOS/qubes-issues#910

This does not fully work, but because of problems unrelated to this PR - VM startup fails. And I don't want to mix too many unrelated changes in one pull request.
TODO: modify qrexec-daemon to really use new qrexec-policy (now in /usr/bin/qrexec-policy, instead of /usr/lib/qubes/qrexec-policy).

/cc @woju

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ install:
   - pip install --quiet -r ci/requirements.txt
   - git clone https://github.com/"${TRAVIS_REPO_SLUG%%/*}"/qubes-builder ~/qubes-builder
 script:
-  - PYTHONPATH=test-packages pylint --rcfile=ci/pylintrc qubes
+  - PYTHONPATH=test-packages pylint --rcfile=ci/pylintrc qubes qubespolicy
   - ./run-tests --no-syslog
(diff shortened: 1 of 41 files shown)


jpouellet:
Is dom0 communicating with the adminvm/guivm over dbus something we want?

Seems kind-of hypocritical to [complain about OpenXT doing that](https://twitter.com/rootkovska/status/505024924097196032) and then do it ourselves. Yes, an adminvm/guivm must be somewhat trusted regardless, but if I understand correctly the implications of https://github.com/QubesOS/qubes-issues/issues/867#issuecomment-268389632, it seems we still have a reason to want to protect the integrity of dom0 from the adminvm/guivm (otherwise "you can do whatever to your own appvms, but policy is still enforced on these specific ones" can not be implemented in meaningfully secure way).

marmarek:
> Is dom0 communicating with the adminvm/guivm over dbus something we want?

No, in 4.0 GUI is still in dom0. And this is similarly simple than setting `DISPLAY=:0` in daemon process, but much more elegant. When moving GUI out of dom0, this will be converted to qrexec service.

PR #98, opened 2017-04-10
Device assignments API

kalkin:
Main changes the the device API:
- `attach`/`detach` expect `DeviceAssignment` as parameter
- `attached(persistent=True)` is now persistent()
- `attached()` returns all attached devices
- `assigned()` returns all device assignments.

I haven't yet tested on `fc25`

### DIFF
--- a/qubes/devices.py
+++ b/qubes/devices.py
@@ -195,7 +195,11 @@ class DeviceCollection(object):
 
     def attached(self):
         '''List devices which are (or may be) attached to this vm '''
-        return self._vm.fire_event('device-list-attached:' + self._class) or []
+        attached = self._vm.fire_event('device-list-attached:' + self._class)
+        if attached:
+            return [dev for dev, _ in attached]
+
+        return []
 
     def persistent(self):
         ''' Devices persistently attached and safe to access before libvirt
@@ -218,7 +222,7 @@ class DeviceCollection(object):
         devices = self._vm.fire_event('device-list-attached:' + self._class,
             persistent=persistent)
         result = []
-        for dev in devices:
+        for dev, options in devices:
             if dev in self._set and persistent is False:
                 continue
             elif dev in self._set:
@@ -228,7 +232,7 @@ class DeviceCollection(object):
             else:
                 result.append(
                     DeviceAssignment(backend_domain=dev.backend_domain,
-                                     ident=dev.ident, options=dev.options,
+                                     ident=dev.ident, options=options,
                                      frontend_domain=self._vm))
         if persistent is not False and len(result) == 0:
             result.extend(self._set)
--- a/qubes/devices.py
+++ b/qubes/devices.py
@@ -234,7 +234,7 @@ class DeviceCollection(object):
                     DeviceAssignment(backend_domain=dev.backend_domain,
                                      ident=dev.ident, options=options,
                                      frontend_domain=self._vm))
-        if persistent is not False and len(result) == 0:
+        if persistent is not False and not result:
             result.extend(self._set)
         return result
 
--- a/qubes/devices.py
+++ b/qubes/devices.py
@@ -47,7 +47,6 @@ Such extension should provide:
 
 import qubes.utils
 
-
 class DeviceNotAttached(qubes.exc.QubesException, KeyError):
     '''Trying to detach not attached device'''
     pass
@@ -327,3 +326,52 @@ class BlockDevice(object):
         self.domain = domain
         self.devtype = devtype
 
+class PersistentCollection(object):
+
+    ''' Helper object managing persistent `DeviceAssignment`s.
+    '''
+
+    def __init__(self):
+        self._dict = {}
+
+    def add(self, assignment: DeviceAssignment):
+        ''' Add assignment to collection '''
+        assert assignment.persistent and assignment.frontend_domain
+        vm = assignment.backend_domain
+        ident = assignment.ident
+        key = (vm, ident)
+        assert key not in self._dict
+
+        self._dict[key] = assignment
+
+    def discard(self, assignment):
+        ''' Discard assignment from collection '''
+        assert assignment.persistent and assignment.frontend_domain
+        vm = assignment.backend_domain
+        ident = assignment.ident
+        key = (vm, ident)
+        if key not in self._dict:
+            raise KeyError
+        del self._dict[key]
+
+    def __contains__(self, device) -> bool:
+        vm = device.backend_domain
+        ident = device.ident
+        key = (vm, ident)
+        return key in self._dict
+
+    def get(self, device: DeviceInfo) -> DeviceAssignment:
+        ''' Returns the corresponding `qubes.devices.DeviceAssignment` for the
+            device. '''
+        vm = device.backend_domain
+        ident = device.ident
+        key = (vm, ident)
+        if key not in self._dict:
+            raise KeyError
+        return self._dict[key]
+
+    def __iter__(self):
+        return self._dict.values().__iter__()
+
+    def __len__(self) -> int:
+        return len(self._dict.keys())
--- a/qubes/devices.py
+++ b/qubes/devices.py
@@ -292,7 +292,7 @@ class DeviceInfo(object):
     ''' Holds all information about a device '''
     # pylint: disable=too-few-public-methods
     def __init__(self, backend_domain, ident, description=None,
-                 frontend_domain=None, options = None, **kwargs):
+                 frontend_domain=None, options=None, **kwargs):
         #: domain providing this device
         self.backend_domain = backend_domain
         #: device identifier (unique for given domain and device type)
@@ -308,6 +308,7 @@ class DeviceInfo(object):
             self.frontend_domain = frontend_domain
         except AttributeError:
             pass
+        self.options = options or dict()
         self.data = kwargs
 
         if hasattr(self, 'regex'):
--- a/qubes/devices.py
+++ b/qubes/devices.py
@@ -55,6 +55,10 @@ class DeviceAlreadyAttached(qubes.exc.QubesException, KeyError):
     '''Trying to attach already attached device'''
     pass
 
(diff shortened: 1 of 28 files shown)


kalkin:
I'm not sure if I should fix the few remaining failing tests  or if they don't make any sense anymore and can be removed.

marmarek:
I haven't reviewed it yet, but probably better to fix those tests.

Hmm, why it can't find your signing key? Have you uploaded it to any keyserver?

kalkin:
@marmarek 

> I haven't reviewed it yet, but probably better to fix those tests.

I haven't fixed them because I don't fully understand why exactly they are testing this events.

> Hmm, why it can't find your signing key? Have you uploaded it to any keyserver?

Nop my bad, I thought updating here on GitHub will be sufficient for the script. Will do that.

marmarek:
> I haven't fixed them because I don't fully understand why exactly they are testing this events.

AFAIR those tests were to test if API works as intended, including calling to extensions when necessary, _and only then_. If it's now expected to call those events, adjust the tests.

qubes/devices.py
```diff
@@ -115,87 +142,113 @@ class DeviceCollection(object):
...
+                continue
+            elif dev in self._set:
+                result.append(self._set.get(dev))
+            elif dev not in self._set and persistent is True:
+                continue
+            else:
+                result.append(
+                    DeviceAssignment(backend_domain=dev.backend_domain,
+                                     ident=dev.ident, options=dev.options,
+                                     frontend_domain=self._vm))
```
marmarek: Using `dev.options` here looks strange - options are property of assignment, not device itself. Better would be to expect event `device-list-attached` to return a tuples (device_info, options).

qubes/devices.py
```diff
@@ -289,8 +344,11 @@ def __init__(self, backend_domain, ident, description=None,
         super(UnknownDevice, self).__init__(backend_domain, ident, description,
             frontend_domain, **kwargs)
 
-
 class BlockDevice(object):
+    ''' Represents a storage block device
+
+        TODO: Move to qubes.storage?
```
marmarek: Yes, definitely.

qubes/ext/pci.py
```diff
@@ -19,10 +19,12 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+''' Qubes PCI Extensions '''
+
 import os
 import re
 import subprocess
-import libvirt
+import libvirt  # pylint:disable=import-error
```
marmarek: Take a look at pylint call in `.travis.yml`. There is mock `libvirt` module in `test-packages` directory exactly for this purpose. Anyway this pylint disable can be left here.

qubes/ext/pci.py
```diff
@@ -227,6 +228,9 @@ def on_device_pre_attached_pci(self, vm, event, device):
             return
 
         try:
+            device = next(
+                self.on_device_get_pci(vm, event, device.ident)
+            )
```
marmarek: I wonder whether more readable would be to put `PCIDevice(vm, device.ident)` here directly... And since this is repeated multiple times below, definitely it will be less code (in characters ;) ) to duplicate.

qubes/ext/pci.py
```diff
@@ -246,9 +250,12 @@ def on_device_pre_detached_pci(self, vm, event, device):
         # provision in libvirt for extracting device-side BDF; we need it for
         # qubes.DetachPciDevice, which unbinds driver, not to oops the kernel
 
+        device = next(
+            self.on_device_get_pci(vm, event, device.ident)
+        )
```
marmarek: And here.

qubes/ext/pci.py
```diff
@@ -270,8 +277,10 @@ def on_device_pre_detached_pci(self, vm, event, device):
     @qubes.ext.handler('domain-pre-start')
     def on_domain_pre_start(self, vm, _event, **_kwargs):
         # Bind pci devices to pciback driver
-        for pci in vm.devices['pci'].attached():
-            self.bind_pci_to_pciback(vm.app, pci)
+        for assignment in vm.devices['pci'].persistent():
+            device = next(
+                self.on_device_get_pci(vm, _event, assignment.ident))
```
marmarek: And here.

qubes/devices.py
```diff
@@ -115,87 +142,113 @@ class DeviceCollection(object):
...
         :param DeviceInfo device: device object
         '''
 
-        if device not in self.attached():
+        if not device_assignment.frontend_domain:
+            device_assignment.frontend_domain = self._vm
+
+        if device_assignment in self._set and not self._vm.is_halted():
+            raise WrongAssignment(
+                "Can not remove a persistent attachment from a non halted vm")
```
kalkin: AFAIR we agreed upon it. I think our motivation was to not being able to remove a devices which is needed for vm functioning. 

I don't mind to allow it.

marmarek: Ok, lets leave it this way for now.

qubes/devices.py
```diff
@@ -56,6 +55,34 @@ class DeviceAlreadyAttached(qubes.exc.QubesException, KeyError):
     '''Trying to attach already attached device'''
     pass
 
+class WrongAssignment(qubes.exc.QubesException, KeyError):
+    '''Trying to attach non permanent assignment to a halted vm'''
```
kalkin: I don't see a need for this exception any more. 2 of 3 use cases can be replaced by `qubes.exc.QubesVMNotHaltedError` & `qubes.exc.QubesVMNotRunningError` and the last one with an assert.

I will remove it

qubes/devices.py
```diff
@@ -56,6 +55,34 @@ class DeviceAlreadyAttached(qubes.exc.QubesException, KeyError):
...
+
+    def __repr__(self):
+        return "[%s]:%s" % (self.backend_domain, self.ident)
+
+    def __hash__(self):
+        return hash((self.backend_domain, self.ident))
+
+    def __eq__(self, other):
+        if not isinstance(self, other.__class__):
+            return False
```
kalkin: Agree

qubes/ext/pci.py
kalkin: Wouldn't `on_device_get_pci()` be a good place for caching the device objects?

marmarek: If so, [dropping `dev.options`](https://github.com/QubesOS/qubes-core-admin/pull/98#discussion_r111396748) is even more important.

qubes/devices.py
```diff
@@ -57,6 +56,30 @@ class DeviceAlreadyAttached(qubes.exc.QubesException, KeyError):
...
+
+    def __repr__(self):
+        return "[%s]:%s" % (self.backend_domain, self.ident)
+
+    def __hash__(self):
+        return hash((self.backend_domain, self.ident))
+
+    def __eq__(self, other):
+        if not isinstance(self, other.__class__):
+            raise NotImplementedError
```
woju: This is still wrong. ``raise NotImplementedError()`` != ``return NotImplemented``.

https://docs.python.org/3/library/constants.html#NotImplemented

qubes/devices.py
```diff
@@ -115,87 +138,117 @@ class DeviceCollection(object):
...
-        for device in seen:
-            # get fresh object - may contain updated information
-            device = device.backend_domain.devices[self._class][device.ident]
-            yield device
+
+        devices = self._vm.fire_event('device-list-attached:' + self._class,
+            persistent=persistent)
+        result = []
+        for dev, options in devices:
+            if dev in self._set and persistent is False:
```
woju: Booleans should not be checked by comparing identity to ``bool()`` instances, but ``... and persistent`` or `` and not persistent``. What happens if someone uses ``0`` or ``1`` instead of ``True`` and ``False`` for ``persistent``, maybe because they freshly came from linux/driver background and only briefly engage with Python?

qubes/devices.py
```diff
@@ -290,15 +345,52 @@ def __init__(self, backend_domain, ident, description=None,
...
+        key = (vm, ident)
+        if key not in self._dict:
+            raise KeyError
+        del self._dict[key]
+
+    def __contains__(self, device) -> bool:
+        vm = device.backend_domain
+        ident = device.ident
+        key = (vm, ident)
+        return key in self._dict
```
woju: ``return (device.backend_domain, device.indent) in self._dict``? Why verbosity?

qubes/devices.py
```diff
@@ -290,15 +345,52 @@ def __init__(self, backend_domain, ident, description=None,
...
+
+    def get(self, device: DeviceInfo) -> DeviceAssignment:
+        ''' Returns the corresponding `qubes.devices.DeviceAssignment` for the
+            device. '''
+        vm = device.backend_domain
+        ident = device.ident
+        key = (vm, ident)
```
woju: Same question.

qubes/devices.py
marmarek: Hmm, shouldn't pylint warn about this kind of things?

And, fixed in my working tree.

qubes/devices.py
marmarek: Already fixed here: https://github.com/marmarek/qubes-core-admin/commit/f1e79338515ce65101271a70d3ccdad615fb8e9e

qubes/devices.py
marmarek: I'll add fix for this too.

qubes/devices.py
marmarek: I'll add fix for this too.

qubes/devices.py
woju: Ok. Thanks.

qubes/devices.py
woju: I don't know if pylint should. In python<3 you could assign something to ``False`` or even do something like:
```python
True, False = False, True
```
and it would do what you think.

If it is fixed, I think that's OK.

woju:
Thanks for the fixes.

PR #100, opened 2017-04-21
Some more mgmt methods

marmarek:
- features
- events
- handle read-only properties (were standard `@property`)
- minor changes/fixes needed to handle gui-daemon startup at mgmt-client side


qubes/mgmt.py
```diff
@@ -534,3 +570,95 @@ def vm_kill(self):
...
+        if self.dest.name == 'dom0':
+            type(self.app).add_handler('*', handler)
+        qubes.vm.BaseVM.add_handler('*', handler)
+
+        # send artificial event as a confirmation that connection is established
+        self.send_event(self.app, 'connection-established')
+
+        try:
+            yield from wait_for_cancel
+        except asyncio.CancelledError:
```
woju: This seems suspicious, because technically ``CancelledError`` may have resulted from either ``self.cancel()`` or ``wait_for_cancel.cancel()``.  Those conditions aren't distinguishable from try/except.

I don't currently know, how this one should be expressed. I'd have to read some other people's code.

qubes/mgmt.py
```diff
@@ -534,3 +570,95 @@ def vm_kill(self):
...
+            del self.dest.features[self.arg]
+        except KeyError:
+            raise qubes.exc.QubesFeatureNotFoundError(self.dest, self.arg)
+        self.app.save()
+
+    @api('mgmt.vm.feature.Set')
+    @asyncio.coroutine
+    def vm_feature_set(self, untrusted_payload):
+        # validation of self.arg done by qrexec-policy is enough
+        value = untrusted_payload.decode('ascii', errors='strict')
```
woju:
```python
del untrusted_payload
```

qubes/tools/qvm_ls.py
```diff
@@ -1,634 +0,0 @@
...
-        columns = formats[args.format]
-
-    table = Table(args.app, columns)
-    table.write_table(sys.stdout)
-
-    return 0
-
-
-if __name__ == '__main__':
-    sys.exit(main())
```
woju: This is a nice one. I like red diffs. :+1:

rpm_spec/core-dom0.spec
```diff
@@ -349,7 +347,7 @@ fi
 %{python3_sitelib}/qubes/tests/tools/init.py
 %{python3_sitelib}/qubes/tests/tools/qvm_device.py
 %{python3_sitelib}/qubes/tests/tools/qvm_firewall.py
-%{python3_sitelib}/qubes/tests/tools/qvm_ls.py
+%{python3_sitelib}/qubes/tests/tools/qubesd.py
```
woju: This breaks lexical ordering, should be before qvm_*

woju:
- I didn't found a signed tag anywhere.
- You may consider rebasing it on top of core3-devel, or even taking ref from current core3-mgmt-methods3 branch.

qubes/events.py
```diff
@@ -115,6 +115,17 @@ def add_handler(cls, event, func):
...
+    def remove_handler(cls, event, func):
+        '''Remove event handler from subject's class.
+
+        :param str event: event identificator
+        :param collections.Callable handler: handler callable
+        '''
+
+        # pylint: disable=no-member
+        cls.__handlers__[event].remove(func)
+
```
marmarek: This will duplicate events distribution code.

qubes/mgmt.py
```diff
@@ -112,6 +112,12 @@ def __init__(self, app, src, method, dest, arg):
         #: name of the method
         self.method = method.decode('ascii')
 
+        #: callback for sending events if applicable
+        self.send_event = send_event
+
+        #: is this operation cancellable?
+        self.cancellable = False
+
```
marmarek: This means: should this operation be canceled by a client (by closing connection in this case).

qubes/mgmt.py
```diff
@@ -135,9 +141,25 @@ def __init__(self, app, src, method, dest, arg):
...
+        self._running_handler = None
         del untrusted_candidates
 
+    def execute(self, *, untrusted_payload):
+        '''Execute management operation.
+
+        This method is a coroutine.
+        '''
+        self._running_handler = asyncio.Task(self._handler(
+            untrusted_payload=untrusted_payload))
```
marmarek: Ok

qubes/mgmt.py
```diff
@@ -534,3 +570,95 @@ def vm_kill(self):
         assert not self.arg
         self.fire_event_for_permission()
         yield from self.dest.kill()
+
+    @api('mgmt.Events', no_payload=True)
```
marmarek: Yes, it is expected. Not sure explicitly documented, but this is exactly the same as for all other methods - it does anything only after sending EOF.

qubes/mgmt.py
marmarek: Both conditions should be handled the same here, so not a problem.

qubes/mgmt.py
```diff
@@ -534,3 +570,95 @@ def vm_kill(self):
...
+        event_filters = self.fire_event_for_permission()
+
+        def handler(subject, event, **kwargs):
+            if self.dest.name != 'dom0' and subject != self.dest:
+                return
+            if event.startswith('mgmt-permission:'):
+                return
+            for selector in event_filters:
+                if not selector((subject, event, kwargs)):
+                    return
```
marmarek: Not sure about this, it's a single if in a simple loop...

qubes/mgmt.py
marmarek: Ok.

marmarek:
> You may consider rebasing it on top of core3-devel, or even taking ref from current core3-mgmt-methods3 branch.

What do you mean? Should I merge this with core3-mgmt-methods3 into one PR?

qubes/mgmt.py
woju: Yeah, and what happens if client closes connection for some method not marked as cancellable?

qubes/events.py
woju: Yes, it somehow will. However, the current code is sequential (since the handlers are called in turns) and this code could be async-aware. And I think it should be around 10 lines, including registering and deregistering sockets.

qubes/mgmt.py
woju: Should it be that way? Or don't we want to write some other, optimised socket interface especially for this function? I don't know. This class (and approach) seems just barely adequate...

qubes/mgmt.py
marmarek: Operation continues until finished. The only "difference" is that caller will not get result,

qubes/mgmt.py
marmarek: We want the protocol as simple as possible. Surely we don't want to have interactive protocol here. And requiring EOF before receiving anything (and actually triggering the operation) seems like a straight forward solution - this way we're sure there wont be additional input coming in.
Re "socket interface" - keep in mind _the same_ should also work over qrexec.

qubes/events.py
marmarek: Ok.
So, how exactly you see this? Keep in mind that each API call is a new instance of QubesMgmt, so registering etc should be stored at class level or such...

qubes/mgmt.py
woju: Can you guarantee no-one will ever change that? I can't.

qubes/events.py
woju: I'd write QubesDaemonEventDispatcher class, which would be instantiated once in qubes.tools.qubesd:main and passed as argument to constructor of QubesDaemonProtocol (see that partial() in loop.create_unix_server(). The mgmt.Event handler would register its connection as the outgoing event channel.

Quick round of affected methods, names are not final:

```python
class QubesDaemonEventDispatcher(object):
  def __init__(self, app):
    self.app = app
    self._connections = collections.defaultdict(list)
    type(self.app).add_handler('*', self._handler)
    qubes.vm.BaseVM.add_handler('*', self._handler)
  
  def _handler(self, subject, event, **kwargs):
    for connection in self._connections[subject]:
      connection.send_event(subject, event, **kwargs)

  def register(self, connection, subject):
    assert connection not in self._connections[subject]
    self._connections[subject].append(connection)

  def unregister(self, connection, subject):
    if subject is not None:
      self._connections[subject].remove(connection)
    else:
      for subjcoll in self._connection.values():
        try:
          subjcoll.remove(connection)
        except ValueError:
          pass

class QubesDaemonProtocol(asyncio.Protocol):
  def __init__(self, *args, app, eventdisp, debug=False, **kwargs):
    # ...
    self.eventdisp = eventdisp

  def events_register(self, subject):
    self.eventdisp.register(self, subject)

  def events_unregister(self, subject):
    self.eventdisp.unregister(self, subject)

  def connection_lost(self):
    self.events_unregister(None)

class QubesMgmt(object):
  @api('mgmt.Event', no_payload=True)
  @asyncio.coroutine
  def event(self):
    # ...
    self.events_register(self.app if self.dest.name == 'dom0' else self.dest)
    yield from wait_for_cancel

File: qubes/mgmt.py (hunk 9 above)
Comment (woju) [inline qubes/mgmt.py]: Thanks.

File: qubes/mgmt.py (hunk 8 above)
Comment (woju) [inline qubes/mgmt.py]: Makes sense. If we can document the EOF requirement as a security feature, developers might actually remember about that.

File: qubes/mgmt.py (hunk 6 above)
Comment (woju) [inline qubes/mgmt.py]: We may want to write that down somewhere. This is probably wise, so long operation don't get interrupted and result for example in some inconsistent filesystem state.

**Comment by woju** [comment]
> What do you mean? Should I merge this with core3-mgmt-methods3 into one PR?

I don't actually care for one or two PRs, as long as they serve common purpose. Do as you find convenient.

**Comment by marmarek** [comment]
I've split it into separate branches to allow partial merges - do not block huge changes on simple issues. I'll leave it as is.

File: qubes/mgmt.py (hunk 8 above)
Comment (marmarek) [inline qubes/mgmt.py]: While at it - I'd like to abandon this part of API:
```
Normally server will not close the connection.
A method yielding events will not ever return a OK or EXCEPTION message.
```

Reason: a method could progress reports as events. This could be especially useful for backup-related calls, but also for VM clone etc.
Alternative: broadcasting progress reports as "normal" events. But in some cases it would be hard to correlate progress reports with actual operation.

qubes/mgmt.py
marmarek: Added to /doc/mgmt1/ as notes. Might be a good idea to include also in (currently WIP) qubes-mgmt devel docs.

qubes/mgmt.py
marmarek: I take that back - broadcasting such events would be better - will not complicate handling methods when the tool doesn't care about progress, and will allow to display the progress by a separate tool (for example tray icon in GUI VM, even if backup operation triggered by Mgmt VM/cron).

qubes/events.py
marmarek: I don't think passing events directly is a good idea - first of all, we need to apply filters on them (for example those returned by mgmt-permission:mgmt.Events event), but also we may want to introduce more events-processing methods, which would behave differently.
The more I think about it, the more I'm convinced that original approach was better (adding remove_handler function).
You objection about `vm.remove_handler` is invalid anyway, because one shouldn't add/remove events on object instance - it will add/remove even for the whole class, as you said. The same apply to `add_handler`. Is it possible to detect such things and raise some exception?
Yes, `remove_handler` should be called on the same class that `add_handler` was called. It's caller responsibility to do that (the same as providing proper event name, a proper callable, and generally using API properly).
I'll add this to the docstring.

woju: We can have filters in Dispatcher:

```python
def register(self, connection, subject, filters):
  self._connections[subject].append((connection, filters))
def _handler(self, subject, event, **kwargs):
  for connection, filters in self._connections[subject]:
    if not tuple(apply_filters([(subject, event, kwargs)], filters)):
      continue
    connection.send_event(subject, event, **kwargs)

No, it's not possible to detect that. If you call ``obj.clsmethod`` (when clsmethod is decorated with ``@classmethod``), you're really calling ``type(obj).clsmethod``. That is, unless you'd like to reimplement ``classmethod``. It would be possible to have per-instance events instead, by having __handlers__ per-instance, and initialised in ``__new__`` with anything the class wishes.

marmarek: Still, I don't see any reason why duplicate event handling code when we have it in place already. As explained above - behavior of _invalid call_ to `remove_handler` does not warrant not implementing it.

PR #106, opened 2017-05-24
Fix regex error preventing USB hubs from working

mdickers47:
On my machine, attaching a device to a USB hub results in a device name such as "5-4_1_1".  This fails to match a regex in qubesutils.py, which only expects one "_1" component.

With this change, qvm-usb and USB passthrough work as expected for me.  I don't know enough about qubes/Xen to exhaustively test if there are any unintended consequences for other use cases.

### DIFF
--- a/core/qubesutils.py
+++ b/core/qubesutils.py
@@ -469,7 +469,7 @@ def block_detach_all(vm):
 ####### USB devices ######
 
 usb_ver_re = re.compile(r"^(1|2)$")
-usb_device_re = re.compile(r"^[0-9]+-[0-9]+(_[0-9]+)?$")
+usb_device_re = re.compile(r"^[0-9]+-[0-9]+(_[0-9]+)*$")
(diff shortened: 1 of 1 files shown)


core/qubesutils.py
```diff
@@ -469,7 +469,7 @@ def block_detach_all(vm):
 ####### USB devices ######
 
 usb_ver_re = re.compile(r"^(1|2)$")
-usb_device_re = re.compile(r"^[0-9]+-[0-9]+(_[0-9]+)?$")
+usb_device_re = re.compile(r"^[0-9]+-[0-9]+([_0-9]+)?$")
```
marmarek: `_` was intentionally outside of [], to not allow it as the last character. The proper fix is to change last `?` to `*`.

mdickers47:
Do you want me to delete this fork and start a new one so that all the commits in the history can be signed?  Or just commit the one-byte change somewhere else?

marmarek:
I think the cleanest thing to do, is to force-push the right change (one commit) to your branch.

mdickers47:
done.

PR #109, opened 2017-06-06
Fixes, improvements

marmarek:
A bunch of mostly unrelated commits - see individual commit messages.

WARNING: https://github.com/QubesOS/qubes-core-admin/commit/fa1da422d1a6b6a06282932869559fa58ca14941 change qubes.xml format - old file will not load. You need to manually remove dom0 attributes (name, qid, uuid).


marmarek:
This also cover QubesOS/qubes-issues#1858

qubes/ext/core_features.py
```diff
@@ -0,0 +1,62 @@
...
+        if getattr(vm, 'template', None):
+            vm.log.warning(
+                'Ignoring qubes.NotifyTools for template-based VM')
+            return
+
+        # for now used only to check for the tools presence
+        if 'version' in untrusted_features:
+            # any suspicious string will raise exception here,
+            # but otherwise ignored
+            int(untrusted_features['version'])
```
woju: This looks wrong. First of all, it's pointless: if we are doing nothing (short of raising exception), maybe we shouldn't be parsing this at all. Second, why this wasn't renamed to "qubes-tools/version" or something like that? Remember this is received by all extensions, not just this one.

marmarek: This is version of qubes tools (or rather: dom-VM interface of it). Currently we don't have any need for negotiating this, but we may need in the future. So, having (Windows) VM exposing "1" here may be useful in the future.

woju: There is nothing that prevents VM from exposing "1" or whatever it pleases without parsing this on dom0's end. When the need arises, someone will be rewriting the extension anyway, so we can just drop this now. Until then, we just ignore unknown keys, this one included.

As I said, I think it's not just useless, it actually sends unqualified "version" feature request to all extensions, without saying what exactly is versioned (gui? qrexec? the qubesdb handshake?).

This is because event handlers of features-request don't generally get context of the call and I'd imagine they contain something like:

```python
if 'my_expected_feature' in untrusted_features:
    # ...
```

From extension writer's perspective, there are some problems hiding around this handler. For one, we never promised that one `qubes.FeaturesRequest` will arrive as single event (and we obviously won't make such promise).

OTOH, other features' names (esp. "gui" and "qrexec") are good precisely because they are very general and I think more than one extension will have interest in them.

PR #111, opened 2017-06-12
Devices API cleanup, fixes

marmarek:
This is first (smaller) part of Devices API cleanup.
Besides this, few fixes from Friday.


qubes/vm/qubesvm.py
```diff
@@ -492,7 +487,7 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.BaseVM):
             dom0 boot.''')
 
     include_in_backups = qubes.property('include_in_backups',
-        default=(lambda self: not self.internal),
+        default=True,
```
woju: What about dom0? Does it even have this property?

marmarek: No, it doesn't.

PR #112, opened 2017-06-14
Bugfixes, remove 'version' feature request

marmarek:
See individual commit messages

### DIFF
--- a/qubes/vm/adminvm.py
+++ b/qubes/vm/adminvm.py
@@ -52,6 +52,10 @@ class AdminVM(qubes.vm.BaseVM):
     def __str__(self):
         return self.name
 
+    def __lt__(self, other):
+        # order dom0 before anything
+        return self.name != other.name
+
     @property
     def attached_volumes(self):
         return []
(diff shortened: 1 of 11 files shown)


qubes/vm/adminvm.py
```diff
@@ -52,6 +52,9 @@ class AdminVM(qubes.vm.BaseVM):
     def __str__(self):
         return self.name
 
+    def __lt__(self, other):
+        return self.name < other.name
+
```
woju: Dom0 should be sorted before anything else, that is, it should unconditionally `return True`.

marmarek: What about dom0 < dom0 ?

woju:
On Mon, Jun 19, 2017 at 08:29:41AM -0700, Marek Marczykowski-Górecki wrote:
> What about dom0 < dom0 ? 

I don't think that happens. And if happened, tough shit: maybe some static
sorting won't be static or I don't know.

Python's docs say nowhere that a < b implies b > a.

marmarek:
> Python's docs say nowhere that a < b implies b > a.

[This](https://docs.python.org/3/reference/datamodel.html#object.__le__) and [this](https://www.python.org/dev/peps/pep-0207/) suggests something different:

      The reflexivity rules *are* assumed by Python.  Thus, the
      interpreter may swap y>x with x<y, y>=x with x<=y, and may swap
      the arguments of x==y and x!=y.  (Note: Python currently assumes
      that x==x is always true and x!=x is never true; this should not
      be assumed.)

PR #113, opened 2017-06-15
qubes: admin extension

woju:
forbid mangling `created-by-<vm>` tags

### DIFF
--- a/qubes/ext/admin.py
+++ b/qubes/ext/admin.py
@@ -29,6 +29,8 @@ class AdminExtension(qubes.ext.Extension):
         '''Forbid changing specific tags'''
         # pylint: disable=no-self-use,unused-argument
         if arg.startswith('created-by-'):
-            raise qubes.api.PermissionDenied()
+            raise qubes.api.PermissionDenied(
+                'changing this tag is prohibited by {}.{}'.format(
+                    __name__, type(self).__name__))
 
-    # TODO create that extension in the first place
+    # TODO create that tag here (need to figure out how to pass mgmtvm name)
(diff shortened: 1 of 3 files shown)


qubes/ext/admin.py
```diff
@@ -0,0 +1,34 @@
...
+class AdminExtension(qubes.ext.Extension):
+    # pylint: disable=too-few-public-methods
+    @qubes.ext.handler(
+        'mgmt-permission:admin.vm.tag.Set',
+        'mgmt-permission:admin.vm.tag.Remove')
+    def on_tag_set_or_remove(self, vm, event, arg, **kwargs):
+        '''Forbid changing specific tags'''
+        # pylint: disable=no-self-use,unused-argument
+        if arg.startswith('created-by-'):
+            raise qubes.api.PermissionDenied()
```
marmarek: Some message here (with this exception it will be in logs only, but still)?
Also, here is a more general question - should the reason for permission denied be reported to the caller? Or rejected without giving any clue (as it is currently)? @rootkovska

rootkovska: I don't have any strong opinion on either option, but generally: if we (dom0) don't have a good reason to excuse ourselves to a VM, than we should not do that. Of course we should log the reason to dom0's journal.

woju: Any message attached here won't get to the log anyway: https://github.com/QubesOS/qubes-core-admin/blob/96014970195a99e828ad59321b311076d92412dc/qubes/tools/qubesd.py#L90-L94

Even if it should, this is not the place to correct it.

marmarek: So, where would be the correct place? I.e. why _not_ put a message here and fix logging later?

PR #114, opened 2017-06-19
admin.vm.tag.*, volume clone

marmarek:


qubes/api/admin.py
```diff
@@ -336,6 +354,49 @@ def vm_volume_import(self):
...
+        return ''.join('{}\n'.format(tag) for tag in sorted(tags))
+
+    @qubes.api.method('admin.vm.tag.Get', no_payload=True)
+    @asyncio.coroutine
+    def vm_tag_get(self):
+        qubes.vm.Tags.validate_tag(self.arg)
+
+        self.fire_event_for_permission()
+
+        return '1' if self.arg in self.dest.tags else '0'
```
woju: As I said earlier, I'd like this one to say 
```python
return self.arg if self.arg in self.dest.tags else ''
```

The only string that's `False` is an empty string. As to the value, we could either use `'1'`, since it's non-empty and clearly "true", or full tag name.

Do we allow zero-length tag name?

qubes/storage/__init__.py
```diff
@@ -431,46 +431,44 @@ def create(self):
...
+            clone_op_ret = dst_pool.import_volume(
+                dst_pool, dst, src_pool, src_volume)
 
         # clone/import functions may be either synchronous or asynchronous
         # in the later case, we need to wait for them to finish
-        clone_op = {}
+        if asyncio.iscoroutine(clone_op_ret):
+            self.vm.volumes[name] = yield from clone_op_ret
+        else:
+            self.vm.volumes[name] = clone_op_ret
```
woju: This is very minor, but Pythonist would write:
```python
clone_op_ret = ...
if asyncio.iscoroutine(clone_op_ret):
  clone_op_ret = yield from clone_op_ret
self.vm.volumes[name] = clone_op_ret
```
That is, take advantage of the fact that Python is not static typed and we don't write the main operation (in this case `self.vm.volumes[name] = ...`) twice.

qubes/api/admin.py
marmarek: > The only string that's False is an empty string. As to the value, we could either use '1', since it's non-empty and clearly "true", or full tag name.

One (of) issues with this (and also using empty string as "false" value for features) is that if something fails and response is truncated (either inside qubesd, or in transit), it will be mistaken for "false" value. This is one reason why it's better to have defined _non empty_ values for for valid responses. Another reason is that some languages have hardly distinction between "no value" and "empty value" (think: shell). This API should be designed to make mistakes harder (this is not the contest who makes bigger footgun!). Management VM will make security-critical decisions based on API responses!

> Do we allow zero-length tag name?

No. And also we allow only very limited character set in tag names, even though parts used in current Admin API implementation would not have troubles with full ASCII or even unicode. Exactly for the reason explained above.

woju: Having convention of `'0'` being false is a footgun for Python and requires some precautions  C. f. `qubes.property.bool`: we were lucky in that case because there is no easy way to get around that setter. In this case it is much easier. And IIRC we already have some places when `False`/`None` gets written as zero-length.

This particular call seems to be only used for `__contains__` dunder in core-admin-client, so this one seems innocent, however I'd be wary of introducing this in more places that needed.

marmarek: Ok, so lets leave `'0'` here and not change in other places (where we already use empty string)

PR #117, opened 2017-06-26
Add support for asynchronous event handlers

marmarek:
Some event handlers need to call async actions (like qrexec service). Add support for it.
Two important design choices:
 - event handlers can still be normal python function (instead of a coroutine), if given handler don't need any async feature
 - not all events support async handlers - see appropriate event documentation for this information

### DIFF
--- a/doc/qubes-events.rst
+++ b/doc/qubes-events.rst
@@ -143,6 +143,75 @@ returned to the caller as list. The order of this list is undefined.
    effect = o.fire_event('event1')
 
 
+Asynchronous event handling
+---------------------------
+
+Event handlers can be defined as coroutine. This way they can execute long
+running actions without blocking the whole qubesd process. To define
+asynchronous event handler, annotate a coroutine (a function defined with
+`async def`, or decorated with `py:func:`asyncio.coroutine`) with
+py:func:`qubes.events.handler` decorator. By definition, order of
+such handlers is undefined.
+
+Asynchronous events can be fired using
+:py:meth:`qubes.events.Emitter.fire_event_async` method. It will handle both
+synchronous and asynchronous handlers. It's an error to register asynchronous
+handler (a coroutine) for synchronous event (the one fired with
+:py:meth:`qubes.events.Emitter.fire_event`) - it will result in
+:py:exc:`RuntimeError` exception.
+
+.. code-block:: python
+
+   import asyncio
+   import qubes.events
+
+   class MyClass(qubes.events.Emitter):
+       @qubes.events.handler('event1', 'event2')
+       @asyncio.coroutine
+       def event_handler(self, event):
+           if event == 'event1':
+               print('Got event 1, starting long running action')
+               yield from asyncio.sleep(10)
+               print('Done')
+
+   o = MyClass()
+   loop = asyncio.get_event_loop()
+   loop.run_until_complete(o.fire_event_async('event1'))
+
+Asynchronous event handlers can also return value - but only a collection, not
+yield individual values (because of python limitation):
+
+.. code-block:: python
+
+   import asyncio
+   import qubes.events
+
+   class MyClass(qubes.events.Emitter):
+       @qubes.events.handler('event1')
+       @asyncio.coroutine
+       def event_handler(self, event):
+            print('Got event, starting long running action')
+            yield from asyncio.sleep(10)
+            return ('result1', 'result2')
+
+       @qubes.events.handler('event1')
+       @asyncio.coroutine
+       def another_handler(self, event):
+            print('Got event, starting long running action')
+            yield from asyncio.sleep(10)
+            return ('result3', 'result4')
+
+       @qubes.events.handler('event1')
+       def synchronous_handler(self, event):
+            yield 'sync result'
+
+   o = MyClass()
+   loop = asyncio.get_event_loop()
+   # returns ['sync result', 'result1', 'result2', 'result3', 'result4'],
+   # possibly not in order
+   effects = loop.run_until_complete(o.fire_event_async('event1'))
+
+
 Module contents
 ---------------
 
--- a/doc/qubes-events.rst
+++ b/doc/qubes-events.rst
@@ -22,10 +22,10 @@ parent class and then for it's child. For each class, first are called handlers
 defined in it's source, then handlers from extensions and last the callers added
 manually.
 
-There is second method, :py:meth:`qubes.events.Emitter.fire_event_pre`, which
-fires events in reverse order. It is suitable for events fired before some
-action is performed. You may at your own responsibility raise exceptions from
-such events to try to prevent such action.
+The :py:meth:`qubes.events.Emitter.fire_event` method have keyword argument
+`pre_event`, which fires events in reverse order. It is suitable for events
+fired before some action is performed. You may at your own responsibility raise
+exceptions from such events to try to prevent such action.
 
 Events handlers may yield values. Those values are aggregated and returned
 to the caller as a list of those values. See below for details.
--- a/doc/qubes-events.rst
+++ b/doc/qubes-events.rst
@@ -149,8 +149,8 @@ Asynchronous event handling
 Event handlers can be defined as coroutine. This way they can execute long
 running actions without blocking the whole qubesd process. To define
 asynchronous event handler, annotate a coroutine (a function defined with
-`async def`, or decorated with `py:func:`asyncio.coroutine`) with
-py:func:`qubes.events.handler` decorator. By definition, order of
+`async def`, or decorated with :py:func:`asyncio.coroutine`) with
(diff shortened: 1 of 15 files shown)


qubes/events.py
```diff
@@ -157,50 +160,92 @@ def _fire_event_in_order(self, order, event, kwargs):
                     key=(lambda handler: hasattr(handler, 'ha_bound')),
                     reverse=True):
                 effect = func(self, event, **kwargs)
+                # "is True" to rule out unittest.mock.Mock...
```
woju: #wat?

qubes/events.py
```diff
@@ -157,50 +160,92 @@ def _fire_event_in_order(self, order, event, kwargs):
...
-            itertools.chain(reversed(self.__class__.__mro__), (self,)),
-            event, kwargs)
-
-
-    def fire_event_pre(self, event, **kwargs):
-        '''Call all handlers for an event.
+        if pre_event:
+            order = itertools.chain((self,), self.__class__.__mro__)
+        else:
+            order = itertools.chain(reversed(self.__class__.__mro__), (self,))
```
woju:
```python
order = itertools.chain((self,), self.__class__.__mro__)
if not pre_event:
  order = reversed(order)
```
Same as the last one: less places to err.

Also, I'd rename `_fire_event_in_order` to `_fire_event` and put the above logic in there, since it is shared between this and and `fire_event_async` (c.f. less places to err). And I'd inline `_handle_event_async` in the latter.

qubes/events.py
```diff
@@ -157,50 +160,92 @@ def _fire_event_in_order(self, order, event, kwargs):
                     key=(lambda handler: hasattr(handler, 'ha_bound')),
                     reverse=True):
                 effect = func(self, event, **kwargs)
+                # "is True" to rule out unittest.mock.Mock...
+                if getattr(func, 'ha_async', False) is True:
+                    assert effect is not None
```
woju: I'd very much prefer `assert inspect.isawaitable(effect)` instead of `is not None`. `*.iscoroutine()` may not be enough, since someone might like to return a plain `Future` (not a `Task`) or whatever.

qubes/events.py
```diff
@@ -157,50 +160,92 @@ def _fire_event_in_order(self, order, event, kwargs):
...
+        Handlers are called for class and all parent classes, in **reversed**
+        or **true** (depending on *pre_event* parameter)
+        method resolution order. For each class first are called bound handlers
+        (specified in class definition), then handlers from extensions. Aside
+        from above, remaining order is undefined.
 
-        Handlers are called for class and all parent classess, in **true**
-        method resolution order. This is intended for ``-pre-`` events, where
-        order of invocation should be reversed.
+        Thi method call both synchronous and asynchronous handlers. Order of
```
woju: s/Thi/This/

qubes/events.py
```diff
@@ -157,50 +160,92 @@ def _fire_event_in_order(self, order, event, kwargs):
                     key=(lambda handler: hasattr(handler, 'ha_bound')),
                     reverse=True):
                 effect = func(self, event, **kwargs)
+                # "is True" to rule out unittest.mock.Mock...
+                if getattr(func, 'ha_async', False) is True:
```
woju: Now I think, couldn't we replace this one with `asyncio.iscoroutinefunction(func)` (not to be confused with `inspect.iscoroutinefunction()`, which is `False` on `@asyncio.coroutine`s)?

That way we'd deal with one corner case:
```python
class MyClass(qubes.events.Emitter):
  @qubes.events.handler('my-event', is_async=True)
  # NOT asyncio.coroutine
  def on_my_event(self, event):
    return asyncio.Future()
```

With `iscoroutinefunction`, the above example, admittedly valid but unreadable, becomes forbidden. And we wouldn't have to remember about `is_async=` and keep it in sync with reality.

Argument against: there are 2 incompatible versions of `iscoroutinefunction()` in python's stdlib (the difference is around `@asyncio.coroutine` as above) and 2.5 versions of `iscoroutine()` (there is also `isinstance(obj, types.CoroutineType)`,which is same as `inspect.iscoroutine()` and not compatible with `asyncio.iscoroutine()`). We can get it right, but after that, we are in the area of "what could possibly go wrong".

qubes/ext/__init__.py
```diff
@@ -86,6 +88,9 @@ def decorator(func):
         else:
             func.ha_vm = qubes.vm.BaseVM
 
+        if kwargs.get('is_async', False):
+            func.ha_async = True
+
```
woju: `kwargs.get()` in that function was a workaround to get keyword-only arguments in python2 (this is one of the oldest parts of core3).

#TODO rewrite as keyword-only argument. Low priority.

qubes/events.py
marmarek: If func is unittest.mock.Mock instance, it do have attribute `ha_async` (in fact: any attribute), which is a new unittest.mock.Mock instance, `bool(unittest.mock.Mock())` is true

qubes/events.py
marmarek: That wouldn't work on your hacky coroutines which do not have `@asyncio.coroutine`, but have a single `return some_other_coroutine(...)`.

qubes/events.py
marmarek: AFAIR there was some problem with `inspect.isawaitable()` - we did considered it for very similar case in storage, but discarded it there.

marmarek:
extracted from commit message:
> One limitation (for now) is that Emiter.add_handler does not support
registering async handlers (`func.ha_async = True` can raise for example
`AttributeError: 'method' object has no attribute 'ha_async'`).

Is it ok to leave if this way, or there is some _simple_ method to solve it?

qubes/events.py
marmarek: That would also solve the problem with `emitter.add_handler()`. 👍 for removing those hacky coroutines without `@asyncio.coroutine`.

qubes/events.py
woju: Please no. Fix your mock instead.

qubes/events.py
woju:
```python
def isawaitable(object):
    """Return true if object can be passed to an ``await`` expression."""
    return (isinstance(object, types.CoroutineType) or
            isinstance(object, types.GeneratorType) and
                bool(object.gi_code.co_flags & CO_ITERABLE_COROUTINE) or
            isinstance(object, collections.abc.Awaitable))
```

What's wrong with it?

qubes/events.py
marmarek: This will be gone after switching to iscoroutinefunction.

qubes/events.py
marmarek: To be honest, I don't remember. I'll try ;)

qubes/events.py
woju: Ah, OK.

qubes/events.py
marmarek: Doesn't work:
```python
>>> import itertools
>>> order = itertools.chain((1, ), (2, 3, 4))
>>> reversed(order)
  File "<stdin>", line 1, in <module>
TypeError: argument to reversed() must be a sequence
```
Should I force it into a list or something? Or leave the old version (but still moved to the other place)?

woju: Sorry, my fault. `order = reversed(list(order))`.

marmarek:
Force-pushed fixed version. Oh, it fails on travis, but works on my machine. The fail is about `inspect.isawaitable()`...

marmarek:
So, back to `assert effect is not None`? Or any better idea?

marmarek:
Dropped that assert (it was to detect non-coroutines registered with is_async=True, so it is irrelevant now).
And now fails on https://bugs.python.org/issue28703 ("Fix asyncio.iscoroutinefunction to handle Mock objects"), fixed in python 3.5.3 (we have 3.5.2 on travis...).
And lets see:
```patch
-    return (getattr(func, '_is_coroutine', False) or
+    return (getattr(func, '_is_coroutine', None) is _is_coroutine or
```
So, exactly the syntax you didn't liked 😛 
Added a workaround in tests...

doc/qubes-events.rst
```diff
@@ -143,6 +143,75 @@ returned to the caller as list. The order of this list is undefined.
    effect = o.fire_event('event1')
 
 
+Asynchronous event handling
+---------------------------
+
+Event handlers can be defined as coroutine. This way they can execute long
+running actions without blocking the whole qubesd process. To define
+asynchronous event handler, annotate a coroutine (a function defined with
+`async def`, or decorated with `py:func:`asyncio.coroutine`) with
```
woju: this should be
```rst
:py:func:`asyncio.coroutine`
```

see the problem with syntax highlighting on github

marmarek:
@woju I need you to approve this again, since I've pushed new commit.

PR #119, opened 2017-06-26
Implement firewall in Admin API

marmarek:


qubes/api/admin.py
```diff
@@ -991,3 +992,37 @@ def vm_device_detach(self, endpoint):
...
+            for rule in self.dest.firewall.rules)
+
+    @qubes.api.method('admin.vm.firewall.Set')
+    @asyncio.coroutine
+    def vm_firewall_set(self, untrusted_payload):
+        assert not self.arg
+        rules = []
+        for untrusted_line in untrusted_payload.decode('ascii',
+                errors='strict').splitlines():
+            rule = qubes.firewall.Rule.from_api_string(untrusted_line)
```
woju: Would you mind calling that as keyword argument (`.from_api_string(untrusted_line=untrusted_line)`)? It would prevent errors if/when someone would forget to name this argument in `RuleOption` subclass.

qubes/firewall.py
```diff
@@ -368,29 +448,30 @@ def __init__(self, vm, load=True):
...
         if isinstance(other, Firewall):
-            return self.policy == other.policy and self.rules == other.rules
+            return self.rules == other.rules
         return NotImplemented
 
     def load_defaults(self):
         '''Load default firewall settings'''
-        self.rules = []
-        self.policy = Action('accept')
+        self.rules = [Rule(None, action='accept')]
```
woju: :+1:

qubes/firewall.py
```diff
@@ -422,9 +503,9 @@ def load_v1(self, xml_root):
         policy_v1 = xml_root.get('policy')
         assert policy_v1 in ('allow', 'deny')
         if policy_v1 == 'allow':
-            self.policy = Action('accept')
+            policy = Action('accept')
         else:
-            self.policy = Action('drop')
+            policy = Action('drop')
```
woju: Do I understand this correctly: this is how firewall model used to work and this code is used to load policies written by previous Qubes releases?

Is `policy` variable used as this object? Wouldn't a simple boolean flag suffice?

qubes/tests/api_admin.py
```diff
@@ -1734,6 +1736,131 @@ def test_561_tag_set_invalid(self):
...
+
+    def test_571_firewall_get_non_default(self):
+        self.vm.firewall.save = unittest.mock.Mock()
+        self.vm.firewall.rules = [
+            qubes.firewall.Rule(action='accept', proto='tcp',
+                dstports='1-1024'),
+            qubes.firewall.Rule(action='drop', proto='icmp',
+                comment='No ICMP'),
+            qubes.firewall.Rule(action='drop', proto='udp',
+                expire='1499450306'),
```
woju: What's this date? Looks like magic value.

qubes/firewall.py
marmarek: I think this is more readable.

qubes/tests/api_admin.py
marmarek: This is any value, just to compare few lines before.

qubes/firewall.py
woju: I disagree, because the reader wonders, what for this variable really is. Could `default_policy_is_accept` (with `bool()` values) be a replacement?

marmarek: Ok

qubes/api/admin.py
marmarek: That wouldn't be enough - see `Rule.from_api_string` - it passes values using kwargs (implemented by PropertyHolder). I can construct actual `RuleOption` objects earlier...

woju: Ah, I understand. Well, I don't know what to do. I don't think passing explicitly untrusted values to PropertyHolder is a good idea, but it isn't particularly bad. We probably do it elsewhere, and setters/classes are kind of expected to handle untrusted data. So this is your call.

PR #120, opened 2017-06-26
Changes to storage handling for Admin API

marmarek:
The most important change here is to split volume.Clone into volume.CloneFrom and volume.CloneTo. But while at it, also fixed some minor issues.


qubes/api/admin.py
```diff
@@ -293,22 +293,57 @@ def vm_volume_revert(self, untrusted_payload):
...
 
         volume = self.dest.volumes[self.arg]
 
-        self.fire_event_for_permission(target_vm=target_vm, volume=volume)
+        self.fire_event_for_permission(volume=volume)
+
+        token = qubes.utils.random_string(32)
+        # save token on self.app, as self is not persistent
+        if not hasattr(self.app, 'api_admin_pending_clone'):
+            self.app.api_admin_pending_clone = {}
```
woju: And the title of "Distinguished Pythonist" goes to...

qubes/api/admin.py
```diff
@@ -293,22 +293,57 @@ def vm_volume_revert(self, untrusted_payload):
...
+        assert self.arg in self.dest.volumes.keys()
+        untrusted_token = untrusted_payload.decode('ascii').strip()
+        del untrusted_payload
+        assert untrusted_token in getattr(self.app,
+            'api_admin_pending_clone', {})
+        token = untrusted_token
+        del untrusted_token
+
+        src_volume = self.app.api_admin_pending_clone[token]
+        del self.app.api_admin_pending_clone[token]
```
woju: This means that to retry failed operation, user should get a new token. This is probably correct, but should be documented in /admin-api/.

marmarek: Done: https://github.com/QubesOS/qubes-doc/commit/c4f41c20216c1f6b983744472c6b7f5ad3055ee3

PR #127, opened 2017-07-08
Dispvm fix

marmarek:
Two fixes:
 - storage definition
 - asyncio usage

Fixes QubesOS/qubes-issues#2896


woju:
Could the `while template` loop get some comment as to what it aims to do? I think I understood that, but it doesn't feel obvious enough and I think it could use a short comment.

woju:
Or maybe whole `if source is not None`. I dunno.

marmarek:
> Could the while template loop get some comment as to what it aims to do? I think I understood that, but it doesn't feel obvious enough and I think it could use a short comment.

Sure, done.

PR #129, opened 2017-07-12
features: change 'service/' prefix to 'service.'

marmarek:
'/' is not allowed in qrexec service call argument, but feature name is
passed there in Admin API.

Fixes QubesOS/qubes-issues#2906


woju:
woju/qubes-core-admin@devel-services-with-dot, and I didn't forgot about tests :)

PR #145, opened 2017-08-31
Fix test suite, esp. fd leaks

woju:
After this merge, the tests should either pass, or fail for reasons outside the test suite (but not necessarily for the reasons related to the expected behaviour in a particular test case). Some core classes were fixed not to leak file descriptors when recycled multiple times. Also there are some smallimprovements to the test tools.

The impact on qubesd is probably negligible, since the latter instantiates the objects only once.

### DIFF
--- a/qubes/app.py
+++ b/qubes/app.py
@@ -139,13 +139,12 @@ class VirConnectWrapper(object):
 class VMMConnection(object):
     '''Connection to Virtual Machine Manager (libvirt)'''
 
-    def __init__(self, app, offline_mode=None):
+    def __init__(self, offline_mode=None):
         '''
 
         :param offline_mode: enable/disable offline mode; default is to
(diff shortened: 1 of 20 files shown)


qubes/app.py
```diff
@@ -138,20 +139,22 @@ def wrapper(*args, **kwargs):
...
+    def __init__(self, app, offline_mode=None):
         '''
 
         :param offline_mode: enable/disable offline mode; default is to
         enable when running in chroot as root, otherwise disable
         '''
-        self._libvirt_conn = None
-        self._xs = None
-        self._xc = None
+        self.app = app
```
marmarek: Why? It looks unused. But add yet another cyclic reference.

marmarek:
Also, see that tests fails on travis.

PR #148, opened 2017-09-05
Policy related changed for DispVM

marmarek:

Fixes QubesOS/qubes-issues#3048
Fixes QubesOS/qubes-issues#3047
Fixes QubesOS/qubes-issues#3057

### DIFF
--- a/qubespolicy/__init__.py
+++ b/qubespolicy/__init__.py
@@ -196,7 +196,9 @@ class PolicyRule(object):
 
         if self.override_target is not None:
             if self.override_target.startswith('$') and \
-                    not self.override_target.startswith('$dispvm') and \
+                    (not self.override_target.startswith('$dispvm') or
+                    self.override_target.startswith('$dispvm:$tag:') or
+                    self.override_target.startswith('$tag:')) and \
                     self.override_target != '$adminvm':
                 raise PolicySyntaxError(filename, lineno,
                     'target= option needs to name specific target')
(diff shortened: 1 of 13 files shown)


marmarek:
@woju ping?

woju:
Still working on it. Since this added more code paths, I felt that this was too much and I've written an exhaustive fuzzer for qubespolicy.

qubespolicy/__init__.py
```diff
@@ -196,7 +196,9 @@ def __init__(self, line, filename=None, lineno=None):
 
         if self.override_target is not None:
             if self.override_target.startswith('$') and \
-                    not self.override_target.startswith('$dispvm') and \
+                    (not self.override_target.startswith('$dispvm') or
+                    self.override_target.startswith('$dispvm:$tag:') or
+                    self.override_target.startswith('$tag:')) and \
```
woju: I really don't like this condition and I had to write down flow diagram to understand this.

Checking for `'$tag:'` is unneccessary, since if it doesn't start with `'$dispvm'`, it won't even evaluate third `or` expression.

IIUC this would be equivalent:

```python
#@staticmethod
def raise_for_override_target(override_target):
    if self.override_target is None \
            or not self.override_target.startswith('$') \
            or self.override_target == '$adminvm':
        return
    if override_target.startswith('$dispvm') and not override_target.startswith('$dispvm:$'): # or $tag
        return
    raise PolicySyntaxError(...)    
```

Or, if you like it in the same function:

```python
if self.override_target is not None \
        and self.override_target.startswith('$') \
        and self.override_target != '$adminvm' \
        and (not self.override_target.startswith('$dispvm')
            or self.override_target.startswith('$dispvm:$')):
    raise PolicySyntaxError()
```

marmarek: I'll adjust verify_special_value and use that. This way, we'll have all syntax-level validation of special values in one place.

marmarek:
@woju

PR #149, opened 2017-09-11
Fix service qubesd not working with some locales

NicolaiSoeborg:
On certain locales (e.g. danish) `usage_percent` will output a comma-separated number, which will make `attr` point the last two decimal points, s.t. `return vol_info['attr'][4] == 'a'` (in the `verify` func) will fail (`IndexError`) and `qubesd` wont run.

### DIFF
--- a/qubes/storage/lvm.py
+++ b/qubes/storage/lvm.py
@@ -159,7 +159,7 @@ def init_cache(log=logging.getLogger('qubes.storage.lvm')):
     for line in out.splitlines():
         line = line.decode().strip()
         pool_name, pool_lv, name, size, usage_percent, attr, \
-            origin = line.split(';', 7)
+            origin = line.split(';', 6)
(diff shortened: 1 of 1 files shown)


qubes/storage/lvm.py
```diff
@@ -159,11 +159,11 @@ def init_cache(log=logging.getLogger('qubes.storage.lvm')):
     for line in out.splitlines():
         line = line.decode().strip()
         pool_name, pool_lv, name, size, usage_percent, attr, \
-            origin = line.split(',', 6)
+            origin = line.split(';', 7)
```
marmarek: The change 6->7 is wrong here.

NicolaiSoeborg: Correct! Fixed in commit 9d797cc

NicolaiSoeborg:
I think this should be ready to be merged (unless I somehow need to sign the first commit also?)

PR #150, opened 2017-09-19
test fixes around memory leaks

woju:
This PR adds one way any test can fail, i.e. by leaking (keeping references) to some predefined objects. If you don't like that, comment out `self.addCleanup(self.cleanup_gc)` in `QubesTestCase.setUp()`.

### DIFF
--- a/qubes/tests/__init__.py
+++ b/qubes/tests/__init__.py
@@ -374,8 +374,28 @@ class QubesTestCase(unittest.TestCase):
 
     def setUp(self):
         super().setUp()
+        self.addCleanup(self.cleanup_gc)
+
         self.loop = asyncio.get_event_loop()
         self.addCleanup(self.cleanup_loop)
+
+    def cleanup_gc(self):
+        gc.collect()
+        leaked = [obj for obj in gc.get_objects() + gc.garbage
+            if isinstance(obj,
+                (qubes.Qubes, qubes.vm.BaseVM,
+                libvirt.virConnect, libvirt.virDomain))]
+
+        if leaked:
+            try:
+                import objgraph
+                objgraph.show_backrefs(leaked,
+                    max_depth=15, extra_info=extra_info,
+                    filename='/tmp/objgraph-{}.png'.format(self.id()))
+            except ImportError:
+                pass
+
+        assert not leaked
 
     def cleanup_loop(self):
         '''Check if the loop is empty'''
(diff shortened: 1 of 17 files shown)


woju:
This is WIP, there are still more leaks. I strongly suggest `--failfast` or commenting out the cleanup.

qubes/tests/__init__.py
```diff
@@ -371,16 +374,55 @@ def __str__(self):
...
+        # will break when messing with: Python version, kernel family, loop
+        # implementation, a combination thereof, or other things.
+        # KEYWORDS for searching:
+        #   win32, SelectorEventLoop, ProactorEventLoop, uvloop, gevent
+
+        global libvirt_event_impl
+
+        # Check for unfinished libvirt business.
+        if libvirt_event_impl is not None:
+            self.loop.run_until_complete(libvirt_event_impl.drain())
```
marmarek: Is the patch adding `drain()` accepted upstream? AFAIR its original form got rejected.

marmarek:
Also, libvirt stub for travis needs `virConnect` and `virDomain`.

qubes/tests/__init__.py
woju: No, it got rejected and I haven't resubmitted, since upstream was around version bump. I have the updated series though. Would you like it to be submitted upstream, included in core-libvirt or both?

woju:
Sure. Though travis is less relevant, given it will certainly fail on some yet unfixed test case.

qubes/tests/__init__.py
marmarek: Both. Replacing https://github.com/QubesOS/qubes-core-libvirt/pull/4

marmarek:
> Though travis is less relevant, given it will certainly fail on some yet unfixed test case.

Yes, but it's better to see actual bugs there.

woju:
Currently this should break somewhere around `qubes.tests.integ.basic`. Unfortunately this breaks on `.drain()`, which means there are some deeper problems around closing libvirt connection (maybe incl. unregistering events?). I don't know yet what's the problem.

woju:
I'm actually surprised that travis passed.

marmarek:
I noticed that after merging this unit tests (run from git repo, in a VM) are much slower: 13s vs 37s. Is there any _simple_ way to improve this?

PR #151, opened 2017-09-27
linux-stubdom: Set cmdline in HVM config

HW42:
This depends on QubesOS/qubes-vmm-xen#17.

### DIFF
--- a/templates/libvirt/xen.xml
+++ b/templates/libvirt/xen.xml
@@ -42,8 +42,8 @@
                 {% endif %}
                 <kernel>{{ vm.storage.kernels_dir }}/vmlinuz</kernel>
                 <initrd>{{ vm.storage.kernels_dir }}/initramfs</initrd>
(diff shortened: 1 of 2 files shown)


templates/libvirt/xen.xml
```diff
@@ -42,8 +42,8 @@
                 {% endif %}
                 <kernel>{{ vm.storage.kernels_dir }}/vmlinuz</kernel>
                 <initrd>{{ vm.storage.kernels_dir }}/initramfs</initrd>
-                <cmdline>root=/dev/mapper/dmroot ro nomodeset console=hvc0 rd_NO_PLYMOUTH rd.plymouth.enable=0 plymouth.enable=0 {{ vm.kernelopts }}</cmdline>
             {% endif %}
+            <cmdline>root=/dev/mapper/dmroot ro nomodeset console=hvc0 rd_NO_PLYMOUTH rd.plymouth.enable=0 plymouth.enable=0 {{ vm.kernelopts }}</cmdline>
```
marmarek: What about mini-os based stubdomain? Will it ignore this, or fail to start? Or something else?

HW42: It will be silently ignored.

PR #153, opened 2017-09-29
Various fixes and improvements

marmarek:
Fixes QubesOS/qubes-issues#3128
Fixes QubesOS/qubes-issues#3045
Fixes QubesOS/qubes-issues#3114
QubesOS/qubes-issues#3100

### DIFF
--- a/qubes/vm/qubesvm.py
+++ b/qubes/vm/qubesvm.py
@@ -872,15 +872,6 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.BaseVM):
                 self.log.warning('Activating the {} VM'.format(self.name))
                 self.libvirt_domain.resume()
 
-                # close() is not really needed, because the descriptor is
-                # close-on-exec anyway, the reason to postpone close() is that
-                # possibly xl is not done constructing the domain after its main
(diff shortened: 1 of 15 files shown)


qubes/vm/qubesvm.py
```diff
@@ -839,24 +804,38 @@ def start(self, start_guid=True, notify_function=None,
...
+                            yield from self.netvm.start(start_guid=start_guid,
+                                notify_function=notify_function)
+
+                qmemman_client = yield from asyncio.get_event_loop().\
+                    run_in_executor(None, self.request_memory, mem_required)
+
+            except Exception as exc:
+                # let anyone receiving domain-pre-start know that startup failed
+                yield from self.fire_event_async('domain-start-failed',
+                    reason=str(exc))
```
woju: Why is this `reason=str(exc)` and not just `reason=exc`? Someone may extract something useful from the exception object (like `errno` or whatever).

marmarek: This event is mostly for those listening on `admin.Events`. I want to be sure it get serialized using sane way. For those wanting actual exception object (so, running inside qubesd process), there is next line: `raise`.

PR #158, opened 2017-10-12
storage/lvm: remove old volume only after successfully cloning new one

marmarek:
In some cases, it may happen that new volume (`self._vid_snap`) does not
exists. This is normally an error, but even in such a case, do not
remove the only remaining instance of volume (`self.vid`). Instead,
rename it temporarily and remove only after new volume is successfully
cloned.

Fixes QubesOS/qubes-issues#3164

### DIFF
--- a/qubes/storage/__init__.py
+++ b/qubes/storage/__init__.py
@@ -276,6 +276,9 @@ class Volume(object):
     def verify(self):
         ''' Verifies the volume.
 
+        This function is supposed to either return :py:obj:`True`, or raise
+        an exception.
+
         This can be implemented as a coroutine.'''
         raise self._not_implemented("verify")
 
--- a/qubes/storage/lvm.py
+++ b/qubes/storage/lvm.py
@@ -455,11 +455,21 @@ class ThinVolume(qubes.storage.Volume):
 
     def verify(self):
         ''' Verifies the volume. '''
+        if not self.save_on_stop and not self.snap_on_start:
+            # volatile volumes don't need any files
+            return True
+        if self.source is not None:
+            vid = str(self.source)
+        else:
+            vid = self.vid
         try:
-            vol_info = size_cache[self.vid]
-            return vol_info['attr'][4] == 'a'
+            vol_info = size_cache[vid]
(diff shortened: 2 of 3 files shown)


HW42:
Updated my branch with the improved event handling. 

I did not only fix the logic flaw you noticed but also addressed another problem. The previous version relied on that after calling `domainEventDeregisterAny()` the callback is not called anymore. But the libvirt documentation is not clear if this is true or if there might be some unhandled event in the queue which will trigger the callback after unregistering the callback.

When I added a second lock for the stopped event handling (instead of using `startup_lock`) I also noticed that the logic can be (IMO) simplified.

This version should be quite resistant against unreliable stopped events (missing, dupplicated, delayed).

I did not opened a PR yet since I first want to test if I can trigger the race (and then verify my patch).

marmarek:
On Mon, Oct 16, 2017 at 10:03:12PM -0700, HW42 wrote:
> Updated my branch with the improved event handling. 
> 
> I did not only fix the logic flaw you noticed but also addressed another problem. The previous version relied on that after calling `domainEventDeregisterAny()` the callback is not called anymore. But the libvirt documentation is not clear if this is true or if there might be some unhandled event in the queue which will trigger the callback after unregistering the callback.
> 
> When I added a second lock for the stopped event handling (instead of using `startup_lock`) I also noticed that the logic can be (IMO) simplified.
> 
> This version should be quite resistant against unreliable stopped events (missing, dupplicated, delayed).
> 
> I did not opened a PR yet since I first want to test if I can trigger the race (and then verify my patch).

See other commits in
https://github.com/QubesOS/qubes-core-admin/pull/158 - there is a
regression test for this. Currently it tests only if private volume
isn't removed, but it's easy to modify it for events ordering test.

PS Why I can't see this comment on github web interface?

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

HW42:
> PS Why I can't see this comment on github web interface?

I wanted to add this comment to QubesOS/qubes-issues#3164, but accidentally used the wrong tab so the comment was posted on this PR. After noticing my error I deleted the comment here and added it to the issue.

PR #159, opened 2017-10-20
qubes/vm: Improve stopped event handling

HW42:
The previous version did not ensure that the stopped/shutdown event was
handled before a new VM start. This can easily lead to problems like in
QubesOS/qubes-issues#3164.

This improved version now ensures that the stopped/shutdown events are
handled before a new VM start.

Additionally this version should be more robust against unreliable
events from libvirt. It handles missing, duplicated and delayed stopped
events.

Instead of one 'domain-shutdown' event there are now 'domain-stopped'
and 'domain-shutdown'. The later is generated after the former. This way
it's easy to run code after the VM is shutdown including the stop of
it's storage.

### DIFF
--- a/qubes/vm/qubesvm.py
+++ b/qubes/vm/qubesvm.py
@@ -151,9 +151,33 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.BaseVM):
 
             *other arguments are as in :py:meth:`start`*
 
+        .. event:: domain-stopped (subject, event)
+
+            Fired when domain has been stopped.
+
+            This event is emitted before ``'domain-shutdown'`` and will trigger
+            the cleanup in QubesVM. So if you require that the cleanup has
+            already run use ``'domain-shutdown'``.
+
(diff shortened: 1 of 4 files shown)


HW42:
Hmm, there were two pylint errors. The too long line is fixed now. What about "Attribute '\_domain\_stopped\_future' defined outside \_\_init\_\_". Should this warning be suppressed or should I initialize it in `__init__` (to `None`?)?

marmarek:
Better initialize in `__init__` (and use `is not None`, instead of `hasattr`).

qubes/vm/qubesvm.py
```diff
@@ -833,12 +897,16 @@ def start(self, start_guid=True, notify_function=None,
                 # let anyone receiving domain-pre-start know that startup failed
                 yield from self.fire_event_async('domain-start-failed',
                     reason=str(exc))
+                self.storage.stop()
```
marmarek: This is a coroutine. I'll fix this in separate commit after merging it.

PR #166, opened 2017-11-10
Simple server optimizations

qubesuser:
This is a set of optimizations consisting of the changes in the "speed up qvm-ls" branch that don't make architectural or API changes.

The default pool change is particularly impactful since it stops spending 100ms+ with subprocess spawning every time we want to find out what the default pool is.


marmarek:
Oh, you've already did this. See comments on #165.

qubesuser:
Finally got the tests to pass...

PR #167, opened 2017-11-20
vm.run*(): Stop leaking terminal escapes via stderr!

jpouellet:
This restores Qubes R3.2 behavior

Before this patch, the following:

    qvm-run -p sys-firewall 'echo -e "\e[0;46mcyan!" >&2' | wc -l

leaks the escape sequences through to the dom0 terminal via stderr,
in this case demonstrated by the ability to change the text color while
it should be fixed to red.

This can also be abused with xterm reporting sequences to cause input
to be sent to the dom0 terminal! This is potentially a security issue!

### DIFF
--- a/qubes/vm/qubesvm.py
+++ b/qubes/vm/qubesvm.py
@@ -32,6 +32,7 @@ import os.path
 import shutil
 import string
 import subprocess
+import sys
 import uuid
 import warnings
 
@@ -1045,7 +1046,8 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.BaseVM):
         return (yield from asyncio.create_subprocess_exec(
             qubes.config.system_path['qrexec_client_path'],
             '-d', str(self.name),
-            *(('-t', '-T') if filter_esc else ()),
+            *(('-t',) if filter_esc else ()),
+            *(('-T',) if filter_esc or os.isatty(sys.stderr.fileno()) else ()),
             '{}:QUBESRPC {} {}'.format(user, service, source),
             **kwargs))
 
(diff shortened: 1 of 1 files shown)


jpouellet:
The problem here is larger than just qvm-run.

jpouellet:
force-pushed a better fix

jpouellet:
I wonder if always escaping stderr by default and adding another arg to vm.run\*() to explicitly request non-escaped errors would be a better idea in the long run. The only place I'm aware of where 8-bit-clean stderr matters so far is in tar unpacking in backup restoration, where stderr is used for the file list (although I've never in my life seen \x1b in legitimate file names, so even there the concern is minimal IMO).

A quick audit suggests vm.run_\*for_stdio() are safe for now simply because no callers appear to do anything with the errors (except for one in tests which doesn't matter), but I could easily see someone simply quoting the untrusted stderr in an error message someplace.

marmarek:
You noticed that vm.run from this module is not the same as qvm-run (from qubes-core-admin-client), right? It is never connected to the terminal (besides tests). If connected anywhere at all, then to syslog.

jpouellet:
Oops, yeah, this patch was definitely wrong and also outright broken (add -p fail missing import sys). Sorry!

jpouellet:
Commits squashed with different commit msg since it was obviously incorrect.

Originally stated problem *actually* addressed in https://github.com/QubesOS/qubes-core-admin-client/pull/39

Clearly I failed at testing things :/ Thanks for the review! Feel free to just close if you think this is a non-issue.

jpouellet:
A more appropriate test (not qvm-run) is:
```
[user@dom0 ~]$ python3 -c 'import qubes, asyncio; asyncio.get_event_loop().run_until_complete(qubes.Qubes().domains["sys-firewall"].run("echo -e \"\\e[46masdf\\e[0m\" >&2"))' > /dev/null
```

PR #169, opened 2017-11-21
volume.Import: write data with sufficient privs

jpouellet:
Not sure how this ever worked before, if it did.

The device nodes pointed to by /dev/qubes_dom0/* are owned by
root:disk with perms 660, qubes user is not in disk group,
and service is invoked as qubes user, not root.

### DIFF
--- a/qubes-rpc/admin.vm.volume.Import
+++ b/qubes-rpc/admin.vm.volume.Import
@@ -43,7 +43,7 @@ size=$(tail -c +3 "$tmpfile"|cut -d ' ' -f 1)
 path=$(tail -c +3 "$tmpfile"|cut -d ' ' -f 2)
 
 # now process stdin into this path
-if dd bs=4k of="$path" count="$size" iflag=count_bytes,fullblock \
+if sudo dd bs=4k of="$path" count="$size" iflag=count_bytes,fullblock \
         conv=sparse,notrunc,nocreat,fdatasync status=none; then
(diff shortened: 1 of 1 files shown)


jpouellet:
Otherwise you get:

```
Nov 21 03:07:46 dom0 qrexec[10203]: admin.vm.volume.Import+private: mgmt-test -> import-test: allowed to dom0
Nov 21 03:07:46 dom0 admin.vm.volume.Import+private-mgmt-test[10221]: dd: failed to open '/dev/qubes_dom0/vm-import-test-private': Permission denied
Nov 21 03:07:47 dom0 qubesd[30419]: QubesException('Data import failed',) while calling src=b'mgmt-test' meth=b'internal.vm.volume.ImportEnd' dest=b'import-test' arg=b'private' len(untrusted_payload)=4
```

jpouellet:
Another option would be making the pool backends' `import_data()` return something writable by qubes user. Happy to do that instead if you prefer.

marmarek:
This could also fix the issue with (not) overriding LVM volume. import_data() could return some temporary path (in case of LVM - another volume + `chgrp qubes`), writable by the user, then import_data_end() would rename it to the target location. This way we also get import operation being atomic.

jpouellet:
Indeed. More thoughts at https://github.com/QubesOS/qubes-issues/issues/2256#issuecomment-345953267

PR #170, opened 2017-11-29
qubes-rpc: policy.RegisterArgument

woju:
This qrexec is meant for services, which require some kind of
"registering" before use. After registering, the backend should invoke
this call with frontend as the intended destination, with the actual
service in argument of this call and the argument as the payload.

By default this qrexec is disabled by policy.

### DIFF
--- a/qubes-rpc/policy.RegisterArgument
+++ b/qubes-rpc/policy.RegisterArgument
@@ -1,0 +1,95 @@
+#!/usr/bin/env python3
+#
+# The Qubes OS Project, https://www.qubes-os.org/
+#
+# Copyright (C) 2017  Wojtek Porczyk <woju@invisiblethingslab.com>
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, see <https://www.gnu.org/licenses/>.
+#
+
+import logging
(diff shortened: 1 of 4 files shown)


qubes-rpc/policy.RegisterArgument
```diff
@@ -0,0 +1,95 @@
...
+    if any(c not in VALID_CHARS for c in untrusted_argument):
+        die('%s: %s → %s request refused: invalid argument',
+            rpcname, frontend, backend)
+
+    # argument may also be too long, so that length of rpcname, separator and
+    # argument exceed 64 bytes, but that's fine, the call just wont work
+
+    argument = untrusted_argument
+    del untrusted_argument
+    argument = argument.decode('ascii')
```
rootkovska: Some comment why we believe this is safe (because of L#66?)?

qubes-rpc/policy.RegisterArgument
```diff
@@ -0,0 +1,95 @@
...
+
+    filename = '{}+{}'.format(rpcname, argument)
+    logging.debug('%s %s → %s argument %s filename %s',
+            rpcname, frontend, backend, argument, filename)
+
+    try:
+        # the 'x' is critical
```
rootkovska: Make this comment less enigmatic, allowing also others to understand why using the eXlusive creation mode is critical here. I.e. because we want to implement the "stickiness", i.e. allow only to register to the first caller and not allow to re-register anytime later. In fact you should describe the expected behavior of this service somewhere at the beginning of the file, IMHO.

marmarek: There should be some documentation (for administrator) on this service, and this file should also point to it. It should contain info:
 - what this service is for
 - how to write policy for it
 - what policy will be generated when this service is called
 - example call

I think this documentation should be in qubes-doc repo (because it is for administrators, not only for developers).

qubes-rpc/policy.RegisterArgument
woju: It is safe because it throws an error if the encoding is invalid. Default for `errors=` is 'strict', but maybe I should have included it explicitly.

Here is the implementation: https://github.com/python/cpython/blob/3.5/Objects/unicodeobject.c#L6743

PR #171, opened 2017-12-05
Various fixes for tests and the code

marmarek:
This PR fixes some integration tests, including deadlock making test run
hanging. Also some more tests are added.

Besides, there are a couple bugs fixed - mostly detectected by those tests.

Fixes QubesOS/qubes-issues#3037
Fixes QubesOS/qubes-issues#3208
QubesOS/qubes-issues#3207


marmarek:
Re travis fail - TestApp do not have default_fw_netvm mockup. But #172 drop this property altogether, so I don't see reason to fix it here.

PR #172, opened 2017-12-05
Dom0 part of IPv6 support, network related changes

marmarek:
This PR adds IPv6 support, enabled with 'ipv6' feature. By default behaviour
should be unchanged.
To enable it, add 'ipv6' feature to (some) netvm, and all connected VMs will have it enabled - if actual VM code will support it. DNS are still configured over IPv4 in all the cases.

Among other things, this PR also kill default_fw_netvm property (and have
migration code for this), and introduce features.check_with_netvm(), similar to features.check_with_template().

This is based on #171
Fixes QubesOS/qubes-issues#718


qubes/app.py
```diff
@@ -1236,8 +1307,7 @@ def on_property_set_clockvm(self, event, name, newvalue, oldvalue=None):
             del oldvalue.features['service.clocksync']
 
     @qubes.events.handler(
-        'property-pre-set:default_netvm',
-        'property-pre-set:default_fw_netvm')
+        'property-pre-set:default_netvm')
```
woju: Just a nitpick: this could have even been wrapped to a single line.

qubes/config.py
```diff
@@ -108,3 +108,6 @@
 
 #: profiles for admin.backup.* calls
 backup_profile_dir = '/etc/qubes/backup'
+
+#: site-local prefix for all VMs
+qubes_ipv6_prefix = 'fd09:24ef:4179:0000'
```
woju: Where did this come from? (I mean, after `fd09`).

qubes/core2migration.py
```diff
@@ -75,11 +75,6 @@ def load_globals(self, element):
             self.default_netvm = int(default_netvm) \
                 if default_netvm != "None" else None
 
-        default_fw_netvm = element.get("default_fw_netvm")
-        if default_fw_netvm is not None:
-            self.default_fw_netvm = int(default_fw_netvm) \
-                if default_fw_netvm != "None" else None
-
```
woju: Why did this went away? This property is obviously present in whatever legacy xmls, shouldn't it be loaded and coerced like in `Qubes._migrate_however_it_was_called()`?

qubes/vm/mix/net.py
```diff
@@ -24,13 +24,16 @@
...
+import socket
 
 import libvirt  # pylint: disable=import-error
 import qubes
+import qubes.config
 import qubes.events
 import qubes.firewall
 import qubes.exc
 
+QUBES_IPV6_PREFIX = 'fd09:24ef:4179:0000'
```
woju: This is duplicated from `qubes.config`.

qubes/vm/mix/net.py
```diff
@@ -61,6 +64,27 @@ def _setter_ip(self, prop, value):
...
+def _setter_ip6(self, prop, value):
+    # pylint: disable=unused-argument
+    if not isinstance(value, str):
+        raise ValueError('IPv6 address must be a string')
+    value = value.lower()
+    try:
+        socket.inet_pton(socket.AF_INET6, value)
+    except socket.error:
+        raise ValueError('Invalid IPv6 address value')
+    return value
```
woju: May I suggest using `type=ipaddress.IPv6Address` instead of this whole setter? It has validation built-in, and str(address) works well for serialising.

qubes/config.py
marmarek: Randomly generated, as required by https://tools.ietf.org/html/rfc4193#section-3.2. The same for all Qubes instances, to not allow specific instance fingerprinting based on this.

qubes/core2migration.py
marmarek: In practice it wasn't used for anything but creating new ProxyVMs. You may think that it was value for `QubesProxyVM.netvm` if `QubesProxyVM.uses_default_netvm==True`, but the later was hardwired to `False`.

qubes/vm/mix/net.py
marmarek: Should I do the same for `ip` property? AFAIR there was some reason why I didn't wanted to use it before, but cannot find one right now. Note that this code is may be exposed to untrusted input, but I assume input validation in `ipaddress` module is safe enough.

marmarek: Replaced for both, check now.

woju: That was a question of compatibility with older Python versions, as `ipaddress` is a relatively recent addition to stdlib (it was a 3rd party package before then). We thought about it before qubes.

marmarek: Ah, it was probably before we've switched to python3. So, that point is mute now.

PR #178, opened 2018-01-09
R3.2: PVH/HVM support

HW42:
Related PRs:
 - https://github.com/QubesOS/qubes-core-admin-linux/pull/35
 - https://github.com/QubesOS/qubes-core-libvirt/pull/8
 - https://github.com/QubesOS/qubes-linux-utils/pull/32
 - https://github.com/QubesOS/qubes-vmm-xen-stubdom-linux/pull/13
 - https://github.com/QubesOS/qubes-linux-kernel/pull/13 (for R3.2 and R4.0)

Additionally needed:
 - vmm-xen: build branch xen-4.8 for R3.2
 - core-vchan-xen: rebuild against new xen-libs
 - gui-daemon: rebuild against new xen-libs

Missing:
 - Migration handling. Currently every domain will be pvh after installing the update.

### DIFF
--- a/core/qubes.py
+++ b/core/qubes.py
@@ -53,6 +53,7 @@ dry_run = False
 if not dry_run:
     import libvirt
     try:
+        import xen.lowlevel.xc
         import xen.lowlevel.xs
     except ImportError:
         pass
@@ -136,6 +137,7 @@ class QubesVMMConnection(object):
         self._offline_mode = False
         self._hvm_supported = None
         self._hap_supported = None
+        self._pvh_supported = None
 
     @property
     def offline_mode(self):
@@ -159,6 +161,8 @@ class QubesVMMConnection(object):
             # Do not initialize in offline mode
             return
 
+        if 'xen.lowlevel.xc' in sys.modules:
+            self._xc = xen.lowlevel.xc.xc()
         if 'xen.lowlevel.xs' in sys.modules:
             self._xs = xen.lowlevel.xs.xs()
         self._libvirt_conn = libvirt.open(defaults['libvirt_uri'])
@@ -181,6 +185,13 @@ class QubesVMMConnection(object):
         return self._common_getter('_libvirt_conn')
 
     @property
+    def xc(self):
+        if 'xen.lowlevel.xc' in sys.modules:
+            return self._common_getter('_xc')
+        else:
+            return None
+
+    @property
     def xs(self):
         if 'xen.lowlevel.xs' in sys.modules:
             return self._common_getter('_xs')
@@ -198,10 +209,16 @@ class QubesVMMConnection(object):
     @property
     def hap_supported(self):
         if self._hap_supported is None:
-            caps_xml = lxml.etree.fromstring(self.libvirt_conn.getCapabilities())
-            self._hap_supported = bool(
-                    caps_xml.xpath('./guest/features/hap'))
+            xen_caps = self.xc.xeninfo()['xen_caps'].split()
+            self._hap_supported = 'qubes-hap' in xen_caps
         return self._hap_supported
+
+    @property
+    def pvh_supported(self):
+        if self._pvh_supported is None:
+            xen_caps = self.xc.xeninfo()['xen_caps'].split()
+            self._pvh_supported = 'qubes-pvh' in xen_caps
+        return self._pvh_supported
 
 
 ##### VMM global variable definition #####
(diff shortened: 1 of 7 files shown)


marmarek:
As for migration, "default" value in 3.2 cannot be callable, but there can be 'func' or 'eval', which can convert arbitrary 'default' value to actual default value based on other options. You can even reuse `__virt_mode` function. Something like this:
```python
def __virt_mode(self, value):
  if value == 'default':
    # consider caching this
    caps_xml = lxml.etree.fromstring(vmm.libvirt_conn.getCapabilities())
    if caps_xml.xpath('./guest/os_type[text()=\'hvm\']'):
      if not self.pcidevs:
        value = 'pvh'
      else:
        value = 'pv'
    else:
      value = 'pv'
  ...
```
Make sure to order this after loading pcidevs.

HW42:
Hmm, not sure I understand your snippet correctly. Do you want a `default` `virt_mode` which automatically selects `pvh`/`pv` or a one time migration when the VM is loaded without any value set yet? The code would do the later but the naming `default` suggest the former.

marmarek:
Yes, one time migration (applicable also for new VMs).

HW42:
Ok, implemented. This has the side-effect that you can run `qvm-run -s ${vm} virt_mode default` which has quite a different behavior than the other `default` values (template/kernel). But I think that's ok.

marmarek:
Detection code needs to be smarter, to check also for EPT/HAP. I have one machine, where `xl dmesg` reports:
```
HVM: VMX enabled
HVM: Hardware Assisted Paging (HAP) not detected
HVM: PVH mode not supported on this platform
```
This is Thinkpad T61, Intel Core2 Duo T7300. A 10 years old machine, but it meets minimum requirements for R3.2...

marmarek:
Ough, updating to Xen 4.8 breaks HVM there too. It's because in 4.8 we explicitly exclude shadow paging compile time. I think for R3.2 we shouldn't do that.

marmarek:
Interesting: if I build Xen with shadow paging enabled, I still get above messages, but PVH starts successfully. But very slow - about 4 minutes to the login prompt on `xl console` (this is long enough to qrexec timeout on startup). Compared to 16s for PV on the same hardware.

I'd say we shouldn't automatically switch to PVH if HAP/EPT is not available.

marmarek:
Part of https://github.com/QubesOS/qubes-issues/issues/3460

core/qubes.py
```diff
@@ -194,6 +195,14 @@ def hvm_supported(self):
                     caps_xml.xpath('./guest/os_type[text()=\'hvm\']'))
         return self._hvm_supported
 
+    @property
+    def hap_supported(self):
+        if self._hap_supported is None:
+            caps_xml = lxml.etree.fromstring(self.libvirt_conn.getCapabilities())
+            self._hap_supported = bool(
+                    caps_xml.xpath('./guest/features/hap'))
```
marmarek: This isn't that simple - on non HAP capable system the hap node is still present: https://gist.github.com/marmarek/05a64b2d3dd23b6d6d35b87d2d008708

HW42:
> This isn't that simple - on non HAP capable system the hap node is still present: https://gist.github.com/marmarek/05a64b2d3dd23b6d6d35b87d2d008708

Hmm, this is a bit tricky. AFAICS the HAP capability is only exposed in `xl dmesg` (and `xl dmesg` looses old messages). I thought about the following options:

 1. Parse `xl dmesg` early and then store the result somewhere (say `/run/qubes/hap-supported`).
 2. Minimal patch to Xen to expose it for our purpose.
 3. Proper Xen/libxl/libvirt patch (probably a new `physinfo.capabilities` flag)

and in theory:

 4. Start a dummy domain to test.

(4) is obviosly a bad idea. I don't like (1) since you need to get the write/read order right (for example by adding locking in `hap_supported`). For now I choose (2) (see QubesOS/qubes-vmm-xen#23) since this doesn't add an ABI change to libxl like (3). Probably we should take this to xen-devel for a long-term solution.

HW42:
PS: I currently don't have hardware available to test the relevant case (64 bit, VT-x, no EPT).

marmarek:
Regarding this approach, see `r3.2-hv` branch on my account. I've fixed confusing virt_mode behavior of   virt_mode for QubesHVm (allowing to set pv, then being ignored).

marmarek:
Closing now, because alternative approach was merged. But keep the branch around if we'd need to backport PVH later.

PR #184, opened 2018-01-19
pci: Add permissive options

HW42:
Depends on https://github.com/QubesOS/qubes-core-libvirt/pull/9

BTW: If I add such dependencies should I also modify the .spec file? OTOH: Depending on the merge order I don't know the correct version yet.


marmarek:
Just adding a note in PR description is ok.

PR #188, opened 2018-01-23
file-reflink, a storage driver optimized for CoW filesystems

rustybird:
This adds the file-reflink storage driver. It is never selected automatically for pool creation, especially not the creation of `varlibqubes` (though it can be used if set up manually).

The code is quite small:

 |               | reflink.py | lvm.py     | file.py + block-snapshot |
 | ------------- | ---------- | ---------- | ------------------------ |
 | **sloccount** | 334 lines  | 447 (134%) | 570 (171%)               |

Background: btrfs and XFS (but not yet ZFS) support instant copies of individual files through the `FICLONE` ioctl behind `cp --reflink`. Which file-reflink uses to snapshot VM image files without an extra device-mapper layer. All the snapshots are essentially freestanding; there's no functional origin vs. snapshot distinction.

In contrast to 'file'-on-btrfs, file-reflink inherently avoids CoW-on-CoW. Which is a bigger issue now on R4.0, where even AppVMs' private volumes are CoW. (And turning off the lower, filesystem-level CoW for 'file'-on-btrfs images would turn off data checksums too, i.e. protection against bit rot.)

Also in contrast to 'file', all storage features are supported, including

   - any number of `revisions_to_keep`
   - `volume.revert()`
   - `volume.is_outdated`
   - online `fstrim`/`discard`

Example tree of a file-reflink pool - `*-dirty.img` are connected to Xen:

  - `/var/lib/testpool/appvms/foo/volatile-dirty.img`
  - `/var/lib/testpool/appvms/foo/root-dirty.img`
  - `/var/lib/testpool/appvms/foo/root.img`
  - `/var/lib/testpool/appvms/foo/private-dirty.img`
  - `/var/lib/testpool/appvms/foo/private.img`
  - `/var/lib/testpool/appvms/foo/private.img@2018-01-02T03:04:05Z`
  - `/var/lib/testpool/appvms/foo/private.img@2018-01-02T04:05:06Z`
  - `/var/lib/testpool/appvms/foo/private.img@2018-01-02T05:06:07Z`
  - `/var/lib/testpool/appvms/bar/...`
  - `/var/lib/testpool/appvms/...`
  - `/var/lib/testpool/template-vms/fedora-26/...`
  - `/var/lib/testpool/template-vms/...`

It looks similar to a 'file' pool tree, and in fact file-reflink is drop-in compatible:

    $ qvm-shutdown --all --wait
    $ systemctl stop qubesd
    $ sed 's/ driver="file"/ driver="file-reflink"/g' -i.bak /var/lib/qubes/qubes.xml
    $ systemctl start qubesd
    $ sudo rm -f /path/to/pool/*/*/*-cow.img*

If the user tries to create a fresh file-reflink pool on a filesystem that doesn't support reflinks, qvm-pool will abort and mention the `setup_check=no` option. Which can be passed to force a fallback on regular sparse copies, with of course lots of time/space overhead. The same fallback code is also used when initially cloning a VM from a foreign pool, or from another file-reflink pool on a different mountpoint.

`journalctl -fu qubesd` will show all file-reflink copy/rename/remove operations on VM creation/startup/shutdown/etc.

---
TODO: port unit tests

### DIFF
--- a/qubes/storage/reflink.py
+++ b/qubes/storage/reflink.py
@@ -1,0 +1,423 @@
+#
+# The Qubes OS Project, https://www.qubes-os.org/
+#
+# Copyright (C) 2018 Rusty Bird <rustybird@net-c.com>
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, see <https://www.gnu.org/licenses/>.
+#
+
+''' Driver for handling VM images as files, without any device-mapper
+    involvement. A reflink-capable filesystem is strongly recommended,
+    but not required.
+'''
+
+import collections
+import errno
+import fcntl
+import glob
+import logging
+import os
+import re
+import subprocess
+import tempfile
+from contextlib import contextmanager, suppress
+
+import qubes.storage
+
+BLKSIZE = 512
+FICLONE = 1074041865  # see ioctl_ficlone manpage
+LOGGER = logging.getLogger('qube.storage.reflink')
+
+
+class ReflinkPool(qubes.storage.Pool):
+    driver = 'file-reflink'
+    _known_dir_path_prefixes = ['appvms', 'vm-templates']
+
+    def __init__(self, dir_path, setup_check='yes', revisions_to_keep=1,
+                 **kwargs):
+        super().__init__(revisions_to_keep=revisions_to_keep, **kwargs)
+        self._volumes = {}
+        self.dir_path = os.path.abspath(dir_path)
+        self.setup_check = qubes.property.bool(None, None, setup_check)
+
+    def setup(self):
+        created = _make_dir(self.dir_path)
+        if self.setup_check and not is_reflink_supported(self.dir_path):
+            if created:
+                _remove_empty_dir(self.dir_path)
+            raise qubes.storage.StoragePoolException(
+                'The filesystem for {!r} does not support reflinks. If you'
+                ' can live with VM startup delays and wasted disk space, pass'
+                ' the "setup_check=no" option.'.format(self.dir_path))
+        for dir_path_prefix in self._known_dir_path_prefixes:
+            _make_dir(os.path.join(self.dir_path, dir_path_prefix))
+        return self
+
+    def init_volume(self, vm, volume_config):
+        # Fail closed on any strange VM dir_path_prefix, just in case
+        # /etc/udev/rules/00-qubes-ignore-devices.rules needs updating
+        assert vm.dir_path_prefix in self._known_dir_path_prefixes, \
+               'Unknown dir_path_prefix {!r}'.format(vm.dir_path_prefix)
+
+        volume_config['pool'] = self
+        if 'revisions_to_keep' not in volume_config:
+            volume_config['revisions_to_keep'] = self.revisions_to_keep
+        if 'vid' not in volume_config:
+            volume_config['vid'] = os.path.join(vm.dir_path_prefix, vm.name,
+                                                volume_config['name'])
+        volume = ReflinkVolume(**volume_config)
+        self._volumes[volume_config['vid']] = volume
+        return volume
+
+    def list_volumes(self):
+        return list(self._volumes.values())
+
+    def get_volume(self, vid):
+        return self._volumes[vid]
+
+    def destroy(self):
+        pass
+
+    @property
+    def config(self):
+        return {
+            'name': self.name,
+            'dir_path': self.dir_path,
+            'driver': ReflinkPool.driver,
+            'revisions_to_keep': self.revisions_to_keep
+        }
+
+    @property
+    def size(self):
+        statvfs = os.statvfs(self.dir_path)
+        return statvfs.f_frsize * statvfs.f_blocks
+
+    @property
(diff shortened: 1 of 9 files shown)


rustybird:
Does CI policy require no new pylint warnings? There's one [protected-access](https://github.com/rustybird/qubes-core-admin/blob/b3a24c8fd7d703d248c830de419a8a039a4f0ebe/qubes/storage/reflink.py#L327) warning that I had planned on leaving in, because the accessed attribute isn't mine.

(I've also fixed one legit warning and silenced one false positive error. They weren't printed by my local pylint version.)

rustybird:
> There's one protected-access warning that I had planned on leaving in, because the accessed attribute isn't mine.

On second thought, let me handle that more cleanly...

marmarek:
> Does CI policy require no new pylint warnings?

Generally yes. But in some specific cases, muting with `# pylint: disable=...` is ok too. It's here to at least force developer to look at those cases.

rustybird:
I've fixed the underlying cause of the warning (and added a previously missing `fsync` after `flush`).

rustybird:
> In contrast to 'file'-on-btrfs, file-reflink inherently avoids CoW-on-CoW.

Huh, that's not actually true for the `root` volume of TemplateBasedVMs and DispVMs - even in R4.0, they still implement their own device-mapper based CoW thing, `/dev/mapper/dmroot`, _inside_ the VM, using the second partition of `volatile`. Is my understanding correct that this approach is simply a remnant of the pre-R4.0 days? @marmarek @kalkin

If so, I propose to make these `root` volumes work exactly like the `private` volume of DispVMs, using a two step transition:

1. Flip their `root` volume from read-only to read-write in dom0.
2. At some later point (maybe R4.1), rip out the `dmroot` stuff inside the VMs.

Old e.g. backed up VMs still using `dmroot` would continue to work after (2).

marmarek:
> Huh, that's not actually true for the root volume of TemplateBasedVMs and DispVMs - even in R4.0, they still implement their own device-mapper based CoW thing, /dev/mapper/dmroot, inside the VM, using the second partition of volatile. Is my understanding correct that this approach is simply a remnant of the pre-R4.0 days? @marmarek @kalkin

This is intentionally left in place, to allow VM to verify original root image. This would allow semi-untrusted storage domain. See original architecture paper. This part is yet to be implemented...

Since this is unused right now, I think its ok to disable it, but not completely remove. And leave clear comment why it is this way. So:

> 1. Flip their root volume from read-only to read-write in dom0.

Yes.

> 2. At some later point (maybe R4.1), rip out the dmroot stuff inside the VMs.

No. And not really needed, as currently if /dev/xvda is read-write, then /dev/mapper/dmroot is just a symlink (not dm-linear as it was in R3.2).

BTW Are you waiting for me on this PR? I've skimmed through it and looks fine, but haven't done careful review yet. Are you going to port tests in this PR, or separate at some future time?

rustybird:
> This would allow semi-untrusted storage domain.

Ah, okay.

> Since this is unused right now, I think its ok to disable it, but not completely remove. And leave clear comment why it is this way. So:
>
>>  Flip their root volume from read-only to read-write in dom0.
>
> Yes.

[Done.](https://github.com/QubesOS/qubes-core-admin/pull/188/commits/093a78bbed98d5d257289bfdbde8b36e7a4604d2#diff-f7a6cac9c0676df6ac38944ea99cdbbbR80)

>>  At some later point (maybe R4.1), rip out the dmroot stuff inside the VMs.
> 
> No. And not really needed, as currently if /dev/xvda is read-write, then /dev/mapper/dmroot is just a symlink (not dm-linear as it was in R3.2).

That's fantastic!

> BTW Are you waiting for me on this PR? I've skimmed through it and looks fine, but haven't done careful review yet. Are you going to port tests in this PR, or separate at some future time?

If possible, I'd implement tests in a separate PR, before submitting a third one to automatically use file-reflink (instead of 'file') for varlibqubes when a btrfs-without-LVM layout has been set up during R4.1 installation. As for this first PR, getting it merged would conveniently stop breaking my qubesd during qubes-core-admin package upgrades. 8)

Hey while I have your ear, should `is_outdated()` return True if the source volume has been started but not yet been stopped since the snapshot? file-reflink currently returns True only after source volume commit (because restarting the consumer volume would still give the same data), but I can change that.

marmarek:
> As for this first PR, getting it merged would conveniently stop breaking my qubesd during qubes-core-admin package upgrades. 8)

Not sure if you know, but you can install additional storage pool drivers from outside of qubes-core-admin. Just provide appropriate entry point (as you've done here in setup.py). That wouldn't work for default_pool, but you can set it using `qubes-prefs`.

> file-reflink currently returns True only after source volume commit (because restarting the consumer volume would still give the same data),

That's ok.

qubes/storage/reflink.py
```diff
@@ -0,0 +1,434 @@
...
+                                                volume_config['name'])
+        if volume_config['name'] == 'root':
+            # When root is read-only, the VM itself will set up a
+            # device-mapper CoW layer at /dev/mapper/dmroot, writing
+            # modified blocks to the volatile volume. That volume
+            # usually does CoW on our layer too, thus creating an
+            # inefficient CoW-on-CoW setup. Avoid this by always
+            # overriding root to be read-write - which may become
+            # incompatible with a future untrusted storage domain!
+            volume_config['rw'] = True
```
marmarek: Better patch it directly where [volume_config is defined](https://github.com/qubesos/qubes-core-admin/blob/master/qubes/vm/appvm.py#L48) (and similar for other VM types).

marmarek: And please separate commit for such changes.

rustybird:
> Not sure if you know, but you can install additional storage pool drivers from outside of qubes-core-admin. Just provide appropriate entry point (as you've done here in setup.py).

Thanks, I'll give that a try.

qubes/storage/reflink.py
rustybird: [Done.](https://github.com/QubesOS/qubes-core-admin/pull/188/commits/958d6c7067138f82e602b18ee8849bbe7ef0a565)

PR #190, opened 2018-02-01
Added default_user property to qubesdb

yilmi:
Added the default_user property from the Qube to the qubesdb so it is available when starting X. This is the 1st part of a fix for issue https://github.com/QubesOS/qubes-issues/issues/2372


qubes/vm/qubesvm.py
```diff
@@ -1798,6 +1798,7 @@ def create_qdb_entries(self):
 
         self.untrusted_qdb.write('/name', self.name)
         self.untrusted_qdb.write('/type', self.__class__.__name__)
+        self.untrusted_qdb.write('/default_user', self.default_user)
```
marmarek: Please use dash in entry name, same as in other entries.

PR #197, opened 2018-02-22
Don't fire domain-stopped/-shutdown while VM is still Dying

rustybird:
Lots of code expects the VM to be Halted after receiving one of these events, but it could also be Dying or Crashed. Get rid of the Dying case at least, by waiting until the VM has transitioned out of it.

Fixes e.g. the following DispVM cleanup bug:

    $ qvm-create -C DispVM --prop auto_cleanup=True -l red dispvm
    $ qvm-start dispvm
    $ qvm-shutdown --wait dispvm  # this won't remove dispvm
    $ qvm-start dispvm
    $ qvm-kill dispvm  # but this will

### DIFF
--- a/qubes/vm/qubesvm.py
+++ b/qubes/vm/qubesvm.py
@@ -975,6 +975,8 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.BaseVM):
             # an exception gets thrown.
             self._domain_stopped_event_handled = True
 
+            while self.get_power_state() == 'Dying':
+                yield from asyncio.sleep(0.25)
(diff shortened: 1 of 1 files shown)


qubes/vm/qubesvm.py
```diff
@@ -975,6 +975,8 @@ def _domain_stopped_coro(self):
             # an exception gets thrown.
             self._domain_stopped_event_handled = True
 
+            while self.get_power_state() == 'Dying':
+                asyncio.sleep(0.25)
```
marmarek: yield from asyncio.sleep

rustybird: Gah, fixed.

PR #198, opened 2018-02-22
include_in_backups fixes

rustybird:


marmarek:
Any idea why restore of DispVM fails? Template for DispVM should be an AppVM, not TemplateVM. Maybe that template is missing and app.default_template is used instead?

rustybird:
> Any idea why restore of DispVM fails? Template for DispVM  should be an AppVM, not TemplateVM. Maybe that template is missing and  app.default_template is used instead?

Yes, that's it. Working on it...

rustybird:
Filtered out the commit that forbade DispVMs

PR #200, opened 2018-03-08
Fixes in firewall temporary access 

donob4n:
Fix https://github.com/QubesOS/qubes-issues/issues/3661 and another related problems.

### DIFF
--- a/qubes/firewall.py
+++ b/qubes/firewall.py
@@ -204,7 +204,7 @@ class SpecialTarget(RuleChoice):
 class Expire(RuleOption):
     def __init__(self, untrusted_value):
         super(Expire, self).__init__(untrusted_value)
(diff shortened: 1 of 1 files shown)


marmarek:
If qubes-manager use local time, that bug there. But it shouldn't matter, because at API level, `%s` format is used, which is defined as "seconds since 1970-01-01 00:00:00 UTC", so it is timezone agnostic.
Relevant line: https://github.com/QubesOS/qubes-manager/blob/master/qubesmanager/firewall.py#L338

Comparing utc or local time here shouldn't matter. Unless you change timezone in the middle of the operation. But that in practice probably require qubesd restart anyway, so the time delta will be recalculated. But there was indeed one inconsistency - time delta was calculated by comparing with `datetime.now()`, not `datetime.utcnow()`. I'm ok with changing it in the direction you propose, instead of fixing the other way around.

PR #202, opened 2018-03-13
api/admin: fix backup exclude

rustybird:
Bugfix on 59abdeb

### DIFF
--- a/qubes/api/admin.py
+++ b/qubes/api/admin.py
@@ -1347,10 +1347,10 @@ class QubesAdminAPI(qubes.api.AbstractQubesAPI):
                 if any(qubes.utils.match_vm_name_with_special(vm, name)
                     for name in include_vms))
 
-            # handle exclude
-            vms_to_backup.difference_update(vm for vm in self.app.domains
-                if any(qubes.utils.match_vm_name_with_special(vm, name)
-                    for name in exclude_vms))
+        # handle exclude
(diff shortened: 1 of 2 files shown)


qubes/api/admin.py
```diff
@@ -1347,11 +1347,6 @@ def _load_backup_profile(self, profile_name, skip_passphrase=False):
                 if any(qubes.utils.match_vm_name_with_special(vm, name)
                     for name in include_vms))
 
-            # handle exclude
-            vms_to_backup.difference_update(vm for vm in self.app.domains
-                if any(qubes.utils.match_vm_name_with_special(vm, name)
-                    for name in exclude_vms))
```
marmarek: This is wrong, because `exclude_vms` argument is expected to be a list of VM names, but `exclude` in [backup profile](https://www.qubes-os.org/doc/admin-api/#backup-profile) may also contain special keywords like `@tag:something`. Probably the easiest thing to do is to construct exclude list based on all VMs here (like it is here right now), but save it into `exclude_vms` instead of applying directly to `vms_to_backup`.

rustybird: Sorry, fixed!

`test_621_qdb_vm_with_network` is [racy](https://travis-ci.org/QubesOS/qubes-core-admin/builds/355669764#L2475) BTW.

PR #203, opened 2018-03-17
Improve LVM volume revisions handling

marmarek:
The main change here is to keep the most recent volume version (the current
one) in LV with revision number already appended. And abandon -back suffix.
This means that committing new revision is just renaming LV to the one with
largest revision number. One atomic operation, no need to worry about
interrupted commits, parallel commits etc. And also faster.
This require unambitious revision ordering. Previously a timestamp was used,
but this is susceptible to time changes. Since now it could result in data
loss, something better is needed. In addition to timestamp, prefix revision id
with monotonous number.

Since this change make volume commit simpler and cheaper, use it also for other
cases. First of all, when importing new volume content, do it into temporary
volume and only commit it when operation is finished (successfully). This has multiple benefits, including security relevant ones - see aae56fec for details.

Since this is potentially fragile change, also add multiple new tests for LVM
storage driver.

QubesOS/qubes-issues#2256

### DIFF
--- a/qubes/storage/lvm.py
+++ b/qubes/storage/lvm.py
@@ -90,6 +90,12 @@ class ThinPool(qubes.storage.Pool):
 
         self._volume_objects_cache = {}
 
+    def __repr__(self):
+        return '<{} at {:#x} name={!r} volume_group={!r} thin_pool={!r}>'.\
+            format(
+                type(self).__name__, id(self),
+                self.name, self.volume_group, self.thin_pool)
+
     @property
     def config(self):
         return {
--- a/qubes/storage/lvm.py
+++ b/qubes/storage/lvm.py
@@ -20,7 +20,6 @@
 ''' Driver for storing vm images in a LVM thin pool '''
 
 import logging
-import operator
 import os
 import subprocess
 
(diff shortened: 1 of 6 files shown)


marmarek:
@rustybird can you take a look at this PR?

jpouellet:
I don't have time to do a proper implementation review, but :+1: on the apparent intent.

qubes/storage/lvm.py
```diff
@@ -263,68 +360,94 @@ def _remove_revisions(self, revisions=None):
...
-        # remove old volume only after _successful_ clone of the new one
-        cmd = ['rename', self.vid, self.vid + '-tmp']
+        # then assign new revision number to the volume we're committing right
+        # now
+        cmd = ['clone' if keep else 'rename',
+               vid_to_commit,
+               '{}-{}.{}'.format(
+                self.vid, self._get_next_revision_number(),
+                int(time.time()))]
         qubes_lvm(cmd, self.log)
```
rustybird: I don't have an LVM installation at hand (and I'm a bit out of my depth in the `lvm.py` code) - but it looks like the revision is now created directly from the dirty data after shutdown, instead of from the clean data committed during the _previous_ shutdown? Changing the meaning of `revisions_to_keep` (because there's no longer an implicit not-yet-a-revision image), and making `revert()` without an explicit `revision` argument just move the latest data to a "new" revision with a clobbered timestamp.

rustybird: Monotonic revision numbers are a nice touch. I shied away from those when coding the reflink driver, IIRC due to being nervous about leaving an obvious VM startup counter around in case of dom0 compromise. But ordering revisions correctly when the clock has changed seems more pressing than that. Maybe a random initial offset could be added, or something? Anyway, I've ported the monotonic design to reflink.

But it might make sense to keep the current dirty -> clean -> revision data lifecycle, which IMO leads itself to a more natural definition of `revisions_to_keep` and `revert`.

marmarek: This is preserved - see that the most recent is excluded from self.revisions.

marmarek: The whole point of this change is to have commit a single step operation, instead of two (mark previous "most recent" as "old", then dirty -> "most recent").

qubes/storage/lvm.py
```diff
@@ -202,34 +256,77 @@ def __init__(self, volume_group, size=0, **kwargs):
...
     @property
     def path(self):
-        return '/dev/' + self.vid
+        if self._vid_current is None:
+            self._refresh_vid_current()
+        return '/dev/' + self._vid_current
 
-    @property
-    def revisions(self):
+    def _get_revisions(self, suffix):
```
rustybird: > This is preserved - see that the most recent is excluded from self.revisions.

 I totally missed that there's now a wrapper function for the public interface. :+1: FWIW then.

PR #209, opened 2018-03-30
qubes/tests: use loadTestsFromNames for nose2 compat

woju:


woju:
There is a bit of dark arts around `create_testcases_...` and especially `qubes.tests.extra.for_templates`. Tell if you're comfortable with the amount.

marmarek:
According to travis, build fails on fc25 with "Invalid syntax"...

qubes/tests/__init__.py
```diff
@@ -1127,6 +1127,50 @@ def list_templates():
             _templates = ()
     return _templates
 
+def create_testcases_for_templates(name, *bases, globals, **kwds):
+    '''Do-it-all helper for generating per-template tests via load_tests proto
+
+    This does several things:
+        - creates per-template classes
+        - adds them to module's :py:func:`globals`
```
marmarek: What about adding them to module object instead (`sys.modules[...]`)? As you've done for `extra` tests. IMHO this would be slightly less hacky... You can then pass simply a module name (to avoid looking at stack, as you've explained below).

qubes/tests/__init__.py
```diff
@@ -1127,6 +1127,50 @@ def list_templates():
             _templates = ()
     return _templates
 
+def create_testcases_for_templates(name, *bases, globals, **kwds):
```
marmarek: If `globals` argument is needed (see below), I'd name it differently, to not cover `globals` builtin.

qubes/tests/__init__.py
woju: This is the same:
```python
import sys
assert globals() is sys.modules[__name__].__dict__
```
And explicit is better than implicit™ [PEP-20]. The reason for `extra` is that there is no other viable place to add this to (endpoint loading is purposefully constrained).

qubes/tests/__init__.py
woju: It is named such for the function's signature, so the caller has to name it globals. It makes error more readable:
```
TypeError: create_testcases_for_templates() missing 1 required keyword-only argument: 'globals'
```
If you'd like, you can `import builtins` or even `from builtins import globals as _globals` or whatever.

qubes/tests/__init__.py
marmarek: Still, IMO better use sys.modules, makes code less magic (easier to understand for non-python-wizard) and more flexible. Right now it greatly depends on the _place_ where `globals()` is called, using `sys.modules` would avoid this.

marmarek:
@woju ?

marmarek:
But then you extract module anyway, Specifically to update `__module__`. Just to be clear, my proposal is to have:
```python
def create_testcases_for_templates(name, *bases, module, **kwds):
    ....

def load_tests(loader, tests, pattern):
    tests.addTests(loader.loadTestsFromNames(
        qubes.tests.create_testcases_for_templates('TC_10_BackupVM',
            TC_10_BackupVMMixin, qubes.tests.SystemTestCase,
            module=sys.modules[__name__])))
```

marmarek:
Anyway, the current version doesn't compile...

marmarek:
Anyway, the current version doesn't compile...

> Anyway, the current version doesn't compile...

@woju?

PR #212, opened 2018-06-12
Do not use assert statement in security logic

woju:
This is because assert statement gets optimised out when Python is run
with -O flag. This was pointed out to me by audience at PyWaw 76.


marmarek:
See travis - minor pylint nit, and many tests to fix.
Generally the change looks good. Do you it worth adding some message to the exception, at least in most common cases ("unexpected argument", "not dom0 target for dom0-only call" etc)? We didn't have it before, so maybe not?

marmarek:
@woju bump

PR #216, opened 2018-07-14
sdwdate-gui-qubes qrexec policy

adrelanos:
Using `sys-whonix $tag:anon-vm allow` verbatim in case anyone is missing the tag. That sorts out at least most users who just go with the defaults.

(As far I know qubes-core-admin-addon-whonix does not have a mechanism to add tags for already created VMs.)

References:

* https://forums.whonix.org/t/sdwdate-gui-for-qubes-testers-wanted-developpers-welcome
* https://phabricator.whonix.org/T534

### DIFF
--- a/qubes-rpc-policy/whonix.GatewayCommand
+++ b/qubes-rpc-policy/whonix.GatewayCommand
@@ -1,0 +1,3 @@
+$tag:anon-gateway $tag:anon-vm allow
+$anyvm $anyvm deny        
+
--- a/qubes-rpc-policy/whonix.GatewayCommand
+++ b/qubes-rpc-policy/whonix.GatewayCommand
@@ -1,3 +1,4 @@
 $tag:anon-gateway $tag:anon-vm allow
+sys-whonix $tag:anon-vm allow
 $anyvm $anyvm deny        
 
--- a/qubes-rpc-policy/whonix.NewStatus
+++ b/qubes-rpc-policy/whonix.NewStatus
@@ -1,0 +1,3 @@
+$tag:anon-vm $tag:anon-gateway allow
+$anyvm $anyvm deny  
+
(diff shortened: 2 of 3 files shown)


marmarek:
Looks fine, but could you put that into qubes-core-admin-addon-whonix package?

> (As far I know qubes-core-admin-addon-whonix does not have a mechanism to add tags for already created VMs.)

Well, technically it is possible - scan domains when loading them from disk ([domain-load event](https://dev.qubes-os.org/projects/core-admin/en/latest/qubes-vm/qubesvm.html#event-domain-load)). But in practice it's going to be similar thing (looking for `sys-whonix` name), just more obscure.
Might be a good idea to do that anyway, if `whonix-gateway` tag would be needed for anything else.

marmarek:
I assume those VMs will locate each other some other mechanism, so anon-whonix will know to call to appropriate whonix gateway, even if user create custom one, right?

adrelanos:
> I assume those VMs will locate each other some other mechanism, so anon-whonix will know to call to appropriate whonix gateway, even if user create custom one, right?

Asked here:
https://phabricator.whonix.org/T534#16396

> Looks fine, but could you put that into qubes-core-admin-addon-whonix package?

Makes sense. Created https://github.com/QubesOS/qubes-core-admin-addon-whonix/pull/2 for it.

> Well, technically it is possible - scan domains when loading them from disk ([domain-load event](https://dev.qubes-os.org/projects/core-admin/en/latest/qubes-vm/qubesvm.html#event-domain-load)). But in practice it's going to be similar thing (looking for `sys-whonix` name), just more obscure.
Might be a good idea to do that anyway, if `whonix-gateway` tag would be needed for anything else.

Created https://github.com/QubesOS/qubes-issues/issues/4094 for it.

PR #221, opened 2018-08-07
$tag:anon-vm $anyvm deny

adrelanos:
Why do we set that using salt?

Simpler and more robust to just set it here.

https://github.com/QubesOS/qubes-mgmt-salt-dom0-virtual-machines/blob/master/qvm/template-whonix-ws.sls#L43


adrelanos:
Somewhat related:
https://github.com/QubesOS/qubes-issues/issues/3994

PR #226, opened 2018-08-30
Make log location more explicit in error message

strugee:
See https://github.com/QubesOS/qubes-issues/issues/4224#issuecomment-414513721.


qubes/app.py
```diff
@@ -1264,7 +1264,7 @@ def on_domain_pre_deleted(self, event, vm):
                             'Cannot remove %s, used by %s.%s',
                             vm, obj, prop.__name__)
                         raise qubes.exc.QubesVMInUseError(vm,
-                            'Domain is in use: {!r}; details in system log'
+                            'Domain is in use: {!r}; see /var/lib/qubes/qubes.log in dom0 for details'
```
marmarek: Should be `/var/log/...`. Also, please break the line to not exceed 80 chars.

strugee:
> Should be `/var/log/...`

Nice catch.

I modified the patch to use a multiline string, but a) I don't know how good the style is (I just massaged what Emacs `C-f 80 RET M-q` did) and b) I couldn't actually figure out a way to test it. I tried just editing the copy of the same file in dom0, but the new message never showed up. I suspect this was due to `.pyc` files, but it makes me nervous as I am not very good at Python and I'm not totally confident this will run right.

marmarek:
''' makes multiline string (with newlines embedded). You can break a single string like this: https://github.com/QubesOS/qubes-core-admin/blob/8362b2470702a21a9ea8e2e831e74a67989bccec/qubes/app.py#L443-L444

strugee:
@marmarek pushed and updated patch. Sorry it took me so long!

PR #227, opened 2018-09-02
Kill all qubes-guid processes on shutdown

marmarek:
Apparently those may stuck on shutdown (miss X server termination?),
delaying the whole system shutdown.

Fixes QubesOS/qubes-issues#1581

### DIFF
--- a/linux/systemd/qubes-core.service
+++ b/linux/systemd/qubes-core.service
@@ -10,10 +10,11 @@ RemainAfterExit=yes
 # Needed to avoid rebooting before all VMs have shut down.
 TimeoutStopSec=180
 ExecStart=/usr/lib/qubes/startup-misc.sh
+ExecStop=-/usr/bin/pkill qubes-guid
 ExecStop=/usr/bin/qvm-shutdown -q --all --wait
 # QubesDB daemons stop after 60s timeout in worst case; speed it up, since no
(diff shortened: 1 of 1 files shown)


constantoverride:
See https://github.com/QubesOS/qubes-issues/issues/1581#issuecomment-420392892

linux/systemd/qubes-core.service
```diff
@@ -10,10 +10,11 @@ RemainAfterExit=yes
 # Needed to avoid rebooting before all VMs have shut down.
 TimeoutStopSec=180
 ExecStart=/usr/lib/qubes/startup-misc.sh
+ExecStop=-/usr/bin/pkill qubes-guid
```
constantoverride: `ExecStop=-/usr/bin/pkill qubes-guid`
Just a reminder that this line should no longer be necessary, because [this PR](https://github.com/QubesOS/qubes-gui-daemon/pull/24) ensures that `qubes-guid` is no longer present at this point in time (there [was](https://github.com/QubesOS/qubes-issues/issues/1581#issuecomment-420847298) only one process but it required SIGKILL to terminate anyway and it was required at an [earlier](https://github.com/QubesOS/qubes-issues/issues/1581#issuecomment-420392892) point, eg. after `graphical.target` stops)

PR #231, opened 2018-09-09
Storage improvements (mostly file-reflink)

rustybird:
Some highlights:

- Make file-reflink a fully non-blocking storage driver, in the sense that all API methods allowed to be coroutines, are coroutines. They still have the same synchronous code, but it's decorated to run in the event loop's thread-based default executor (under a per-volume lock). This approach should have a low risk of introducing bugs. It makes the Qubes UIs responsive during VM startup/shutdown/etc., and allows multiple volumes to be processed in parallel.

- Plug file-reflink into the storage integration tests and add some initial unit tests. FWIW, in my daily use I haven't encountered any bug or data loss, and I'm confident now that it's a better and safer choice than dusty ol' `file` on btrfs. So:

- When creating the /var/lib/qubes pool, use file-reflink if the filesystem supports FICLONE. (Tested using a freshly built qubes-os-r4.0.conf ISO.) ~~This is the last commit.~~ You might not want to cherry-pick it for the release4.0 branch, depending on how conservative R4.0.1 should be.

### DIFF
--- a/qubes/tests/app.py
+++ b/qubes/tests/app.py
@@ -30,6 +30,7 @@ import qubes.events
 
 import qubes.tests
 import qubes.tests.init
+import qubes.tests.storage_reflink
 
 class TestApp(qubes.tests.TestEmitter):
     pass
@@ -262,6 +263,44 @@ class TC_30_VMCollection(qubes.tests.QubesTestCase):
 
 #   def test_201_get_vms_connected_to(self):
 #       pass
+
+
+class TC_80_QubesInitialPools(qubes.tests.QubesTestCase):
+    def setUp(self):
+        super().setUp()
+        self.app = qubes.Qubes('/tmp/qubestest.xml', load=False,
+            offline_mode=True)
+        self.test_dir = '/var/tmp/test-varlibqubes'
+        self.test_patch = mock.patch.dict(
+            qubes.config.defaults['pool_configs']['varlibqubes'],
+            {'dir_path': self.test_dir})
+        self.test_patch.start()
+
+    def tearDown(self):
+        self.test_patch.stop()
+        self.app.close()
+        del self.app
+
+    def get_driver(self, fs_type, accessible):
+        qubes.tests.storage_reflink.mkdir_fs(self.test_dir, fs_type,
+            accessible=accessible, cleanup_via=self.addCleanup)
+        self.app.load_initial_values()
+
+        varlibqubes = self.app.pools['varlibqubes']
+        self.assertEqual(varlibqubes.dir_path, self.test_dir)
+        return varlibqubes.driver
+
+    def test_100_varlibqubes_btrfs_accessible(self):
+        self.assertEqual(self.get_driver('btrfs', True), 'file-reflink')
+
+    def test_101_varlibqubes_btrfs_inaccessible(self):
+        self.assertEqual(self.get_driver('btrfs', False), 'file')
+
+    def test_102_varlibqubes_ext4_accessible(self):
+        self.assertEqual(self.get_driver('ext4', True), 'file')
+
+    def test_103_varlibqubes_ext4_inaccessible(self):
+        self.assertEqual(self.get_driver('ext4', False), 'file')
 
 
 class TC_89_QubesEmpty(qubes.tests.QubesTestCase):


marmarek:
Travis:

> qubes.tests.storage_reflink/TC_00_ReflinkHelpers/test_001_create_and_resize_files_and_update_loopdevs ... losetup: no permission to look at /dev/loop<N>
> ERROR (CalledProcessError: CalledProcessError(255, ('losetup', '-f', '--show', '/var/tmp/test-reflink-helpers/img-real')))

Tests not under integ/ needs to cope with being called as non-root. Simply adding `sudo` to `losetup` calls should be enough.

There is also one pylint nit: unused `name` variable in qubes/app.py:1077 -> change to `for config in ...values()`?

rustybird:
> Tests not under integ/ needs to cope with being called as non-root. Simply adding `sudo` to `losetup` calls should be enough.

I'm on it. And not just integ/, lots of other tests are failing because they can't create /var/lib/qubes, which is nastier because it's [coming from the real qubes/app.py](https://github.com/QubesOS/qubes-core-admin/pull/231/commits/5307d3a969a8ae834f53ff61ae8468dbe8727a47#diff-5e9fd92f515d48b36562578ec8c4c07fR1079), not from test code. Looks like I need to do this differently somehow.

Come to think of it, _why_ isn't the test suite running with the same credentials as a real qubesd?

> There is also one pylint nit: unused `name` variable in qubes/app.py:1077 -> change to `for config in ...values()`?

And I didn't even write that line.

I recently noticed I'm starting to resent _people_ named Travis...

marmarek:
> I'm on it. And not just integ/, lots of other tests are failing because they can't create /var/lib/qubes, which is nastier because it's coming from the real qubes/app.py, not from test code. Looks like I need to do this differently somehow.

I think at least some of those failures are actual bugs. For example just calling load_initial_values shouldn't change any files on disk (like doing `mkdir('/var/lib/qubes')`). This should be done by `pool.create()`.

As last resort method for dealing with this, you can use `unittest.mock.patch`, see `qubes.tests.storage_file.TC_03_FilePool`.

> Come to think of it, why isn't the test suite running with the same credentials as a real qubesd?

1. To spot problems like this.
2. To not accidentally cripple development VM (or at least reduce risk of such thing).

rustybird:
> I think at least some of those failures are actual bugs. For example just calling load_initial_values shouldn't change any files on disk (like doing `mkdir('/var/lib/qubes')`). This should be done by `pool.create()`.

If you mean `pool.setup()`, that one's also called by `load_initial_values()` via `add_pool()`, so `mkdir`s are bound to happen anyway. Although - before [this commit](https://github.com/QubesOS/qubes-core-admin/pull/231/commits/5ec5a764185d276840e6e574580d39bc9414a03c), `setup()` was called on the LVM pool but never at all for /var/lib/qubes and /var/lib/qubes/vm-kernels. Now I wonder if it was actually intentional.

The reason for the additional `mkdir` before `add_pool()` is that AFAIK there's no other way to find out if reflinks are supported: You have to try the ioctl on files in the directory you're interested in.

> As last resort method for dealing with this, you can use `unittest.mock.patch`, see `qubes.tests.storage_file.TC_03_FilePool`.

Direct or indirect calls to `load_initial_values()` are everywhere... 😱

I'm looking for the code that previously (before this PR) created /var/lib/qubes without causing a permission problem during tests, do you know how/where it did that?

>> Come to think of it, why isn't the test suite running with the same credentials as a real qubesd?

> 1. To spot problems like this.
> 2. To not accidentally cripple development VM (or at least reduce risk of such thing).

The second point is solid, but I'd argue that (1) is rather a problem caused by an unrealistic test environment. Maybe both could by avoided by running the tests in a fakeroot or container?

rustybird:
> I'm looking for the code that previously (before this PR) created /var/lib/qubes without causing a permission problem during tests

Oh, it didn't at all. Sorry for the noise.

marmarek:
> Although - before this commit, setup() was called on the LVM pool but never at all for /var/lib/qubes and /var/lib/qubes/vm-kernels. Now I wonder if it was actually intentional.

No, it shouldn't be called on LVM either. If anything, `qubes-create` tool (the one actually writing initial qubes.xml to disk) should iterate over initial pools and call `setup()` on each of them.

> The reason for the additional mkdir before add_pool() is that AFAIK there's no other way to find out if reflinks are supported: You have to try the ioctl on files in the directory you're interested in.

I see. And since driver choice depends on it, it can't be deferred until `setup()`. I think here you can deal with it this way:
 - directory doesn't exist - default to some hardcoded value (i.e. if you want reflink detection, create `/var/lib/qubes` beforehand - it is the case during installation)
 - directory exists, but you get permission denied - default to something
 - to test `load_initial_values()`, `unittest.mock.patch.dict()` relevant parts of `qubes.config.defaults`
 - for other file-reflink tests, create test pool explicitly, pointing at some writable directory (likely `/tmp/something`)

marmarek:
note to self: probably worth adding pool selection test to OpenQA. Example btrfs based installation: https://openqa.qubes-os.org/tests/232

rustybird:
Thanks for the hints, I'll give them a try in another PR.

In this one, I've backed out the last two commits (the `add_pool()` thing and file-reflink /var/lib/pool creation) and added plenty of `sudo` to the unit test commit.

qubes/storage/reflink.py
```diff
@@ -391,27 +416,31 @@ def _create_sparse_file(path, size):
...
+    ''' Resolve img; update the size of loop devices backed by it. '''
+    needle = os.fsencode(os.path.realpath(img)) + b'\n'
+    for sys in glob.iglob('/sys/block/loop[0-9]*/loop/backing_file'):
+        try:
+            with open(sys, 'rb') as sys_io:
+                if sys_io.read() != needle:
+                    continue
+        except FileNotFoundError:
+            continue
+        with open('/dev/' + re.search('loop[0-9]+', sys).group(0)) as dev_io:
```
marmarek: `sys.split('/')[3]` ?

qubes/storage/reflink.py
```diff
@@ -391,27 +416,31 @@ def _create_sparse_file(path, size):
         tmp.truncate(size)
         LOGGER.info('Created sparse file: %s', tmp.name)
 
+def _update_loopdev_sizes(img):
+    ''' Resolve img; update the size of loop devices backed by it. '''
+    needle = os.fsencode(os.path.realpath(img)) + b'\n'
+    for sys in glob.iglob('/sys/block/loop[0-9]*/loop/backing_file'):
```
marmarek: Clobbering `sys` module name (even if not imported here) is asking for troubles.

qubes/storage/reflink.py
```diff
@@ -115,12 +118,37 @@ def included_in(self, app):
...
+        coroutine that runs the original method in the event loop's
+        thread-based default executor, under a per-volume lock.
+    '''
+    @asyncio.coroutine
+    @functools.wraps(method)
+    def wrapper(self, *args, **kwargs):
+        with (yield from self._lock):  # pylint: disable=protected-access
+            return (yield from asyncio.get_event_loop().run_in_executor(
+                None, functools.partial(method, self, *args, **kwargs)))
+    return wrapper
```
marmarek: Generally, using run_in_executor should be considered last resort solution. If proper API exists, better use that.
Looking at the code, I think the only thing benefiting from this is `_copy_file` function. So, why not convert it to a proper coroutine (using [asyncio.create_subprocess_exec](https://docs.python.org/3/library/asyncio-subprocess.html)), then change this decorator to `_locked` or such (or inline lock handling)?

qubes/tests/storage_reflink.py
```diff
@@ -0,0 +1,110 @@
...
+    return int(_sudo("blockdev", "--getsize64", blockdev))
+
+def _update_loopdev_sizes(img):
+    env = [k + "=" + v for k, v in os.environ.items()  # 'sudo -E' alone would
+           if k.startswith("PYTHON")]                  # drop some of these
+    code = ("from qubes.storage import reflink\n"
+            "reflink._update_loopdev_sizes(%r)" % img)
+    _sudo("-E", "env", *env, sys.executable, "-c", code)
+
+def _sudo(*cmd):
```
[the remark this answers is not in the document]
rustybird: That wouldn't show stdout/stderr if the command fails, so it's hard to debug.

I get the impression that `check_output()` is now just a [close to deprecated](https://github.com/python/cpython/blob/1f36bf6077d93cb43fd84bea4a8a625fa772d1fa/Lib/subprocess.py#L17-L41) wrapper.

qubes/storage/reflink.py
rustybird: Definitely.

qubes/storage/reflink.py
rustybird: Eek, yes.

qubes/storage/reflink.py
rustybird: On btrfs, even unlink and overwriting rename can take pretty long when operating on large fragmented snapshots, like easily 5-10 seconds under high load. So these would have to run in executors anyway. But then it just seems cleaner to do it once per method. There's probably not a lot to gain here with fine grained intra-method parallelization?

Another obstacle is that these helpers are also called from methods that are not allowed to be coros, so I guess there'd either have to be both a sync and an async version of each helper, or the sync methods would become responsible for scheduling async helper calls. :\

qubes/tests/storage_reflink.py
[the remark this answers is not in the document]
marmarek: CalledProcessError contains stdout/stderr as attributes. But then, one wrapper would need to be changed to another...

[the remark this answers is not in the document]
rustybird: Exactly - [one of the commits](https://github.com/QubesOS/qubes-core-admin/pull/231/commits/9aeab83b1560c4bf36f1b559cdedbb8bbb209a76) removes a couple of lines by getting rid of such a wrapper around CalledProcessError.

qubes/storage/reflink.py
rustybird: Ew/thanks ;), I'll read up on that.

rustybird:
Pushed the fixup for `_update_loopdev_sizes()` (which I'll squash when it's convenient for you) and the reworked `add_pool()` and reflink detection commits. Tests pass with root and unprivileged now.

Haven't done the `load_initial_values()` unit test yet.

qubes/storage/reflink.py
rustybird: > One concern is handling children processes from multiple threads (SIGCHLD etc) and asyncio (which I believe use signalfd) is potentially fragile.

I'm trying out a small commit (not pushed yet) that calls the FICLONE ioctl natively in `_copy_file()`, and only runs `cp --sparse=always` as a fallback. Too bad that Python doesn't seem to offer a native sparse copy function [yet](https://bugs.python.org/issue10016), if it did there wouldn't have to be any subprocesses in file-reflink at all.

rustybird:
Latest commits:

- Python FICLONE in `_copy_file()`, with a fallback to `cp --sparse=always`
- file-reflink unit/integration tests overhauled to check everything on both btrfs and ext4. [Ugh, Travis has some off-brand `umount` without `--lazy`. Do you know how to switch to the one from util-linux?]
- New unit test for `load_initial_values()` varlibqubes driver selection ...
- ... and the fix for a bug it found (shallow copy of a dict of dicts -> changed to deep copy)

They still need to be autosquashed.

marmarek:
You've probably seen: ancient ubuntu on travis doesn't like `umount --lazy`
Is it necessary? If not - remove. But if yes, I already have code to backport `mount` package from newer ubuntu, can enable it here too.

rustybird:
`-l` works, tests pass, praise the lord. (I thought it was busybox umount or some bullshit, but the old util-linux umount just didn't even mention the option in the usage message.)

marmarek:
Looks good now, please squash those fixups.

rustybird:
done

rustybird:
Sorry, had to repush to reword a commit message:
`factor out _ficlone()` -> `factor out _attempt_ficlone()`

marmarek:
Too late

rustybird:
Oh well.

rustybird:
I wonder if GitHub will automatically mark this PR as merged if I repush the older (non reworded) version, let's see...

PR #232, opened 2018-09-10
add missing /sbin/ to hwclock call (so it will work as cronjob)

xaki23:
straight up bugfix, this has been nonfunctional "forever" (since the rewrite of the sync mechanism)


xaki23:
afaict i pushed a signed_tag_for_f872e2d5 but i can not get github webui to show it in any way ... 
please recheck

PR #235, opened 2018-09-16
Add default_qrexec_timeout to qubes-prefs

rustybird:
When a VM (or its template) does not explicitly set a `qrexec_timeout`, fall back to a global `default_qrexec_timeout` (with default value 60), instead of hardcoding the fallback value to 60.

This makes it easy to set a higher timeout for the whole system, which helps users who habitually launch applications from several (not yet started) VMs at the same time. 60 seconds can be too short for that.

### DIFF
--- a/qubes/app.py
+++ b/qubes/app.py
@@ -718,6 +718,13 @@ class Qubes(qubes.PropertyHolder):
         setter=_setter_pool,
         doc='Default storage pool for kernel volumes')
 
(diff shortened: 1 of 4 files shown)


marmarek:
I [feel your pain](https://github.com/marmarek/openqa-tests-qubesos/blob/542a1957f3524e20eca754cea0b006444ac86108/tests/update.pm#L39).

PR #237, opened 2018-10-22
Avoid 'progress' events flooding

donob4n:
Just a fix for https://github.com/QubesOS/qubes-issues/issues/4406 and https://github.com/QubesOS/qubes-issues/issues/3035

### DIFF
--- a/qubes/backup.py
+++ b/qubes/backup.py
@@ -326,6 +326,8 @@ class Backup(object):
         #: callback for progress reporting. Will be called with one argument
         #: - progress in percents
         self.progress_callback = None
+        self.last_progress = 0
+        self.last_progress_time = time.time()
         #: backup ID, needs to be unique (for a given user),
         #: not necessary unpredictable; automatically generated
         self.backup_id = datetime.datetime.now().strftime(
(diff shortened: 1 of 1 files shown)


qubes/backup.py
```diff
@@ -512,11 +514,15 @@ def _send_progress_update(self):
...
-            progress = (
-                100 * (self._done_vms_bytes + self._current_vm_bytes) /
-                self.total_backup_bytes)
-            # pylint: disable=not-callable
-            self.progress_callback(progress)
+            if time.time() - self.last_progress_time >= 1: # avoid flooding
+                progress = (
+                    100 * (self._done_vms_bytes + self._current_vm_bytes) /
+                    self.total_backup_bytes)
+                if progress - self.last_progress >= 0.5:
```
marmarek: Command line `qvm-backup` displays up to 2 digits. With large backups it makes a difference.
IMO this should be either 0.01, or totally dropped (leaving only time based ratelimit).

donob4n: ok better dropped

donob4n:
Sorry Marek, I did not read your first comment.

PR #240, opened 2018-11-13
qubes/ext/gui: set videoram qdb entries

woju:
This is for adjusting videoram size. See
QubesOS/qubes-gui-agent-linux@d222e16003c072bd6b420feafcbabcfa3988559a


woju:
@andrewdavidwong This is a new feature to change how much video memory is allocated per-vm. Long story short, after hotplugging 4K monitors, notify-monitor-layout fails and mouse clicking is effective only in upper-left portion of screen (a quarter or so). The solution is to reserve more RAM for video, but I don't want to do it for everyone. This is slightly less than 24MiB *additional* memory per running VM, totally wasted if user never attaches 4K display. So this is a configuration option to have more RAM. How it should be documented and where?

The intended usage is to do in dom0, for each template:
```sh
qvm-features $TEMPLATE gui-videoram-min $(($WIDTH * $HEIGHT * 4 / 1024))
qvm-features $TEMPLATE gui-videoram-overhead 0
```
Where `$WIDTH`×`$HEIGHT` is display resolution.

qubes/ext/gui.py
```diff
@@ -33,3 +33,13 @@ def send_gui_mode(vm):
...
             if vm.features.get('gui-seamless', False)
             else 'FULLSCREEN'))
+
+    @qubes.ext.handler('domain-qdb-create')
+    def on_domain_qdb_create(self, vm, event):
+        # pylint: disable=unused-argument,no-self-use
+        for feature in ('gui-videoram-overhead', 'gui-videoram-min'):
+            try:
+                vm.untrusted_qdb.write('/qubes-{}'.format(feature),
+                    vm.features.check_with_template(feature))
```
marmarek: It would be good to have global default for this, not only per-template (you want to use all your VMs with external 4K display, right?). With current API, that could be a feature on `dom0`, but that isn't the most elegant solution. Maybe we need also features on `Qubes()` itself?

marmarek: Yes, at least `check-updates` (default was promoted to `Qubes()`'s property for this reason), but I think that problem will happen from time to time.

woju: This looks like change for 5.0. For example, this throws events `domain-feature-{set,delete}`, which should be changed if features are to be reused on app object. Admin API change should be easy.

marmarek:
> Would it be possible and desirable to enable this automatically when a 4k or higher display is detected?

The problem is when you plug 4K monitor after starting a qube, it is too late for changing its video RAM - this needs to be done at qube startup. If you plug that monitor before starting a qube, this feature isn't needed as initial video ram is calculated based on currently present monitors (+ a margin for one FHD monitor).
Or are you proposing to remember if 4K monitor was ever connected and from that time always reserve video RAM for it (by setting those features)? This means the first time you connect 4K monitor, you need to restart qubes, but later it should work seamlessly.
Then, if you want to release that memory, you'd need to modify that values manually (or remove them to reset to default behavior).

qubes/ext/gui.py
marmarek: I see. So maybe use features on dom0 for now?

woju: There is no "for now". If I do it like that, it will remain. :)

andrewdavidwong:
> The problem is when you plug 4K monitor after starting a qube, it is too late for changing its video RAM - this needs to be done at qube startup. If you plug that monitor before starting a qube, this feature isn't needed as initial video ram is calculated based on currently present monitors (+ a margin for one FHD monitor).

Ah, I see.

> Or are you proposing to remember if 4K monitor was ever connected and from that time always reserve video RAM for it (by setting those features)? This means the first time you connect 4K monitor, you need to restart qubes, but later it should work seamlessly.
Then, if you want to release that memory, you'd need to modify that values manually (or remove them to reset to default behavior).

This behavior might make sense. Usually, someone who plugs in a 4K monitor once will do so again repeatedly. The very first time, there could be an pop-up informing the user that she needs to restart some qubes (just this first time).

andrewdavidwong:
> This will bite people, who once connected to projector/wall display/whatever in conference room, but otherwise don't use 4K display.

Right. This is why I said, "*Usually,* someone who plugs in a 4K monitor once will do so again repeatedly." Ideally, we'd explain the situation to the user, then provide a few options:

```
A high-resolution display has been detected. In order for this display to work properly,
each of your qubes will have to be modified to use additional memory. Each qube must
be restarted afterward in order for this to take effect. Would you like to do this now?

[Yes, just this once. (You'll see this message again the next time you connect a
high-resolution monitor, and each qube will have to be restarted again at that time.
Increased memory usage will last until the next system reboot.)]

[Yes, permanently. (You won't see this message again. Increased memory usage will be
permanent. If you ever wish to undo this, instructions are available in the Qubes
documentation.)]

[Cancel. (This window will close, and no action will be taken.)]

File: qubes/ext/gui.py (hunk 1 above)
Comment (woju) [reply qubes/ext/gui.py]: Here is your global knob.

Please carefully review the commit about features.

**Comment by woju** [comment]
@andrewdavidwong The interaction you proposed seems good. But this pull request is more about mechanism in qubesd, so I'll leave it for now.

Now after the global knob change, the only thing to do for user is:
```sh
qvm-features dom0 gui-videoram-min $(($WIDTH * $HEIGHT * 4 / 1024))
qvm-features dom0 gui-videoram-overhead 0
```

I'm going to write /doc/ page about it.

qubes/features.py
```diff
@@ -0,0 +1,189 @@
...
+
+        if check_adminvm:
+            adminvm = self.subject.app.domains['dom0']
+            if adminvm not in (None, self.subject):
+                try:
+                    return adminvm.features[feature]
+                except KeyError:
+                    pass
+
+        if check_app:
```
marmarek: Since it doesn't work yet (and in fact is broken, not only because app does not have `features` attribute), please remove it or at least comment it out. You may leave a `TODO` comment here.

qubes/features.py
```diff
@@ -0,0 +1,189 @@
...
+            return self[feature]
+        except KeyError:
+            pass
+
+        if attr is not None:
+            next_subject = getattr(self.subject, attr, None)
+            assert not next_subject is self.subject, 'infinite recurse'
+            if next_subject is not None:
+                # pylint: disable=protected-access
+                return next_subject.features._recursive_check(attr,
```
marmarek: I'd prefer `while` loop over real recursion. Otherwise it isn't obvious in what order other conditions are checked (`check_app`, `check_adminvm`).

andrewdavidwong:
> The interaction you proposed seems good. But this pull request is more about mechanism in qubesd, so I'll leave it for now.

Branched to https://github.com/QubesOS/qubes-issues/issues/4509

qubes/features.py
```diff
@@ -0,0 +1,179 @@
...
+        implemented, as app does not have features yet.
+        '''
+        if check_app:
+            raise NotImplementedError('app does not have features yet')
+
+        assert isinstance(self.subject, _vm.BaseVM), (
+            'recursive checks do not work for {}'.format(
+                type(self.subject).__name__))
+
+        if attr is not None:
```
marmarek: Now you skip checking the `self.subject` itself if `attr is None`...

PR #241, opened 2018-11-15
Use maxmem=0 to disable dynamic memory management, instead of cryptic service.meminfo-writer feature

marmarek:
In general, this change makes "maxmem" property strictly bound to dynamic
memory management (aka qmemman). In fact setting maxmem!=(initial) memory makes
only sense in such a case.
If maxmem=0, dynamic memory management is disabled and VM have just 'memory' RAM.
If it is greater than zero, dynamic memory management is enabled and VM can use
up to 'maxmem' RAM. Under the hood, meminfo-writer service is also enabled, as
it's necessary on the VM side for this mechanism; but that's only internal
detail, should not be used as a main knob for this feature.

Having the main knob as a proparty (instead of a "feature"), allows also to
provide meaningful default. Generally, enable qmemman by default for given
domain if it's possible. Disable if not supported, or would crash the domain.

Fixes QubesOS/qubes-issues#4135
Fixes QubesOS/qubes-issues#4480 (kind of - make the thing described there intended behavior)

### DIFF
--- a/templates/libvirt/xen.xml
+++ b/templates/libvirt/xen.xml
@@ -2,11 +2,12 @@
     {% block basic %}
         <name>{{ vm.name }}</name>
         <uuid>{{ vm.uuid }}</uuid>
-        {% if vm.virt_mode == 'hvm' and vm.devices['pci'].persistent() | list %}
+        {% if ((vm.virt_mode == 'hvm' and vm.devices['pci'].persistent() | list)
+            or vm.maxmem == 0) -%}
             <memory unit="MiB">{{ vm.memory }}</memory>
-        {% else %}
+        {% else -%}
             <memory unit="MiB">{{ vm.maxmem }}</memory>
-        {% endif %}
+        {% endif -%}
         <currentMemory unit="MiB">{{ vm.memory }}</currentMemory>
         <vcpu placement="static">{{ vm.vcpus }}</vcpu>
     {% endblock %}
--- a/qubes/ext/services.py
+++ b/qubes/ext/services.py
@@ -39,6 +39,10 @@ class ServicesExtension(qubes.ext.Extension):
             vm.untrusted_qdb.write('/qubes-service/{}'.format(service),
                 str(int(bool(value))))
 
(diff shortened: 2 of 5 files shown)


templates/libvirt/xen.xml
```diff
@@ -2,11 +2,12 @@
     {% block basic %}
         <name>{{ vm.name }}</name>
         <uuid>{{ vm.uuid }}</uuid>
-        {% if vm.virt_mode == 'hvm' and vm.devices['pci'].persistent() | list %}
+        {% if ((vm.virt_mode == 'hvm' and vm.devices['pci'].persistent() | list)
+            or vm.maxmem == 0) -%}
```
woju: Here: this should be `vm.is_memory_balancing_possible()`.

marmarek: I don't think it should be `vm.is_memory_balancing_possible()`. If anything, it should be just `vm.maxmem==0` (and let the core handle the rest).

woju: Hmm, you're right. If balancingis not possible and maxmem != 0, the XML will not be rendered at all.

marmarek:
> I wonder when this logic will fail and falsely prevent startup of some VM.

In all the cases of manual memory=maxmem as a way to avoid the crash. Like we [recommend](https://www.qubes-os.org/doc/windows-vm/#qubes-40---windows-vm-installation) when installing Windows. And also all the cases of Linux HVMs without qubes packages installed (but kernel having balloon driver) and maxmem set manually. This may also apply to some *BSD, but I'm not sure about its default kernel configuration.

So, maybe I should keep the way I modified libvirt xml (you almost convinced me to change it to just `vm.maxmem==0`) and remove the check on domain startup? The PCI devices check doesn't have false positives - Xen refuse to start such (HVM, maxmem > memory and PCI devices) domain, even before looking at what you want to start there.

marmarek:
(pushed `fixup!` commits to ease review changes, will squash them before merge of course)

PR #242, opened 2018-11-22
Drop async future check with python3 version 

fepitre:
QubesOS/qubes-issues#4225

### DIFF
--- a/qubes/tools/qubesd_query.py
+++ b/qubes/tools/qubesd_query.py
@@ -7,12 +7,6 @@ import sys
 
 QUBESD_SOCK = '/var/run/qubesd.sock'
 
-try:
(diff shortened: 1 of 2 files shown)


qubes/tools/qubesd_query.py
```diff
@@ -11,7 +11,7 @@
     asyncio.ensure_future
 except AttributeError:
     # pylint: disable=no-member
-    asyncio.ensure_future = asyncio.async
+    asyncio.ensure_future = asyncio.ensure_future
```
marmarek: This doesn't make sense - if `asyncio.ensure_future` is already there, the assignment isn't needed. The whole `try/except` block is for compatibility with python 3.4. Can be dropped.

PR #244, opened 2018-12-07
Fix typo

strugee:


strugee:
Hah, funny! Thanks for catching that :D

PR #246, opened 2018-12-09
Add admin.repos.* qrexec services

strugee:
This is a prerequisite for QubesOS/qubes-issues#4550.

### DIFF
--- a/qubes-rpc/admin.repos.Disable
+++ b/qubes-rpc/admin.repos.Disable
@@ -1,0 +1,24 @@
+#!/usr/bin/python3
+
+# Empty output indicates success; any input indicates error (probably an exception)
+
+import dnf
+import iniparse
+import sys
+
+base = dnf.Base()
+
+base.read_all_repos()
+
+reponame = sys.stdin.readline()
+repo = base.repos[reponame]
+
+# Loosely based on write_raw_configfile() from DNF source code,
+# because that method was introduced in DNF 2.0 but Qubes dom0 has DNF 1.x.
+with open(repo.repofile) as fp:
+	ini = iniparse.INIConfig(fp)
+
+ini[reponame]['enabled'] = 0
+
+with open(repo.repofile, 'w') as fp:
+	fp.write(str(ini))
--- a/qubes-rpc/admin.repos.Enable
+++ b/qubes-rpc/admin.repos.Enable
@@ -1,0 +1,24 @@
+#!/usr/bin/python3
+
+# Empty output indicates success; any input indicates error (probably an exception)
+
+import dnf
+import iniparse
+import sys
+
+base = dnf.Base()
+
+base.read_all_repos()
+
+reponame = sys.stdin.readline()
+repo = base.repos[reponame]
+
+# Loosely based on write_raw_configfile() from DNF source code,
+# because that method was introduced in DNF 2.0 but Qubes dom0 has DNF 1.x.
+with open(repo.repofile) as fp:
+	ini = iniparse.INIConfig(fp)
+
+ini[reponame]['enabled'] = 1
+
+with open(repo.repofile, 'w') as fp:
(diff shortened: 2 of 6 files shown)


qubes-rpc/admin.repos.Enable
```diff
@@ -0,0 +1,24 @@
...
+repo = base.repos[reponame]
+
+# Loosely based on write_raw_configfile() from DNF source code,
+# because that method was introduced in DNF 2.0 but Qubes dom0 has DNF 1.x.
+with open(repo.repofile) as fp:
+	ini = iniparse.INIConfig(fp)
+
+ini[reponame]['enabled'] = 1
+
+with open(repo.repofile, 'w') as fp:
```
marmarek: This should write to a temporary file and rename it over original only when the whole file is written. Otherwise interrupting the operation (power failure, lack of disk space etc) would result in a broken (empty) repository configuration.

qubes-rpc/admin.repos.Enable
```diff
@@ -0,0 +1,24 @@
...
+
+import dnf
+import iniparse
+import sys
+
+base = dnf.Base()
+
+base.read_all_repos()
+
+reponame = sys.stdin.readline()
```
marmarek: Better use service argument (`sys.argv[1]`), as it will allow policing specific repositories (for example allow to enable/disable current-testing, but not security-testing).

qubes-rpc/admin.repos.Disable
```diff
@@ -0,0 +1,24 @@
...
+
+import dnf
+import iniparse
+import sys
+
+base = dnf.Base()
+
+base.read_all_repos()
+
+reponame = sys.stdin.readline()
```
marmarek: Better use service argument (`sys.argv[1]`), as it will allow policing specific repositories (for example allow to enable/disable current-testing, but not security-testing).

qubes-rpc/admin.repos.Disable
```diff
@@ -0,0 +1,24 @@
...
+repo = base.repos[reponame]
+
+# Loosely based on write_raw_configfile() from DNF source code,
+# because that method was introduced in DNF 2.0 but Qubes dom0 has DNF 1.x.
+with open(repo.repofile) as fp:
+	ini = iniparse.INIConfig(fp)
+
+ini[reponame]['enabled'] = 0
+
+with open(repo.repofile, 'w') as fp:
```
marmarek: This should write to a temporary file and rename it over original only when the whole file is written. Otherwise interrupting the operation (power failure, lack of disk space etc) would result in a broken (empty) repository configuration.

See https://github.com/QubesOS/qubes-mgmt-salt/blob/master/qubessalt/__init__.py#L282-L293 (in that case there is also a locking mechanism, but I think it isn't needed here, as in the worst case, a change will be reverted by a parallel call - something that would happen anyway).

qubes-rpc/admin.repos.List
```diff
@@ -0,0 +1,11 @@
...
+
+import dnf
+
+base = dnf.Base()
+
+base.read_all_repos()
+
+for repo in base.repos.all():
+	l = [repo.id, repo.name, 'enabled' if repo.enabled else 'disabled']
+	print('\0'.join(l))
```
marmarek: I think it is weird to use `\0` for field separation, but `\n` for record separation. But I see why you've done it this way.

strugee: Yeah, I thought about using `:` for field separation, like traditional Unix record fields. The problem with that is that the name of repositories might include `:` and if we add another field to hold the URL in the future, that will _definitely_ include `:`.

I didn't want to use the same character to separate records as I used to separate fields. Hence, `\n`. I'm open to other suggestions, it certainly takes some getting used to.

PR #250, opened 2019-01-22
Change interpreter from python2 to python 3 to address issue 3872

bxatnarf:
`qvm-sync-clock` has hard-coded its interpreter to be `/usr/bin/python2`, however it is not compatible with `python2` due to how it uses the `subprocess` module -- `subprocess.DEVNULL` is only available in `python3`'s version of the `subprocess`.
To address issue https://github.com/QubesOS/qubes-issues/issues/3872

### DIFF
--- a/qvm-tools/qvm-sync-clock
+++ b/qvm-tools/qvm-sync-clock
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/python2
 # -*- encoding: utf8 -*-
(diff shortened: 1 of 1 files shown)


marmarek:
Thanks for the contribution!
Please sign the code, see [here](https://www.qubes-os.org/doc/code-signing/).

marmarek:
Contrary to the description, you're changing python3 to python2.

PR #251, opened 2019-02-09
[R5.0] qubes/config: change qrexec binaries' paths

woju:


marmarek:
Travis says makes file needs updating.

PR #259, opened 2019-05-05
Showterminal

fepitre:


qubes-rpc-policy/admin.vm.Terminal.policy
```diff
@@ -0,0 +1,10 @@
+## Note that policy parsing stops at the first match,
+## so adding anything below "$anyvm $anyvm action" line will have no effect
+
+## Please use a single # to start your custom comments
+
+# WARNING: The qubes.ShowTerminal service is dangerous and allows any
```
marmarek: Update service name here.

qvm-tools/qvm-terminal-dispvm
```diff
@@ -0,0 +1,18 @@
...
+}
+
+if [ $# -lt 1 ]; then
+    print_usage
+    exit 1
+fi
+
+QREXEC_REQUESTED_TARGET="$1"
+
+qvm-check --quiet "$QREXEC_REQUESTED_TARGET" > /dev/null 2>&1 || { echo "Error: no such domain: '$QREXEC_REQUESTED_TARGET'"; exit 1; }
```
marmarek: Check also if the VM is running.

qvm-tools/qvm-terminal-dispvm
```diff
@@ -0,0 +1,18 @@
...
+
+sudo qvm-run -p --localcmd="QREXEC_REQUESTED_TARGET=$QREXEC_REQUESTED_TARGET /etc/qubes-rpc/admin.vm.Terminal" --service --dispvm="$(qubes-prefs management_dispvm)" -- admin.vm.TerminalDispVM
```
marmarek: Retrieve `management_dispvm` separately and check if it isn't empty. 
Use VM's property, not the global one.
`admin.vm.TerminalDispVM` of VM service is bad - `admin.*` are services to control qubes configuration ("Admin API"). This service merely show in terminal whatever you sent on its stdin. Also, the service itself is equally useful in non-DispVM, so including DispVM in its name is misleading. Maybe `qubes.ShowInTerminal`?

qubes-rpc/admin.vm.Console
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+lock="/var/run/qubes/$QREXEC_REQUESTED_TARGET.terminal.lock"
+
```
marmarek: This service (especially when called from mgmt vm) should (also?) check if domain is running. Maybe the check should be moved here instead?

Another thing is, all `admin.*` calls normally go to qubesd and (among other things) emit `mgmt-permission:...` event, to allow addons to plug some more permission checks. I don't think it's a blocker to have it here, but please add a TODO comment here. And maybe create an issue on github, to not forget about it.

qvm-tools/qvm-console-dispvm
```diff
@@ -0,0 +1,22 @@
...
+    exit 1
+fi
+
+QREXEC_REQUESTED_TARGET="$1"
+
+qvm-check --quiet --running "$QREXEC_REQUESTED_TARGET" > /dev/null 2>&1 || { echo "Error: domain '$QREXEC_REQUESTED_TARGET' does not exist or is not running"; exit 1; }
+
+DISPVM="$(qvm-prefs "$QREXEC_REQUESTED_TARGET" management_dispvm)"
+
+[[ "x$DISPVM" == "x" ]] && { echo "Error: cannot determine default DispVM to use"; exit 1; }
```
marmarek:
```suggestion
[[ "x$DISPVM" == "x" ]] && { echo "Error: cannot determine management DispVM to use"; exit 1; }
```

PR #260, opened 2019-05-16
qubesdb: add qubes-mac path entry

fepitre:


qubes/vm/mix/net.py
```diff
@@ -128,6 +128,11 @@ def firewall_conf(self):
     # used in networked appvms or proxyvms (netvm is not None)
     #
 
+    @qubes.stateless_property
+    def visible_mac(self):
+        '''MAC address of this domain as seen by the domain.'''
+        return self.features.check_with_template('net.fake-mac', None) or \
```
marmarek: This doesn't really make sense, as actual eth0 will have mac set based on `self.mac` anyway (see `templates/libvirt/devices/net.xml`). Also, feature for fake-ip (internally doing IP translation) exists because you can't have duplicated IPs when connecting multiple VMs to the same network-providing VM.
On the other hand, it isn't an issue to have duplicated MAC addresses, as long as they are on separate network interfaces. Which means, if one want to set the same MAC for multiple VMs (like all DispVMs), one can do that with `mac` property directly.

fepitre: Ok in this case it's more clear. I was thinking that would allow to override the provided value but it's not the case.

qubes/vm/mix/net.py
```diff
@@ -128,6 +128,10 @@ def firewall_conf(self):
     # used in networked appvms or proxyvms (netvm is not None)
     #
 
+    @qubes.stateless_property
+    def visible_mac(self):
+        '''MAC address of this domain as seen by the domain.'''
+        return self.mac
```
marmarek: `visible_mac` property isn't needed at all, you can use `self.mac` directly in `create_qdb_entries()` below.

fepitre: Ok it was an anticipate code if we would need to do some tweaks here in the future.

marmarek: Better drop it for now. If we'll need to, it's easy enough to add it. But if we add it now, we'll need to keep it in this form. Which means if it turns out we'll something slightly different here, we'll need to introduce yet another property then.

fepitre: Done :)

PR #261, opened 2019-05-26
tiny fix in documentation

maltesl:


marmarek:
BTW for external tools (outside of `qubesd` process), you should use `qubesadmin` module (which will connect to `qubesd`). API is very similar.

PR #262, opened 2019-05-29
Add 'dns' property to Admin API

marmarek:
Make 'dns' property accessible in Admin API. Since it contains two entries, some serialization is needed. Add 'list' property type, with entries separated with '\n', including the last one. Similar to other Admin API methods listing things.

QubesOS/qubes-issues#5050


marmarek:
The 'list' property type introduce only problems. I'm going to convert it to a single 'str' property with two entries separated with a space. For read-only property, it should be good enough.

PR #266, opened 2019-06-15
storage/reflink: fix (re)size logic

rustybird:


rustybird:
Ping @marmarek in case this one's slipped through the cracks and there's still enough time to squeeze it in for R4.0.2.

marmarek:
Oh, indeed I forgot about this. Will review (and possibly merge) today.

rustybird:
Thanks!

PR #272, opened 2019-07-26
Remove CPU usage normalization

airelemental:
Removes CPU normalization in the domains widget.

For example, a VM with 4 cores assigned (such as dom0 on a quad-core system) can go up to 400%.

* Easier to compare CPU usage (no hidden normalization factors).
* Consistent with xentop.
* Fixes an issue with erroneously high CPU percentages in the domain widget when a VM is shutting down.

### DIFF
--- a/qubes/app.py
+++ b/qubes/app.py
@@ -350,8 +350,7 @@ class QubesHost:
             domid = vm['domid']
             current[domid] = {}
             current[domid]['memory_kb'] = vm['mem_kb']
-            current[domid]['cpu_time'] = int(
-                vm['cpu_time'] / max(vm['online_vcpus'], 1))
+            current[domid]['cpu_time'] = vm['cpu_time']
             if domid in previous:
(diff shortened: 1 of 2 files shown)


marmarek:
As explained in https://github.com/QubesOS/qubes-issues/issues/4531#issuecomment-516911113, I'd prefer normalization to stay, at least in default configuration.
CPU stats behaving funny on shutdown fixed in https://github.com/QubesOS/qubes-core-admin/pull/275/commits/39d64eabc82e315ff8c96d318907c78a413dc7e1

ejose19:
@marmarek Is there an option to use the proposed PR? I agree on this being more precise (for example I see 50% on the widget, but without going to the qubes configuration I can't be certain of how many cores it's taking from the host). Also the RAM is already displayed as an absolute value, so I don't see why the CPU shouldn't (but toggeable with an option so nobody is affected). In case we add a toggle for this where should it be placed?

PR #276, opened 2019-08-04
Fix invalid timezone

ejose19:
This fixes an invalid response generated by get_timezone when the time zones are composed by 3 parts, for example:

America/Argentina/Buenos_Aires
America/Indiana/Indianapolis


ejose19:
Ok, I have squashed the commits in a single one.

PR #277, opened 2019-08-06
Devicelist

fepitre:


qubes/api/admin.py
```diff
@@ -1146,6 +1146,19 @@ def vm_remove(self):
...
+    @qubes.api.method('admin.deviceclass.List', no_payload=True,
+        scope='global', read=True)
+    @asyncio.coroutine
+    def deviceclass_list(self):
+        """List all DEVICES classes"""
+        self.enforce(not self.arg)
+        self.enforce(self.dest.name == 'dom0')
+
+        entrypoints = self.fire_event_for_filter(
+            pkg_resources.iter_entry_points('qubes.devices'))
```
marmarek: I think `fire_event_for_filter` should get list of class names, not entrypoint objects. I mean move `ep.name` extraction here.

marmarek: Hmm, on the other hand, we already have enumeration over entry points in vmclass list. Lets keep it this way for consistency.

PR #278, opened 2019-08-08
Add metadata info to LVM AdminAPI

marmarta:
Added usage_details method to Pool class
(returns a dictionary with detailed information
on pool usage) and LVM implementation that returns
metadata info.
Needed for QubesOS/qubes-issues#5053

### DIFF
--- a/qubes/storage/lvm.py
+++ b/qubes/storage/lvm.py
@@ -195,10 +195,28 @@ class ThinPool(qubes.storage.Pool):
         except KeyError:
             return 0
 
+    @property
+    def usage_details(self):
+        result = {}
+        result['data_size'] = self.size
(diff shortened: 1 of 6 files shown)


qubes/storage/lvm.py
```diff
@@ -195,25 +195,49 @@ def usage(self):
...
 
+    @property
+    def usage_details(self):
+        result = {}
+        result['data_size'] = self.size
+        result['data_usage'] = self.usage
+
+        try:
+            metadata_size = qubes.storage.lvm.size_cache[
+                self.volume_group + '/' + self.thin_pool]['size']
```
marmarek: This should be `metadata_size`, similar `metadata_usage` below.

PR #282, opened 2019-09-24
Various fixes

marmarek:
- fix handling libvirt restart (re-register event handlers)
- avoid misleading stack trace in log on DispVM removal
- improve tests

Fixes QubesOS/qubes-issues#5105
Fixes QubesOS/qubes-issues#5303


qubes/app.py
```diff
@@ -102,14 +102,19 @@ def wrapper(*args, **kwargs):
...
         is_dead = not self._conn.isAlive()
         if is_dead:
-            self._conn = libvirt.open(self._conn.getURI())
-            # TODO: re-register event handlers
+            uri = self._conn.getURI()
+            old_conn = self._conn
+            self._conn = libvirt.open(uri)
+            if callable(self._reconnect_cb):
+                self._reconnect_cb(old_conn)
+            old_conn.close()
```
marmarek: This causes regression: domain objects do have references to this connection too and closing it breaks `isAlive()` call, breaking reconnect logic. Keeping old connection alive is wrong, I'll adjust reconnection logic for domain objects.

woju:
>  Move pylint configuration into standard location
>
> Make it easier to run pylint with correct configuration. This was
> especially bad as .pylintrc did existed, but was plain default only.

That there were two pylintrc files was because for some time the version in travis was sufficiently distant from Fedora one (with choice of either very old from LTS or very recent from pip). I wonder what else was left there because of some old versions, since then updated.

marmarek:
In travis we use the one from pip, so quite new. We definitely notice every new warning. And Fedora has only slight older version, so it isn't a big issue.

PR #284, opened 2019-10-01
Add missing source admin.vm.Console call in qvm-console-dispvm

marmarek:
Qrexec service called manually needs both source and destination

Fixes QubesOS/qubes-issues#5362


DemiMarie:
This doesn’t work for me.  After installing socat I get a `No such file or directory` error from socat.

marmarek:
I think it's this race condition: https://github.com/QubesOS/qubes-issues/issues/5156
Also, note you need [this commit](https://github.com/QubesOS/qubes-core-admin/commit/f728801eb2bc832cfa3c1801e3cb5bc6b1097ba3#diff-fed36e93a0509e20f2dc96cbbd85b678) to work with HVM domains (qubes-core-dom0 >= 4.0.44).

PR #287, opened 2019-10-19
GuiVM

fepitre:


qubes/vm/qubesvm.py
```diff
@@ -618,6 +622,12 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.BaseVM):
         type=int,
         doc='Time of last backup of the qube, in seconds since unix epoch')
 
+    default_guivm = qubes.VMProperty('default_guivm',
```
marmarek: This should be only a global property, not here.

qubes/ext/gui.py
```diff
@@ -29,17 +29,45 @@ class GUI(qubes.ext.Extension):
...
+            if vm.guivm is not None:
+                if str(vm.name) != str(vm.guivm.name):
+                    vm.untrusted_qdb.write('/qubes-gui-domain-xid',
+                                           str(vm.guivm.xid))
+        except AttributeError:
+            vm.untrusted_qdb.write('/qubes-gui-domain-xid', '')
+
+        # Add keyboard layout from that of GuiVM
+        try:
+            kbd_layout = vm.guivm.features['kbd_layout']
```
marmarek: Generally we use '-' as word separator in tags, features etc. And also it would be more readable to use full word (`keyboard-layout`).

qubes/ext/gui.py
```diff
@@ -30,16 +30,46 @@ class GUI(qubes.ext.Extension):
...
                 pass
+
+        # Add GuiVM Xen ID for gui-daemon
+        try:
+            if vm.guivm is not None:
+                if str(vm.name) != str(vm.guivm.name):
+                    vm.untrusted_qdb.write('/qubes-gui-domain-xid',
+                                           str(vm.guivm.xid))
+        except AttributeError:
+            vm.untrusted_qdb.write('/qubes-gui-domain-xid', '')
```
marmarek: I think in case of no attribute at all, it shouldn't set anything. Empty could be set if it's `None`.

qubes/tests/vm/qubesvm.py
```diff
@@ -1469,6 +1469,71 @@ def test_621_qdb_vm_with_network(self, mock_qubesdb, mock_urandom,
...
+            '-A INPUT -m conntrack --ctstate '
+            'RELATED,ESTABLISHED -j ACCEPT\n'
+            '-A INPUT -p icmp -j ACCEPT\n'
+            '-A INPUT -i lo -j ACCEPT\n'
+            '-A INPUT -j REJECT --reject-with '
+            'icmp-host-prohibited\n'
+            '-A FORWARD -m conntrack --ctstate '
+            'RELATED,ESTABLISHED -j ACCEPT\n'
+            '-A FORWARD -i vif+ -o vif+ -j DROP\n'
+            'COMMIT\n'.format(datetime.datetime.now().ctime()))
```
marmarek: This is racy and in fact fails some time. Since actual content is not relevant for this test, try using `unittest.mock.ANY` instead.

PR #291, opened 2019-11-06
gui: set guivm windows prefix and default_guivm tag

fepitre:
https://github.com/QubesOS/qubes-issues/issues/4186

### DIFF
--- a/qubes/ext/gui.py
+++ b/qubes/ext/gui.py
@@ -70,3 +70,9 @@ class GUI(qubes.ext.Extension):
             kbd_layout = vm.guivm.features.get('keyboard-layout', None)
             if kbd_layout:
                 vm.untrusted_qdb.write('/keyboard-layout', kbd_layout)
+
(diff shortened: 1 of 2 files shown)


qubes/ext/gui.py
```diff
@@ -70,3 +70,9 @@ def on_domain_qdb_create(self, vm, event):
             kbd_layout = vm.guivm.features.get('keyboard-layout', None)
             if kbd_layout:
                 vm.untrusted_qdb.write('/keyboard-layout', kbd_layout)
+
+        # Set GuiVM prefix
+        guivm_windows_prefix = vm.features.get('guivm-windows-prefix', 'GuiVM')
+        if vm.features.get('service.guivm-gui-agent', None) == '1':
```
marmarek: It doesn't need to be `1` (in practice it is, but lets be consistent). Technically it can be any value and should be evaluated to true/false according to python rules. In short: drop `== '1'`.

PR #297, opened 2019-11-22
Dom0 fc31

fepitre:


qubes/dochelpers.py
```diff
@@ -41,12 +41,17 @@
...
+from sphinx.util import logging
 
 import qubes.tools
 
 SUBCOMMANDS_TITLE = 'COMMANDS'
 OPTIONS_TITLE = 'OPTIONS'
 
+try:
+    log = logging.getLogger(__name__)
+except:
```
marmarek: pylint don't like bare except and I don't like it either. Please use `except AttributeError`, same as in core-admin-client.

qubes/dochelpers.py
```diff
@@ -380,15 +396,22 @@ def visit_section(self, node):
...
     if os.path.basename(dirname) != 'manpages':
         return
 
-    app.info('Checking arguments for {!r}'.format(command))
+    msg = 'Checking arguments for {!r}'.format(command)
+    if log:
+        log.info(msg)
+    else:
+        # Handle legacy
+        log.info(msg)
```
marmarek: I think this should be `app.info`.

qubes/qmemman/__init__.py
```diff
@@ -57,6 +57,10 @@ def __init__(self):
         self.log.debug('SystemState()')
 
         self.domdict = {}
+        self.xc = None
+        self.xs = None
+
+    def init(self):
```
marmarek: I don't think I'm against this change, but can you provide the reason (preferably in the commit message)?

PR #303, opened 2020-01-09
Maintain a list of connected machine IPs in qubesdb

pwmarcz:
Necessary for anti-spoofing, see QubesOS/qubes-issues#5540.


marmarek:
You need the same for IPv6 when enabled. Also, you need to update tests.

pwmarcz:
Updated for IPv6, and tests are passing now.

PR #305, opened 2020-01-14
Test file copy hang when target machine qrexec is disabled

pwmarcz:
See QubesOS/qubes-issues#5347.

Fix in QubesOS/qubes-core-qrexec#15.

### DIFF
--- a/qubes/tests/integ/vm_qrexec_gui.py
+++ b/qubes/tests/integ/vm_qrexec_gui.py
@@ -664,6 +664,32 @@ class TC_00_AppVMMixin(object):
                 'test -d /home/user/QubesIncoming/{}'.format(
                     self.testvm1.name)))
 
+    def test_115_qrexec_filecopy_no_agent(self):
+        # The operation should not hang when qrexec-agent is down on target
+        # machine, see QubesOS/qubes-issues#5347.
+
+        self.loop.run_until_complete(asyncio.wait([
+            self.testvm1.start(),
+            self.testvm2.start()]))
+
+        with self.qrexec_policy('qubes.Filecopy', self.testvm1, self.testvm2):
+            try:
+                self.loop.run_until_complete(
+                    self.testvm2.run_for_stdio(
+                        'systemctl stop qubes-qrexec-agent.service', user='root'))
+            except subprocess.CalledProcessError:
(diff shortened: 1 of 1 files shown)


qubes/tests/integ/vm_qrexec_gui.py
```diff
@@ -664,6 +664,34 @@ def test_110_qrexec_filecopy_deny(self):
...
+        self.loop.run_until_complete(asyncio.wait([
+            self.testvm1.start(),
+            self.testvm2.start()]))
+
+        with self.qrexec_policy('qubes.Filecopy', self.testvm1, self.testvm2):
+            # Make sure qrexec-agent is running before killing it
+            self.loop.run_until_complete(asyncio.wait([
+                self.wait_for_session(self.testvm2)]))
+            self.loop.run_until_complete(
+                self.testvm2.run_for_stdio('sudo pkill qrexec-agent'))
```
marmarek: Please use `user='root'` keyword argument instead of sudo, it may be not configured (for example in minimal template). It may be also better to use `systemctl stop ...` instead of killing (if we ever add automatic restart there),

qubes/tests/integ/vm_qrexec_gui.py
```diff
@@ -664,6 +664,34 @@ def test_110_qrexec_filecopy_deny(self):
...
+        # machine, see QubesOS/qubes-issues#5347.
+
+        self.loop.run_until_complete(asyncio.wait([
+            self.testvm1.start(),
+            self.testvm2.start()]))
+
+        with self.qrexec_policy('qubes.Filecopy', self.testvm1, self.testvm2):
+            # Make sure qrexec-agent is running before killing it
+            self.loop.run_until_complete(asyncio.wait([
+                self.wait_for_session(self.testvm2)]))
```
marmarek: Didn't you wanted to add both testvm1 and testvm2 here? Anyway, qrexec is started as part of `vm.start()` already, wait_for_session is needed only for gui-agent (which I don't think you use here).

pwmarcz: I had some weird intermittent failures on my machine, might be a local setup issue. I will remove that.

PR #308, opened 2020-01-16
Prevent removing VM if it provides devices in persistent mode

pwmarcz:
Fixes QubesOS/qubes-issues#5136.

### DIFF
--- a/qubes/api/admin.py
+++ b/qubes/api/admin.py
@@ -1201,6 +1201,14 @@ class QubesAdminAPI(qubes.api.AbstractQubesAPI):
             if not self.dest.is_halted():
                 raise qubes.exc.QubesVMNotHaltedError(self.dest)
 
+            assignments = self.dest.get_provided_assignments()
+            if assignments:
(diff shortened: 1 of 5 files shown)


qubes/api/admin.py
```diff
@@ -1201,6 +1201,14 @@ def vm_remove(self):
...
                 raise qubes.exc.QubesVMNotHaltedError(self.dest)
 
+            assignments = self.dest.get_provided_assignments()
+            if assignments:
+                desc = ', '.join(
+                    assignment.ident for assignment in assignments)
+                raise qubes.exc.QubesVMInUseError(self.dest,
+                    'VM has devices attached persistently to other VMs: ' +
+                    desc)
+
```
marmarek: This would better fit generic VM removal checks, not only through Admin API. See `qubes/app.py:Qubes.on_domain_pre_deleted`. This would also give you a little more context for https://github.com/QubesOS/qubes-core-admin-client/pull/126

PR #309, opened 2020-01-16
Add admin.vm.volume.ImportWithSize

pwmarcz:


qubes/api/internal.py
```diff
@@ -56,6 +56,52 @@ def getsysteminfo(self):
...
+            self.arg, requested_size)
+        self.enforce(' ' not in path)
+        if requested_size is None:
+            size = self.dest.volumes[self.arg].size
+        else:
+            size = requested_size
+
+        # when we know the action is allowed, inform extensions that it will
+        # be performed
+        self.dest.fire_event('domain-volume-import-begin', volume=self.arg)
```
marmarek: It's worth adding a size arg here too (since it isn't always the same as volume.size anymore).

qubes/api/admin.py
```diff
@@ -486,36 +486,6 @@ def vm_volume_resize(self, untrusted_payload):
...
-        Note that this function only returns a path to where data should be
-        written, actual importing is done by a script in /etc/qubes-rpc
-        When the script finish importing, it will trigger
-        internal.vm.volume.ImportEnd (with either b'ok' or b'fail' as a
-        payload) and response from that call will be actually send to the
-        caller.
-        """
-        self.enforce(self.arg in self.dest.volumes.keys())
-
-        self.fire_event_for_permission()
```
marmarek: This needs to be added in the new version too, so core-admin extension could deny import based on other factors. It may be need to open-code this function, to hide `internal.` vs `admin.` difference.

pwmarcz: You mean, fire an event `admin-permission:admin.vm.volume.Import` or `admin-permission:admin.vm.volume.ImportWithSize`, depending on how we are called?

marmarek: Yes.

marmarek:
Please also add this new method to https://www.qubes-os.org/doc/admin-api/ (is it really the only documentation page without the link to the source file?! it's [here](https://github.com/QubesOS/qubes-doc/blob/master/developer/services/admin-api.md)).

qubes/tests/rpc_import.py
```diff
@@ -0,0 +1,180 @@
...
+        ])
+        self.assertEqual(payload, expected_payload)
+
+    # pylint:disable=invalid-name
+    def assertFileData(self, path, expected_data):
+        with open(path, 'rb') as data_f:
+            data = data_f.read()
+        self.assertEquals(data, expected_data)
+
+    def setup_import(self, data, size):
```
marmarek: `data` unused?

pwmarcz:
I removed `data`, and rebased the whole change into 2 commits.

PR #311, opened 2020-01-21
Fix XML loading volume config

pwmarcz:
Some values should be treated as integers, not strings.

Fixes QubesOS/qubes-issues#5219.

### DIFF
--- a/qubes/storage/lvm.py
+++ b/qubes/storage/lvm.py
@@ -314,9 +314,9 @@ class ThinVolume(qubes.storage.Volume):
     '''  # pylint: disable=too-few-public-methods
 
 
-    def __init__(self, volume_group, size=0, **kwargs):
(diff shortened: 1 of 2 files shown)


rustybird:
Maybe just revert 8974b2c instead? Initialization of `_size` is handled - with int conversion - by the Volume base class since b863152.

(`revisions_to_keep` initialization looks like it's already okay, because lvm_thin doesn't override it.)

marmarek:
> Maybe just revert [8974b2c](https://github.com/QubesOS/qubes-core-admin/commit/8974b2cbc09715d81aebd57a402eb02b5ecd7f85) instead?

Sounds like a good idea. Better than duplicating list of (int) attributes elsewhere.
But the test should stay :)

pwmarcz:
Okay, done.

Test stays, but it checks for strings now, of course.

PR #312, opened 2020-01-21
Added handling for 'force' option in vm shutdown

marmarta:
To be more precise: just allowed using it, as the option was
actually implemented previously.

references QubesOS/qubes-issues#5591


qubes/api/admin.py
```diff
@@ -899,9 +899,9 @@ def vm_start(self):
         scope='local', execute=True)
     @asyncio.coroutine
     def vm_shutdown(self):
-        self.enforce(not self.arg)
         self.fire_event_for_permission()
```
marmarek: `fire_event_for_permission()` should also get `force` parameter, so extensions can make more informed decision.

PR #318, opened 2020-01-29
Pass network parameters to linux-stubdom

pwmarcz:
See QubesOS/qubes-issues#5022.

### DIFF
--- a/qubes/tests/vm/qubesvm.py
+++ b/qubes/tests/vm/qubesvm.py
@@ -1439,7 +1439,7 @@ class TC_90_QubesVM(QubesVMTestsMixin, qubes.tests.QubesTestCase):
                 <script path="vif-route-qubes" />
             </interface>
             <!-- server_ip is the address of stubdomain. It hosts it's own DNS server. -->
-            <emulator type="stubdom-linux" />
+            <emulator type="stubdom-linux" cmdline="-qubes-net:client_ip=10.137.0.1,dns_0=10.139.1.1,dns_1=10.139.1.2,gw=10.137.0.2,netmask=255.255.255.255" />
             <input type="tablet" bus="usb"/>
             <video>
                 <model type="vga"/>
--- a/templates/libvirt/xen.xml
+++ b/templates/libvirt/xen.xml
@@ -166,13 +166,20 @@
                     {% else %}
                         type="stubdom"
                     {% endif %}
-                    {% if vm.netvm and not
-                        vm.features.check_with_template('linux-stubdom', True) %}
+                    {% if vm.netvm %}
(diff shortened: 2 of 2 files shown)


qubes/tests/vm/qubesvm.py
```diff
@@ -1439,7 +1439,7 @@ def test_610_libvirt_xml_network(self):
                 <script path="vif-route-qubes" />
             </interface>
             <!-- server_ip is the address of stubdomain. It hosts it's own DNS server. -->
-            <emulator type="stubdom-linux" />
+            <emulator type="stubdom-linux" cmdline="-qubes-net:client_ip=10.137.0.1,dns_0=10.139.1.1,dns_1=10.139.1.2,gw=10.137.0.2,netmask=255.255.255.255" />
```
jandryuk: It may not matter since this is a test, but netmask changed from 255.255.255.0 in [linux-stubdom](https://github.com/pwmarcz/qubes-vmm-xen-stubdom-linux/commit/53648843f698cf45f1e5d556d46976dfd025a31b#diff-9b7847ba61b5a8a5e3ecc2a7952b1c3fL36) to 255.255.255.255 here.

marmarek: Ah, indeed, thanks! Generally this is what we set in Linux VMs, together with explicit route to a gateway, but AFAIR Windows is not happy about it.
We have proxy_arp enabled, so exact netmask doesn't matter anyway (connected network-providing vm will respond with its own MAC for ARP requests for routed addresses too),

BTW Where (if at all) do you run DHCP server in OpenXT?

jandryuk: DHCP runs inside the Network Device VM (ndvm).  ndvm has 5 bridges: brbridged, brinternal0, brwireless, brshared & brany.  brany, brshared & brwireless provide DHCP & NAT for guests.  brshared is for wired, brwireless is wireless, and brany will use either.  brbridged just bridges guests onto the wired ethernet, and brinternal is a disconnected bridge.

Why do you run DHCP out of stubdom instead of sys-firewall or sys-net?  Do you also run DHCP for the Xen vifs or push their IPs in through QubesRPC?

marmarek: > Why do you run DHCP out of stubdom instead of sys-firewall or sys-net? 

To not extend attack surface of them, since they see traffic of other VMs and also apply policies (firewall).

> Do you also run DHCP for the Xen vifs or push their IPs in through QubesRPC?

Not qubes-rpc specifically, but a qubes-specific startup script, which gets info from qubesdb (xenstore-like thing).

jandryuk: Thanks!

PR #319, opened 2020-02-10
gui: handle legacy keymap setting /qubes-keyboard

fepitre:
https://github.com/QubesOS/qubes-issues/issues/5662
https://github.com/QubesOS/qubes-issues/issues/833

### DIFF
--- a/qubes/ext/gui.py
+++ b/qubes/ext/gui.py
@@ -21,6 +21,8 @@
 # License along with this library; if not, see <https://www.gnu.org/licenses/>.
 #
 
+import subprocess
+
 import qubes.config
 import qubes.ext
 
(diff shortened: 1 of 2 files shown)


qubes/ext/gui.py
```diff
@@ -79,6 +81,17 @@ def on_domain_qdb_create(self, vm, event):
             if kbd_layout:
                 vm.untrusted_qdb.write('/keyboard-layout', kbd_layout)
 
+            # Legacy value for setting keyboard layout
+            try:
+                xkb_keymap = subprocess.run(['/usr/bin/setxkbmap', '-print'],
+                                            stdout=subprocess.PIPE,
+                                            check=False)
```
marmarek: This won't work this easily. This is run in qubesd service, which don't have access to X session. Some workaround could be setting `DISPLAY=:0` variable (assuming X server is running at `:0`), and switching to the right user (first in `qubes` group). Similar problem was [here](https://github.com/QubesOS/qubes-desktop-linux-common/blob/master/qubesappmenusext/__init__.py#L34-L55). Generally this is not ideal, but as a just migration solution should be fine.

marmarek: Maybe, a good option would be to look at `setxkbmap -print` format and build a value based on this and `/keyboard-layout`?

fepitre: I don't know if we can easily predict output. Notably 'qwerty/azerty' or any other options. That's the reason of such tool :)

fepitre:
https://github.com/QubesOS/qubes-issues/issues/5662

PR #326, opened 2020-03-09
api/admin: (ext/admin) limit listing VMs based on qrexec policy

marmarek:
Various Admin API calls, when directed at dom0, retrieve global system view
instead of a specific VM. This applies to admin.vm.List (called at dom0
retrieve full VM list) and admin.Events (called at dom0 listen for events
of all the VMs). This makes it tricky to configure a management VM with
access to limited set of VMs only, because many tools require ability to
list VMs, and that would return full list.

Fix this issue by adding a filter to admin.vm.List and admin.Events calls
(using event handlers in AdminExtension) that filters the output using
qrexec policy. This version evaluates policy for each VM or event
(but loads only once). If the performance will be an issue, it can be 
optimized later.

Fixes QubesOS/qubes-issues#5509

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,10 @@ install:
   - sudo apt-get -y install python3-gi gir1.2-gtk-3.0
   - pip install --quiet -r ci/requirements.txt
   - git clone https://github.com/"${TRAVIS_REPO_SLUG%%/*}"/qubes-builder ~/qubes-builder
+  - git clone https://github.com/"${TRAVIS_REPO_SLUG%%/*}"/qubes-core-qrexec ~/qubes-core-qrexec
 script:
-  - PYTHONPATH=test-packages pylint qubes
-  - ./run-tests
+  - PYTHONPATH=test-packages:~/qubes-core-qrexec pylint qubes
+  - PYTHONPATH=test-packages:~/qubes-core-qrexec ./run-tests
   - ~/qubes-builder/scripts/travis-build
 env:
  - DIST_DOM0=fc31 USE_QUBES_REPO_VERSION=4.1 USE_QUBES_REPO_TESTING=1
--- a/ci/requirements.txt
+++ b/ci/requirements.txt
@@ -8,3 +8,4 @@ lxml
 pylint
 sphinx
 PyYAML
+pyinotify
--- a/qubes/api/internal.py
+++ b/qubes/api/internal.py
@@ -30,6 +30,23 @@ import qubes.vm.adminvm
 import qubes.vm.dispvm
 
 
+def get_system_info(app):
+    system_info = {'domains': {
+        domain.name: {
+            'tags': list(domain.tags),
+            'type': domain.__class__.__name__,
+            'template_for_dispvms':
+                getattr(domain, 'template_for_dispvms', False),
+            'default_dispvm': (domain.default_dispvm.name if
+                getattr(domain, 'default_dispvm', None) else None),
+            'icon': str(domain.label.icon),
+            'guivm': (domain.guivm.name if getattr(domain, 'guivm', None)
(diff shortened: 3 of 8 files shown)


pwmarcz:
Testing with my `sys-gui` setup.

* In dom0, the call still returns everything.

* In sys-gui, with the policy of `admin.vm.List * sys-gui @adminvm allow` the call returns just dom0.

* After adding a second line: `admin.vm.List * sys-gui @tag:guivm-sys-gui allow target=dom0` the call returns dom0 and the VMs using `sys-gui`.

* `admin.vm.Events` is broken, I think. I ran it manually, and got the following on a VM shutdown:

```
  File "/usr/lib/python3.7/site-packages/qubes/api/__init__.py", line 275, in respond
    untrusted_payload=untrusted_payload)
  File "/usr/lib/python3.7/site-packages/qubes/api/admin.py", line 874, in vm_shutdown
    yield from self.dest.shutdown(force=force)
  File "/usr/lib/python3.7/site-packages/qubes/vm/qubesvm.py", line 1289, in shutdown
    reason=str(ex))
  File "/usr/lib/python3.7/site-packages/qubes/events.py", line 231, in fire_event_async
    kwargs, pre_event=pre_event)
  File "/usr/lib/python3.7/site-packages/qubes/events.py", line 166, in _fire_event
  File "/usr/lib/python3.7/site-packages/qubes/api/admin.py", line 67, in vm_handler
    self.filters)):
TypeError: filter_events() missing 2 required positional arguments: 'event' and 'kwargs'
```

pwmarcz:
Works with the following fix:

```diff
diff --git a/qubes/ext/admin.py b/qubes/ext/admin.py
index 1209f976..2ed22b99 100644
--- a/qubes/ext/admin.py
+++ b/qubes/ext/admin.py
@@ -100,7 +100,8 @@ class AdminExtension(qubes.ext.Extension):
             # dom0 can always list everything
             return None
 
-        def filter_events(subject, event, kwargs):
+        def filter_events(event):
+            subject, event, kwargs = event
             try:
                 dest = subject.name
             except AttributeError:
```

Again, when allowing the call to dom0 I don't get any notifications except "connection established", adding the line with `@tag:guivm-sys-gui` adds notifications about the related VMs and no others.

pwmarcz:
For the GuiVM to function, I also had to enable listing itself, `qvm-start-gui` crashes otherwise (by the way, XFCE displays a notification about a crash).

GuiVM works now, with following policy added:

```
admin.vm.List                   *	sys-gui         @adminvm                allow
admin.vm.List                   *	sys-gui         @tag:guivm-sys-gui	allow target=dom0
admin.vm.List                   *	sys-gui         sys-gui                 allow target=dom0
admin.Events                    *	sys-gui         dom0                    allow
admin.Events                    *	sys-gui         @tag:guivm-sys-gui	allow target=dom0
admin.Events                    *	sys-gui         sys-gui                 allow target=dom0
admin.vm.property.Get           *	sys-gui         @tag:guivm-sys-gui	allow target=dom0
admin.vm.property.Get           *	sys-gui         sys-gui                 allow target=dom0
admin.label.Get                 *	sys-gui         dom0                    allow
admin.label.Index               *	sys-gui         dom0                    allow
```

marmarek:
> * In sys-gui, with the policy of `admin.vm.List * sys-gui @adminvm allow` the call returns just dom0.
> 
> * After adding a second line: `admin.vm.List * sys-gui @tag:guivm-sys-gui allow target=dom0` the call returns dom0 and the VMs using `sys-gui`.

This is exactly how it should work.

Can you update the policy in https://github.com/QubesOS/qubes-mgmt-salt-dom0-virtual-machines/blob/master/qvm/sys-gui.sls ?

PR #327, opened 2020-03-15
gui: set xkb_keymap from features

fepitre:


qubes/ext/gui.py
```diff
@@ -96,14 +96,9 @@ def on_domain_qdb_create(self, vm, event):
...
-                # Legacy value for setting keyboard layout
-                xkb_keymap = \
-                    'xkb_keymap {\x0a\x09xkb_keycodes  { include ' \
-                    '"evdev"\x09};\x0a\x09xkb_types     { include ' \
-                    '"complete"\x09};\x0a\x09xkb_compat    { include ' \
-                    '"complete"\x09};\x0a\x09xkb_symbols   { include ' \
-                    '"pc+%s+inet(evdev)"\x09};\x0a\x09xkb_geometry  ' \
-                    '{ include "pc(pc105)"\x09};\x0a};' % kbd_layout
+            # Legacy value for setting keyboard layout
+            xkb_keymap = vm.guivm.features.get('qubes-keyboard', None)
```
marmarek: As noted in https://github.com/QubesOS/qubes-core-admin-client/pull/118, I don't like raw `setxkbmap -print` value in features.
In fact, it would be good to validate 'keyboard-layout' feature value when it's being set (handling `domain-feature-pre-set:keyboard-layout` event and raising `QubesValueError` for improperly formatted values). See #328

PR #330, opened 2020-03-17
Fix layout

fepitre:


qubes/ext/gui.py
```diff
@@ -128,3 +121,22 @@ def on_domain_start(self, vm, event, **kwargs):
         for attached_vm in attached_vms:
             attached_vm.untrusted_qdb.write('/qubes-gui-domain-xid',
                                             str(vm.xid))
+
+    @qubes.ext.handler('domain-feature-pre-set:feature')
```
marmarek:
```suggestion
    @qubes.ext.handler('domain-feature-pre-set:keyboard-layout')
```
(and then, you don't need to check if `features` is really `keyboard-layout`)
This is probably also why the test fails now.

fepitre: 'value' in this case is redondent?

PR #331, opened 2020-04-01
Fix multiple qmemman issues

marmarek:
First the main bug: when meminfo xenstore watch fires, in some cases
(just after starting some domain) XS_Watcher refreshes internal list of
domains before processing the event. This is done specifically to
include new domain in there. But the opposite could happen too - the
domain could be destroyed. In this case refres_meminfo() function raises
an exception, which isn't handled and interrupts the whole xenstore
watch loop. This issue is likely to be triggered by killing the domain,
as this way it could disappear shortly after writing updated meminfo
entry. In case of proper shutdown, meminfo-writer is stopped earlier and
do not write updates just before domain destroy.
Fix this by checking if the requested domain is still there just after
refreshing the list.

Then, catch exceptions in xenstore watch handling functions, to not
interrupt xenstore watch loop. If it gets interrupted, qmemman basically
stops memory balancing.

And finally, clear force_refresh_domain_list flag after refreshing the
domain list. That missing line caused domain refresh at every meminfo
change, making it use some more CPU time.

Thanks @conorsch for capturing valuable logs.

Fixes QubesOS/qubes-issues#4890

### DIFF
--- a/qubes/tools/qmemmand.py
+++ b/qubes/tools/qmemmand.py
@@ -123,13 +123,18 @@ class XS_Watcher(object):
                 self.handle.unwatch(get_domain_meminfo_key(i), self.watch_token_dict[i])
                 self.watch_token_dict.pop(i)
                 system_state.del_domain(i)
+        except:
+            self.log.exception('Updating domain list failed')
         finally:
             if got_lock:
                 global_lock.release()
                 self.log.debug('global_lock released')
 
         if not refresh_only:
-            system_state.do_balance()
+            try:
+                system_state.do_balance()
+            except:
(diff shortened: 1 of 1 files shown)


conorsch:
Looks great! Have patched locally, will keep an eye on behavior and report back.

conorsch:
For what it's worth: I've been running this patch locally for nearly a week, and I've yet to observe a single failure of the `qubes-memman` service. Have been running `watch xl list` to observe constant rebalancing. Hardly a definite confirmation of the fix, but it's already a dramatic improvement over the behavior I was seeing before the patch!

So, :+1: on merge from me!

PR #332, opened 2020-04-01
tests: ensure proper cleanup of auxiliary process

marmarek:
Various qrexec tests create auxiliary process (service_proc) as a local
variable. In case of test failure, process cleanup isn't called and may
lead to FD leaks and breaking subsequent tests.

Fix this by always saving such process instance in self.service_proc and
cleaning it up in self.tearDown() (this code is already there).


pwmarcz:
If I undestand correctly, these two tests still do not wait for the result of run_service - there is `p.stdout.read()` but no `p.wait()`.

marmarek:
There is `self.service_proc.communicate()` in `tearDown`.

marmarek:
Just after `self.service_proc.terminate()`

marmarek:
hmm, wait, run_service,

marmarek:
Yes, waiting for run_service process helped.

PR #333, opened 2020-04-07
tests/file: run_until_complete(); coro_maybe() for single arg

rustybird:
I wonder if there is any way to get a warning for objects from an `@asyncio.coroutine` function that are never yielded from? Like the `RuntimeWarning: coroutine 'foo' was never awaited` for objects from `async def foo()`.

### DIFF
--- a/qubes/tests/storage_file.py
+++ b/qubes/tests/storage_file.py
@@ -22,7 +22,6 @@
 import os
 import shutil
 
-import asyncio
 import unittest.mock
 
 import subprocess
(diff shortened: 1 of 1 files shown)


qubes/tests/storage_file.py
```diff
@@ -460,20 +461,22 @@ def test_007_resize_running(self):
         }
         vm = qubes.tests.storage.TestVM(self)
         volume = self.app.get_pool(self.POOL_NAME).init_volume(vm, config)
-        qubes.utils.void_coros_maybe([volume.create()])
-        qubes.utils.void_coros_maybe([volume.start()])
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(qubes.utils.coro_maybe(volume.create()))
```
marmarek, line 465: `self.loop`?

marmarek:
> I wonder if there is any way to get a warning for objects from an `@asyncio.coroutine` function that are never yielded from?

I'm afraid there isn't. On the other hand, in R4.1-only code we can use `async def` (but better not in a code that should be backported to R4.0 - there were issues with `async def` in R4.0).

qubes/tests/storage_file.py
rustybird: Done.

PR #335, opened 2020-04-11
spec: update R for PyQt4

fepitre:
QubesOS/qubes-issues#5763


marmarek:
In fact, we don't use PyQt4 here anymore.

PR #337, opened 2020-04-14
AdminVM support for Qubes network server (release 4.1).

Rudd-O:
These modifications create a new feature `routing-method` which defaults to the normal Qubes OS behavior of masquerading outgoing traffic from AppVMs.  When the `routing-method` feature is set on a VM, its value is written to a `/qubes-routing-method/<IP>` Qubes DB entry within its NetVM.

NetVMs can thus use that information to switch from masquerading to normal IP forwarding for VMs designated with `routing-method=forward` by the administrator.

Other than creating the necessary `/qubes-routing-method/<IP>` hierarchy in Qubes DB, this code does nothing else.

This feature does not yet support chains of NetVMs -- only the NetVM directly attached to the AppVM is affected.

To see the companion agent that uses this new Qubes DB information, please refer to branch `r4.0` of https://github.com/Rudd-O/qubes-network-server/tree/r4.0 .  The agent in that branch supports reading from the `/qubes-routing-method` tree to configure the NetVM appropriately.

I am working, in parallel, on a Qubes OS >= 4.1.compatible implementation.

### DIFF
--- a/qubes/tests/vm/qubesvm.py
+++ b/qubes/tests/vm/qubesvm.py
@@ -1794,6 +1794,7 @@ class TC_90_QubesVM(QubesVMTestsMixin, qubes.tests.QubesTestCase):
             expected['/qubes-firewall/10.137.0.3/0000'] = 'action=accept'
             expected['/qubes-firewall/10.137.0.3/policy'] = 'drop'
             expected['/connected-ips'] = '10.137.0.3'
+            expected['/qubes-routing-method/10.137.0.3'] = 'masquerade'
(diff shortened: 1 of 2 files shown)


Rudd-O:
About code signing: is someone set up for signing interested in taking this code up as theirs?  I have no interest in personally appearing in the commit logs.

Rudd-O:
I'm closing this as the PR has now been folded into https://github.com/Rudd-O/qubes-network-server/tree/r4.0.

PR #338, opened 2020-04-17
Pools

fepitre:


qubes/storage/__init__.py
```diff
@@ -921,15 +926,20 @@ def _init(cls, dir_path):
                 if os.getuid():
                     sudo = ['sudo']
                 root_table = subprocess.check_output(sudo + ["dmsetup",
-                    "-j", str(fs_major), "-m", str(fs_minor),
-                    "table"], stderr=subprocess.DEVNULL)
+                                                             "-j",
+                                                             str(fs_major),
+                                                             "-m",
+                                                             str(fs_minor),
+                                                             "table"],
```
marmarek, line 933: Can you format it better? Like, break the line after `check_output(`

qubes/storage/lvm.py
```diff
@@ -33,19 +33,258 @@
...
+    try:
+        for key, vol in size_cache.items():
+            if vol['attr'][0] == 't':
+                # e.g. 'qubes_dom0/pool00'
+                parsed_key = key.split('/')
+                if len(parsed_key) == 2:
+                    volume_group = parsed_key[0]
+                    thin_pool = parsed_key[1]
+                    thpools.append((volume_group, thin_pool))
+    except KeyError:
```
marmarek, line 173: When KeyError can happen? Shouldn't it skip just that line then, instead of stopping processing everything?

qubes/storage/lvm.py
```diff
@@ -33,19 +33,258 @@
...
+def pool_exists(pool_id):
+    """ Return true if pool exists """
+    try:
+        vol_info = size_cache[pool_id]
+        return vol_info['attr'][0] == 't'
+    except KeyError:
+        return False
+
+
+def list_thin_pools():
```
marmarek, line 161: Where this function is used? Only in anaconda addon? In that case, better add it there and also add relevant `lvs` call there (skipping many columns that are not needed there).

qubes/storage/__init__.py
fepitre: Yes. It should be better.

qubes/storage/lvm.py
fepitre: In some previous attempt it was, not it's not used anymore. I drop it.

qubes/storage/lvm.py
fepitre: Yes it's ugly. I changed the check on `vol['attr']`. I don't remember if in some fake data I was not generating empty strings at some point in case of empty `vol['attr']`.

fepitre:
Moved here: https://github.com/fepitre/qubes-anaconda-addon/commit/3724988f64554b9f7d0c3158563e4e2971dee310

PR #340, opened 2020-04-20
add ability to setup multiply layouts in features

jevank:
Add comma as correct symbol on sanitize. According to setxkbmap manual:               
    Specifies  the  name of the layout used to determine the components which make  up  the  keyboard  description.  The  -layout option may only be used once. Multiple layouts can be specified as a comma-separated list.

### DIFF
--- a/qubes/ext/gui.py
+++ b/qubes/ext/gui.py
@@ -132,10 +132,11 @@ class GUI(qubes.ext.Extension):
         untrusted_variant = untrusted_xkb_layout[1]
         untrusted_options = untrusted_xkb_layout[2]
 
+        re_layout = r'^[a-zA-Z,]*$'
         re_variant = r'^[a-zA-Z0-9-_]*$'
         re_options = r'^[a-zA-Z0-9-_:,]*$'
 
-        if not untrusted_layout.isalpha():
+        if not re.match(re_layout, untrusted_layout):
             raise qubes.exc.QubesValueError("Invalid layout provided")
         if not re.match(re_variant, untrusted_variant):
             raise qubes.exc.QubesValueError("Invalid variant provided")


jevank:
with [this](https://github.com/QubesOS/qubes-gui-agent-linux/pull/96) PR I get worked soluiton for us-ru layout combination.

jevank:
@fepitre could you review it?

marmarek:
How layout switching works in this case? Doesn't setting multiple layouts easily de-synchronize between VMs and dom0?

jevank:
>How layout switching works in this case?

Guest side (qubes-keymap.sh) almost correctly dispatch this key from qubesdb and apply it.
To prevent manual setting in every session I`ve also patched qvm_start_daemon.py, which setting this feature with only first layout from root xprop.
I guess this is correct way to setup default layouts from guivm, as it works on previous versions.  Correct me if I miss smth pls.

>Doesn't setting multiple layouts easily de-synchronize between VMs and dom0?

I guess it needs to prevent populating qubesdb with default guivm value if perhaps features set exactly to vm. It need to make  some changes too.

marmarek:
What about keeping it at a single layout here, but supporting dynamic change which layout is set?

jevank:
Layout changes with qubes-rpc call looks better, agree. Integration with desktop will not be a problem? Can I help smth? I work with 4.1 on main laptop.

marmarek:
It "just" needs a way to detect layout switch at dom0/GUI domain level. TBH, I have no idea how to even check which one is currently active (from those multiple configured ones), leave alone how to get notification about the switch. Requires either googling more about specification, or inspecting some of the keyboard layout widgets code (it somehow gets that info).

jevank:
It may be a good idea to fork [xfce4 keyboard plugin](https://git.xfce.org/panel-plugins/xfce4-xkb-plugin/about/), which allow to store the layout globally (for all windows), per application or per window and indicate current layout. It could store layout per domain and translate it to vm on change.

marmarek:
I don't think there is no need to fork anything. Also, I'm pretty sure much of what you've described is a standard X11 feature, that plugin only exposes an interface to configure it.
A better option would be to use standard mechanisms for detecting layout change, so it would work with any keyboard layout plugin (for example KDE one). It indeed may be a good idea to use the xfce one to learn what those standard mechanisms are.

jevank:
xfce4-keyboard-plugin use gnome libxklavier library, so this is not good for non-gnome environment. I`ve look at [kbdd](https://github.com/qnikst/kbdd) source and [this doc](https://www.x.org/releases/X11R7.7/doc/libX11/XKB/xkblib.html#Xkb_Events).

>The Xkb extension uses events to communicate the keyboard status to interested clients. These events are not associated with a particular window. Instead, all Xkb keyboard status events are reported to all interested clients, regardless of which window currently has the keyboard focus and regardless of the grab state of the keyboard.

Do you think Qubes should use special service like [kbdd](https://github.com/qnikst/kbdd) to detect layout changes and propagate to guests? Or use one if existent (I could not find the candidate)?

P.S. may be it is not best place to discuss, but it has started here...

PR #341, opened 2020-04-24
libvirt xen: Add gfx_passthru and device_video

cytadela8:
Add gfx_passthru and device_video options to libvirt Xen domain template.

By default gfx_passthrou is disabled and video device is enabled as before this change. When we want to pass a GPU device settings gfx_passthru is required and device_video can be disabled as it is not usefull.

Depends on: https://github.com/QubesOS/qubes-core-libvirt/pull/23

Part of: QubesOS/qubes-issues#2618

### DIFF
--- a/templates/libvirt/xen.xml
+++ b/templates/libvirt/xen.xml
@@ -61,13 +61,15 @@
                 <apic/>
                 <viridian/>
             {% endif %}
-
-            {% if vm.devices['pci'].persistent() | list
-                    and vm.features.get('pci-e820-host', True) %}
                 <xen>
+                {% if vm.devices['pci'].persistent() | list
+                      and vm.features.get('pci-e820-host', True) %}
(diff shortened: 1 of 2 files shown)


templates/libvirt/xen.xml
```diff
@@ -186,13 +188,15 @@
                     {% endif %}
                     />
                 <input type="tablet" bus="usb"/>
-                <video>
-                    <model type="{{ vm.features.check_with_template('video-model', 'vga') }}"/>
-                </video>
+                {% if vm.features.check_with_template('device-video', True) %}
```
marmarek, line 191: What do you think about reusing `video-model` with value `none` instead?

marmarek:
The actual code is fine now. But tests needs to be updated to match new libvirt xml layout (`<xen>` features node always present).

PR #342, opened 2020-05-05
Add vm.icon property

marmarta:
This is a property for handling vm icons that change depending on vm type.
Depends on QubesOS/qubes-artwork#17

references QubesOS/qubes-issues#5767


qubes/ext/core_features.py
```diff
@@ -62,3 +62,20 @@ def qubes_features_request(self, vm, event, untrusted_features):
...
+
+    @qubes.ext.handler('property-set:provides_network')
+    def on_property_set(self, subject, event, name, newvalue, oldvalue=None):
+        # pylint: disable=unused-argument
+        self.set_servicevm_feature(subject)
+
+    @qubes.ext.handler('property-del:provides_network')
+    def on_property_del(self, subject, event, name):
+        # pylint: disable=unused-argument
+        self.set_servicevm_feature(subject)
```
marmarek, line 81: This all will work fine for new VMs, because default value for provides_network is false (so default not having `servicevm` feature is correct).
But the above miss adding the feature for existing VMs (on system update from the version before this change). The easiest way to do that is adding similar handler for `domain-load` event (loading a domain from qubes.xml - basically qubesd service start).

qubes/ext/core_features.py
```diff
@@ -62,3 +62,23 @@ def qubes_features_request(self, vm, event, untrusted_features):
...
+
+    # pylint: disable=no-self-use
+    def set_servicevm_feature(self, subject):
+        if subject.provides_network:
+            subject.features['servicevm'] = 1
+        elif 'servicevm' in subject.features:
+            del subject.features['servicevm']
+
+    @qubes.ext.handler('property-set:provides_network')
+    def on_property_set(self, subject, _event, _name, _newvalue,
```
marmarek, line 74: This won't work - event's arguments are keyword arguments so must have specific names (see tests that fail now).

qubes/ext/core_features.py
```diff
@@ -62,3 +62,23 @@ def qubes_features_request(self, vm, event, untrusted_features):
...
+                        _oldvalue=None):
+        self.set_servicevm_feature(subject)
+
+    @qubes.ext.handler('property-del:provides_network')
+    def on_property_del(self, subject, _event, _name):
+        self.set_servicevm_feature(subject)
+
+    @qubes.ext.handler('domain-load')
+    def on_property_del(self, subject, _event):
```
marmarek, line 83: Duplicated function name (see pylint).

qubes/ext/core_features.py
```diff
@@ -62,3 +62,25 @@ def qubes_features_request(self, vm, event, untrusted_features):
             # if this is the first time qrexec was advertised, now can finish
             #  template setup
             yield from vm.fire_event_async('template-postinstall')
+
+    # pylint: disable=no-self-use
+    def set_servicevm_feature(self, subject):
+        if getattr(subject, 'provides_network', False):
+            subject.features['servicevm'] = 1
+        elif 'servicevm' in subject.features:
+            del subject.features['servicevm']
```
ben-grande, line 71: I was deleting `servicevm` necessary? Later on this commit:

Came here from https://github.com/QubesOS/qubes-mgmt-salt-dom0-virtual-machines/pull/63

```python3
    def test_100_servicevm_feature(self):
        self.vm.provides_network = True
        self.ext.set_servicevm_feature(self.vm)
        self.assertEqual(self.features['servicevm'], 1)

        self.vm.provides_network = False
        self.ext.set_servicevm_feature(self.vm)
        self.assertNotIn('servicevm', self.features)
```

I don't think not having `provides_network` must enforce not having `servicevm`.

PR #344, opened 2020-05-12
Rename default root thin pool from 'lvm' to 'root'

fepitre:
New partition output split dom0 and VM thin pools

https://github.com/QubesOS/qubes-anaconda-addon/pull/7
QubesOS/qubes-issues#5763

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,14 @@ import:
   - source: QubesOS/qubes-continuous-integration:R4.1/travis-dom0-r4.1.yml
 python:
   - '3.6'
+language: python
 virtualenv:
   system_site_packages: true
 services:
   - xvfb
 install:
   - sudo apt-get -y install python3-gi gir1.2-gtk-3.0
-  - pip install --quiet -r ci/requirements.txt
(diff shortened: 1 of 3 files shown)


marmarek:
I'm not fully convinced to the `root` name. We have two cases:
1. One thin pool for all (custom partitioning, installing over existing system, perhaps even upgrade from R4.0)
2. Two thin pools

For the first case, storage pool doesn't matter that much, because there is just one lvm pool (although it needs to be obvious enough to distinguish from file `varlibqubes` pool). But for the second case, it needs to be obvious that this pool generally should not be used to store VMs, because it is very small.
IMO the `root` name does not convey this message.

I see two options:
1. Use a different name, maybe `dom0-root`? But this may interfere with the first case.
2. Do not create storage pool for root-pool thin pool at all, if there is separate one for VMs.

I slightly more like the second option. In fact, we can simplify this (I hope) further, by removing the whole automatic storage pool creation in core-admin (let it only create `varlibqubes` and `linux-kernel` pools) and let initial setup create `default` LVM pool based on whatever user have chosen or initial-setup detected. In this case, choosing default thin pool (the largest one) would be done at initial-setup level only.

PR #349, opened 2020-06-15
adding new qubes.LockScreen and cooresponding policy file for issue:

maltfield:
* https://github.com/QubesOS/qubes-issues/issues/5893

### DIFF
--- a/qubes-rpc-policy/qubes.LockScreen.policy
+++ b/qubes-rpc-policy/qubes.LockScreen.policy
@@ -1,0 +1,11 @@
+## Note that policy parsing stops at the first match,
+## so adding anything below "$anyvm $anyvm action" line will have no effect
+
+## Please use a single # to start your custom comments
+
+# WARNING: The qubes.ConnectTCP service is dangerous and allows any
+# qube to access any other qube TCP port. It should be restricted
+# only to restricted qubes. This is why the default policy is 'deny'
+
+# Example of policy: mytcp-client @default allow,target=mytcp-server
+sys-usb dom0 allow
--- a/qubes-rpc/qubes.LockScreen
+++ b/qubes-rpc/qubes.LockScreen
@@ -1,0 +1,2 @@
+#!/bin/bash
+DISPLAY=:0 xscreensaver-command -lock


qubes-rpc/qubes.LockScreen
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
ben-grande, line 1: Bash is unecessary. Use `#!/bin/sh`

qubes-rpc/qubes.LockScreen
```diff
@@ -0,0 +1,2 @@
...
+DISPLAY=:0 xscreensaver-command -lock
```
ben-grande, line 2: Nowaways, I believe we are using `xfce4-screensaver-command -l`. People on `i3` are using `i3lock ...`. Is there a solution not tool dependent?

qubes-rpc-policy/qubes.LockScreen.policy
```diff
@@ -0,0 +1,11 @@
...
+## so adding anything below "$anyvm $anyvm action" line will have no effect
+
+## Please use a single # to start your custom comments
+
+# WARNING: The qubes.ConnectTCP service is dangerous and allows any
+# qube to access any other qube TCP port. It should be restricted
+# only to restricted qubes. This is why the default policy is 'deny'
+
+# Example of policy: mytcp-client @default allow,target=mytcp-server
+sys-usb dom0 allow
```
ben-grande, line 11: This is using the old policy format.

Also, I don't think `sys-usb` should be allowed to lock the screen by default. It shouldn't be `ask` either, should be `deny`.

Also, the comment message is irrelevant, it is talking about TCP service.

maltfield: the purpose of this would be to lock the screen in an emergency. Eg if a udev event is triggered by a BusKill usb drive from being ejected (evidence that someone has physically stolen the laptop) or an unknown usb drive being inserted (eg Mallory trying to insert a rubber ducky or mouse jiggler).

I understand there is some "risk" to let sys-usb lock the screen. But the result of that "risk" is that the computer is in a *more* secure state.

If we have to wait for the user to allow `sys-usb` to lock the screen in a dom0 GUI prompt, it would completely defeat the purpose of this PR.

ben-grande: You don't need to wait, we default to `deny` and you document that in a preceding policy, it should have `allow`.

maltfield: so you're saying that I should change this line to `deny` for the purpose of this PR, and then we update our documentation to tell everyone to manually edit the file by changing `deny` to `allow`?

marmarek: Or, move this whole thing (the service and the policy file) to a separate package. Then, the user could install it and have it automatically enabled. But in any case, this service must be opt-in, not opt-out.

maltfield: @marmarek as a "separate package," are you referring to a [contrib package](https://doc.qubes-os.org/en/latest/developer/general/package-contributions.html)?

If yes, can you please confirm that Qubes devs would accept a contrib package with a policy file that defaults to `sys-usb dom0 allow`, if it's done in a contrib package (as opposed to being built-into Qubes directly, as this PR is doing)?

ben-grande: The important part is locking the screen according to the WM and screenlocker the user has, see this that is used on the app menu: https://github.com/QubesOS/qubes-desktop-linux-menu/blob/bc5aaa8e13b22c213113d0eabbcccbd80a10dc89/qubes_menu/appmenu.py#L203.

And most likely yes to the other question.

PR #350, opened 2020-06-17
Changed feature keyboard_layout to a property

marmarta:
Purpose: make it easier to implement more robust keyboard layout tools and
propagation.

references QubesOS/qubes-issues#1396
references QubesOS/qubes-issues#4294

### DIFF
--- a/qubes/ext/gui.py
+++ b/qubes/ext/gui.py
@@ -21,8 +21,6 @@
 # License along with this library; if not, see <https://www.gnu.org/licenses/>.
 #
 
-import re
-
 import qubes.config
 import qubes.ext
 import qubes.exc
@@ -89,14 +87,12 @@ class GUI(qubes.ext.Extension):
 
         # Add GuiVM Xen ID for gui-daemon
         if getattr(vm, 'guivm', None):
-            if vm != vm.guivm and vm.guivm.is_running():
-                vm.untrusted_qdb.write('/qubes-gui-domain-xid',
-                                       str(vm.guivm.xid))
+            if vm != vm.guivm:
+                vm.untrusted_qdb.write('/keyboard-layout', vm.keyboard_layout)
 
-            # Add keyboard layout from that of GuiVM
-            kbd_layout = vm.guivm.features.get('keyboard-layout', None)
-            if kbd_layout:
-                vm.untrusted_qdb.write('/keyboard-layout', kbd_layout)
+                if vm.guivm.is_running():
+                    vm.untrusted_qdb.write('/qubes-gui-domain-xid',
+                                           str(vm.guivm.xid))
 
         # Set GuiVM prefix
         guivm_windows_prefix = vm.features.get('guivm-windows-prefix', 'GuiVM')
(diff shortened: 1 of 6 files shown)


qubes/vm/adminvm.py
```diff
@@ -61,6 +62,13 @@ class AdminVM(qubes.vm.BaseVM):
         setter=qubes.property.forbidden,
         doc='True if this machine may be updated on its own.')
 
+    # for changes in keyboard_layout, see also the same property in QubesVM
+    keyboard_layout = qubes.property(
+        'keyboard_layout',
+        type=str,
+        setter=_setter_kbd_layout,
+        doc='Keyboard layout for this VM')
```
marmarek, line 70: Missing default. Or was it intentional?

qubes/ext/gui.py
```diff
@@ -121,22 +117,15 @@ def on_domain_start(self, vm, event, **kwargs):
...
-        if not re.match(re_variant, untrusted_variant):
-            raise qubes.exc.QubesValueError("Invalid variant provided")
-        if not re.match(re_options, untrusted_options):
-            raise qubes.exc.QubesValueError("Invalid options provided")
+    @qubes.ext.handler('property-reset:keyboard_layout')
+    def on_keyboard_reset(self, subject, event, name, oldvalue=None):
+        if getattr(subject, 'guivm'):
+            kbd_layout = subject.guivm.keyboard_layout
+        else:
+            kbd_layout = 'us++'
```
marmarek, line 125: Since the logic for getting the default is already embedded into the property, you can simply use `subject.keyboard_layout` here.

qubes/ext/gui.py
```diff
@@ -121,22 +117,15 @@ def on_domain_start(self, vm, event, **kwargs):
...
+
+        subject.untrusted_qdb.write('/keyboard-layout', kbd_layout)
```
marmarek, line 127: Check if vm is running first, qubesdb works only when vm is running.

qubes/ext/gui.py
```diff
@@ -121,22 +117,15 @@ def on_domain_start(self, vm, event, **kwargs):
...
+
+    @qubes.ext.handler('property-set:keyboard_layout')
+    def on_keyboard_set(self, subject, event, name, newvalue, oldvalue=None):
+        subject.untrusted_qdb.write('/keyboard-layout', newvalue)
```
marmarek, line 131: And here too check if vm is running.

qubes/tests/vm/qubesvm.py
```diff
@@ -1968,8 +1971,90 @@ def test_624_qdb_guivm_invalid_keyboard_layout(self, mock_qubesdb,
...
+        }
+
+        with self.subTest('default'):
+            vm.create_qdb_entries()
+            self.assertEqual(test_qubesdb.data, expected)
+
+        with self.subTest('value_change'):
+            vm.keyboard_layout = 'de++'
+            expected['/keyboard-layout'] = 'de++'
+            vm.create_qdb_entries()
```
marmarek, line 2042: Event handler should make this call unnecessary (and its's worth checking if that's really the case).

qubes/tests/vm/qubesvm.py
```diff
@@ -1968,8 +1971,90 @@ def test_624_qdb_guivm_invalid_keyboard_layout(self, mock_qubesdb,
...
+            self.assertEqual(test_qubesdb.data, expected)
+
+        with self.subTest('value_revert'):
+            vm.keyboard_layout = qubes.property.DEFAULT
+            expected['/keyboard-layout'] = 'fr++'
+            vm.create_qdb_entries()
```
marmarek, line 2048: Same.

qubes/tests/vm/qubesvm.py
```diff
@@ -1968,8 +1971,90 @@ def test_624_qdb_guivm_invalid_keyboard_layout(self, mock_qubesdb,
...
+            self.assertEqual(test_qubesdb.data, expected)
+
+        with self.subTest('no_default'):
+            guivm.keyboard_layout = qubes.property.DEFAULT
+            vm.keyboard_layout = qubes.property.DEFAULT
+            expected['/keyboard-layout'] = 'us++'
+            vm.create_qdb_entries()
```
marmarek, line 2055: Same.

qubes/tests/vm/qubesvm.py
```diff
@@ -1968,8 +1971,90 @@ def test_624_qdb_guivm_invalid_keyboard_layout(self, mock_qubesdb,
...
+            '/qubes-iptables-error': '',
+            '/qubes-iptables-header': unittest.mock.ANY,
+            '/qubes-service/qubes-update-check': '0',
+            '/qubes-service/meminfo-writer': '1',
+            '/connected-ips': '',
+            '/connected-ips6': '',
+        }
+
+        with self.subTest('default'):
+            vm.create_qdb_entries()
```
marmarek, line 2036: Already called above.

qubes/ext/gui.py
```diff
@@ -93,10 +91,8 @@ def on_domain_qdb_create(self, vm, event):
                 vm.untrusted_qdb.write('/qubes-gui-domain-xid',
                                        str(vm.guivm.xid))
 
-            # Add keyboard layout from that of GuiVM
-            kbd_layout = vm.guivm.features.get('keyboard-layout', None)
-            if kbd_layout:
-                vm.untrusted_qdb.write('/keyboard-layout', kbd_layout)
+        # Add keyboard layout
+        vm.untrusted_qdb.write('/keyboard-layout', vm.keyboard_layout)
```
marmarek, line 95: This (and other below) should be set only if all:
 - vm has some guivm
 - vm is not its own guivm
 - vm is running (it is already the case in this handler, but not necessary the others)

Otherwise a keyboard layout set via desktop environment tools in guivm (like xfce settings) would be overridden by this property (possibly reverting recent user change)

qubes/ext/gui.py
```diff
@@ -121,22 +117,15 @@ def on_domain_start(self, vm, event, **kwargs):
...
+    @qubes.ext.handler('property-reset:keyboard_layout')
+    def on_keyboard_reset(self, vm, event, name, oldvalue=None):
+        if not vm.is_running():
+            return
+        kbd_layout = vm.keyboard_layout
+
+        vm.untrusted_qdb.write('/keyboard-layout', kbd_layout)
+
+    @qubes.ext.handler('property-set:keyboard_layout')
+    def on_keyboard_set(self, vm, event, name, newvalue, oldvalue=None):
```
marmarek, line 129: This needs also to loop over VMs for which it is set as guivm and keyboard_layout property is default there (in short: VMs where keyboard_layout value just changed), and fire `property-reset:keyboard_layout` event.

qubes/ext/gui.py
```diff
@@ -121,22 +117,22 @@ def on_domain_start(self, vm, event, **kwargs):
...
+        for domain in vm.app.domains:
+            if getattr(domain, 'guivm', None) == vm and \
+                    domain.property_is_default('keyboard_layout'):
+                domain.fire_event('property-reset:keyboard_layout',
+                                  name='keyboard_layout', newvalue=newvalue,
```
marmarek, line 134: There is no `newvalue` argument to property-reset event.

PR #351, opened 2020-06-22
report unexpected options in admin.pool.Add; clean up pool __init__()

rustybird:


rustybird:
"fatal: destination path '/home/travis/qubes-builder' already exists and is not an empty directory." wat

marmarek:
Fixed in master, you can rebase.

PR #352, opened 2020-06-23
Audio: rework audio tests

easydozen:
QubesOS/qubes-issues#2624

Replace constant audio bytearray with generated sine wave sample
to get more robust results across test environments.

Use zero-crossings as an audio fingerprint
and compare it with sine wave frequency.

### DIFF
--- a/qubes/tests/integ/vm_qrexec_gui.py
+++ b/qubes/tests/integ/vm_qrexec_gui.py
@@ -551,11 +551,14 @@ class TC_00_AppVMMixin(object):
                 'parecord --raw --format=float32le --rate=44100 \
                             --channels=1 audio_rec.raw'))
         # give it time to start recording
+        self.loop.run_until_complete(asyncio.sleep(0.5))
         p = subprocess.Popen(['sudo', '-E', '-u', local_user,
             'paplay', '--raw', '--format=float32le',
                     '--rate=44100', '--channels=1'],
             stdin=subprocess.PIPE)
         p.communicate(audio_in.astype(np.float32).tobytes())
+        # wait for possible parecord buffering
+        self.loop.run_until_complete(asyncio.sleep(1))
         self.loop.run_until_complete(
             self.testvm1.run_for_stdio('pkill parecord || :'))
         _, record_stderr = self.loop.run_until_complete(record.communicate())
--- a/qubes/tests/integ/vm_qrexec_gui.py
+++ b/qubes/tests/integ/vm_qrexec_gui.py
@@ -430,7 +430,7 @@ class TC_00_AppVMMixin(object):
         # and some more...
         self.loop.run_until_complete(asyncio.sleep(1))
 
-    def is_vm_suitable(self):
+    def prepare_audio_vm(self):
         if 'whonix-gw' in self.template:
             self.skipTest('whonix-gw have no audio')
         self.loop.run_until_complete(self.testvm1.start())
@@ -439,10 +439,10 @@ class TC_00_AppVMMixin(object):
                 self.testvm1.run_for_stdio('which parecord'))
         except subprocess.CalledProcessError:
             self.skipTest('pulseaudio-utils not installed in VM')
+        self.wait_for_pulseaudio_startup(self.testvm1)
+
 
     def common_audio_playback(self):
-        self.is_vm_suitable()
-        self.wait_for_pulseaudio_startup(self.testvm1)
         # sine frequency
         sfreq = 4400
         # generate signal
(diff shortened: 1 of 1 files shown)


marmarek:
This looks like a great improvement!

qubes/tests/integ/vm_qrexec_gui.py
```diff
@@ -535,53 +534,79 @@ def test_221_audio_record_muted(self):
...
+    @unittest.skipUnless(spawn.find_executable('parecord'),
+                         "pulseaudio-utils not installed in dom0")
+    def test_222_audio_rec_unmuted(self):
+        self.common_audio_record_unmuted()
+
+    @unittest.skipUnless(spawn.find_executable('parecord'),
+                         "pulseaudio-utils not installed in dom0")
+    def test_223_audio_play_hvm(self):
+        self.testvm1.virt_mode = 'hvm'
+        self.testvm1.features['audio-model'] = 'ich6'
```
marmarek, line 594: I've tested it on a system without the stubdomain changes included and it still passed. I think some extra step is needed to verify if it really uses stubdomain audio. Or do you think core-admin-client change connecting `pacat-simple-vchan` to stubdomain only is enough?

Another thing is, if the `module-vchan-sink` is still loaded in the VM, it will create an output even if dom0/sys-audio part is not running. And it will freeze the playback until `pacat-simple-vchan` is started. Perhaps by coincidence the emulated audio card is chosen by default, but this sounds like a risky approach to rely on it. Better set the default audio sink explicitly. And while doing this, you can also check if emulated card is detected, as a preliminary test (this would solve the issue above too).

qubes/tests/integ/vm_qrexec_gui.py
```diff
@@ -441,21 +440,30 @@ def test_220_audio_playback(self):
...
+                recorded_audio.name], stdout=subprocess.PIPE)
             try:
                 self.loop.run_until_complete(
-                    self.testvm1.run_for_stdio('paplay --raw audio_in.raw'))
+                    self.testvm1.run_for_stdio(
+                    'paplay --format=float32le --rate=44100 \
+                            --channels=1 --raw audio_in.raw'))
             except subprocess.CalledProcessError as err:
                 self.fail('{} stderr: {}'.format(str(err), err.stderr))
             # wait for possible parecord buffering
```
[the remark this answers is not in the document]
easydozen: I'm going to add basic test with silence threshold.

qubes/tests/integ/vm_qrexec_gui.py
easydozen: Don't you mind if I just remove guest' module-vchan-sink in corresponding hvm tests? (pacmd unload-module)

qubes/tests/integ/vm_qrexec_gui.py
```diff
@@ -441,21 +440,30 @@ def test_220_audio_playback(self):
...
-        audio_in = b'\x20' * 44100
-        self.loop.run_until_complete(
-            self.testvm1.run_for_stdio('cat > audio_in.raw', input=audio_in))
+        # sine frequency
+        sfreq = 4400
+        # generate signal
+        audio_in = np.sin(2*np.pi*np.arange(44100)*sfreq/44100)
+        self.loop.run_until_complete(self.testvm1.run_for_stdio(
+            'pacat --raw --rate=44100 --channels=1 \
+            --format=float32le > audio_in.raw',
```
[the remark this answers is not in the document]
easydozen: You're right. That was silly.

qubes/tests/integ/vm_qrexec_gui.py
marmarek: That should work too.

marmarek:
Interestingly, `test_223_audio_play_hvm` fails for debian-10 based VMs (whonix-ws-15 is debian 10 too), but works on Fedora. At least in openQA environment. Manually inspecting such VM, just after failed test, I don't see anything wrong here. Maybe some race condition?

easydozen:
> Interestingly, `test_223_audio_play_hvm` fails for debian-10 based VMs (whonix-ws-15 is debian 10 too), but works on Fedora. At least in openQA environment. Manually inspecting such VM, just after failed test, I don't see anything wrong here. Maybe some race condition?

It seams sink starts muted. I suppose this happens due to one of  PA *-restore modules.
pacmd set-sink-mute and pacmd set-sink-volume should help.

easydozen:
Actually, that was known PA issue. 
alsa-utils package installed in template will help to properly restore sink.

PR #354, opened 2020-06-25
callback pool driver

3hhh:
Proxy storage pool driver adding callback functionality to other pool drivers.

Can help solve the following issues:
https://github.com/QubesOS/qubes-issues/issues/904
https://github.com/QubesOS/qubes-issues/issues/1293

Related discussion with some design explanations:
https://github.com/QubesOS/qubes-issues/issues/5895

**Notes**:

I didn't implement any protection against concurrency (yet?) simply because I don't fully understand the Pool & Volume APIs wrt parallelism.

In particular I wondered:
- What parts of the Pool API can be called in parallel by Qubes OS for a single Pool instance?
- What parts of the Volume API can be called in parallel by Qubes OS for a single Volume instance?

If Qubes OS doesn't protect against certain race conditions, a lock might have to be added here and there.

Other than that, more testing from someone != me would certainly not hurt. ;-)


3hhh:
Hm and looks like my rpm skills might be insufficient @travis `qubes/storage/qubes_callback.json.example`...

marmarek:
> I didn't implement any protection against concurrency (yet?) simply because I don't fully understand the Pool & Volume APIs wrt parallelism.

The API itself doesn't give much guarantee about parallelism on it's own. Most (async) functions can be called in parallel. There are some exceptions imposed by other parts of the API - like `start` and `stop` for the same volume can't happen at the same time. The key thing is "async functions" - right now none of your functions are async (see asyncio library), which means any of them will freeze the whole qubesd process. This is okay-ish for very short actions, but since the driver allows connecting arbitrary scripts, that is definitely not good here.
See lvm driver for examples about running processes in async functions and also about locking.

> Other than that, more testing from someone != me would certainly not hurt. ;-)

You can start with looking pylint output (see travis) - there are quite a few code style issues, but some are actually harmful.

As a general comment, it might be better to make a separate repository from this. By using entry_points it doesn't really matter where the implementation is stored - the important part is setup.py line pointing at it. See [here](https://github.com/QubesOS/qubes-core-admin-addon-whonix) for example addon structure (that one is for handling events, not storage driver, but the structure is very similar).

> Hm and looks like my rpm skills might be insufficient @travis `qubes/storage/qubes_callback.json.example`...

I don't think `/usr/lib/...` is the right place for example config. I'd rather place it in either documentation, or directly /etc. You can copy it with an independent `cp`/`install` call in the `Makefile` and then include the path in rpm spec.

qubes/storage/callback.py
```diff
@@ -0,0 +1,442 @@
...
+    def destroy(self):
+        self._assertInitialized()
+        ret = self._cb_impl.destroy()
+        self._callback('on_destroy')
+        return ret
+
+    def init_volume(self, vm, volume_config):
+        return CallbackVolume(self, self._cb_impl.init_volume(vm, volume_config))
+
+    def setup(self, callback=True):
```
marmarek, line 286: Please keep original parameters (none here). If you need to call the same from elsewhere, with custom params, add a helper function with those params, then call it here too.

qubes/storage/callback.py
```diff
@@ -0,0 +1,442 @@
...
+
+    def list_volumes(self):
+        for vol in self._cb_impl.list_volumes():
+            yield CallbackVolume(self, vol)
+
+    def get_volume(self, vid):
+        return CallbackVolume(self, self._cb_impl.get_volume(vid))
+
+    def included_in(self, app):
+        if self._cb_requires_init:
```
marmarek, line 305: If it requires late init, you can also check if the init is already done - in which case call the actual implementation. This way, at least sometimes you'll get accurate data.

qubes/storage/callback.py
```diff
@@ -0,0 +1,442 @@
...
+            return None
+        else:
+            return self._cb_impl.included_in(app)
+
+    @property
+    def size(self):
+        if self._cb_requires_init:
```
marmarek, line 312: Same here.

qubes/storage/callback.py
```diff
@@ -0,0 +1,442 @@
...
+            return None
+        else:
+            return self._cb_impl.size
+
+    @property
+    def usage(self):
+        if self._cb_requires_init:
```
marmarek, line 319: And here.

qubes/storage/callback.py
```diff
@@ -0,0 +1,442 @@
...
+    def create(self):
+        self._assertInitialized()
+        self._callback('on_volume_create')
+        return self._cb_impl.create()
+
+    def remove(self):
+        self._assertInitialized()
+        ret = self._cb_impl.remove()
+        self._callback('on_volume_remove')
+        return ret
```
rustybird, line 380: Here you're actually doing the callback after the backend's `remove` method has started (not finished!) if it's a coroutine, i.e. with ThinVolume and ReflinkVolume. As a quick and dirty fix, I think you could check if `ret` is a coroutine object and if so, splice in `ret = asyncio.get_event_loop().run_until_complete(ret)`. But the real fix would be to make your method a coroutine too and use `qubes.utils.coro_maybe`. Then you'd also need an async `_callback` variant. Ah, function coloring hell...

Same in `CallbackVolume.stop` and `CallbackVolume.import_data_end`, and theoretically in `CallbackPool.destroy`.

marmarek: No, you definitely can't do `asyncio.get_event_loop().run_until_complete()`. It is not legal operation if event loop is running elsewhere (it is). This needs to be a proper coroutine.

3hhh:
You guys are too fast for me...
Next time you can choose between my code and sleep, please don't consider sacrificing your health for my code. ;-)

Anyway thanks a lot for the comments & support once again!

I'll comment on the ones where I don't fully agree and hopefully fix the remaining issues incl. pylint, asyncio & proper building in the next 1-2 weeks.

> As a general comment, it might be better to make a separate repository from this. By using entry_points it doesn't really matter where the implementation is stored - the important part is setup.py line pointing at it. See [here](https://github.com/QubesOS/qubes-core-admin-addon-whonix) for example addon structure (that one is for handling events, not storage driver, but the structure is very similar).

I don't really like that idea for the following reasons:
- PRs & fixes related to storage drivers would simply forget about that "other repo there"
- putting 500 loc in a separate repo appears to be overkill for me (incl. for maintenance)

If you don't like the risk associated with an allowed `admin.pool.Add` RPC policy in combination with this driver, please recall that this driver won't do anything, if the config doesn't exist (you'll currently see some stack trace in journalctl about it missing). If the config exists, the user is hopefully aware of this risk, the `config_id` must be correctly guessed by an attacker _and_ the code that he possibly doesn't know would have to cause something useful for him (which is unlikely as it does something for a different non-attacker-VM & probably breaks something causing the user to look into it). So while I agree that this is not perfect, it felt like too risky and the best I could do with the current RPC policy.

If future improvements to the RPC policy granularity make it possible to impose limits on constructor arguments as you suggested above, this would be fixed. A short-term alternative might be to limit this driver to `dom0` usage only, yes.

The config would still have to remain though as you'd otherwise into policy management hell wrt limits on what command should be allowed or not...

qubes/storage/callback.py
```diff
@@ -0,0 +1,442 @@
...
+    Check the `qubes_callback.json.example` for an overview of the available options.
+
+    Example applications of this driver:
+        - custom pool mounts
+        - encryption
+        - debugging
+
+
+    **Integration tests**:
+    (all of these tests assume the `qubes_callback.json.example` configuration)
```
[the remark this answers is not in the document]
3hhh: I agree.

Manual tests felt better than no tests though. In fact I'm not sure whether I'll get proper unit tests done any time soon.

qubes/storage/callback.py
```diff
@@ -0,0 +1,442 @@
...
+    config_path='/etc/qubes_callback.json'
+
+    def __init__(self, *, name, conf_id):
+        '''Constructor.
+        :param conf_id: Identifier as found inside the user-controlled configuration at `/etc/qubes_callback.json`.
+                       Non-ASCII, non-alphanumeric characters may be disallowed.
+                       **Security Note**: Depending on your RPC policy (admin.pool.Add) this constructor and its parameters
+                       may be called from an untrusted VM (not by default though). In those cases it may be security-relevant
+                       not to pick easily guessable `conf_id` values for your configuration as untrusted VMs may otherwise
+                       execute callbacks meant for other pools.
```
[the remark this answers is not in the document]
3hhh: See my long text in the end.

qubes/storage/callback.py
```diff
@@ -0,0 +1,442 @@
...
+                        Only passed on for the generic command specified as `cmd`, not for `on_xyz` callbacks.
+        '''
+        if self._cb_ctor_done:
+            cmd = self._cb_conf.get(cb)
+            args = [] #on_xyz callbacks should never receive arguments
+            if not cmd:
+                cmd = self._cb_conf.get('cmd')
+                args = [ self.name, self._cb_conf['bdriver'], cb, self._cb_cmd_arg, *cb_args ]
+            if cmd and cmd != '-':
+                args = filter(None, args)
```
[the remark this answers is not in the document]
3hhh: Only `cb_args` holds a variable number of arguments (the comment line above the code is wrong & will be fixed).

So the first four positions are fixed, then follows a variable number. Most script authors should be able to handle that. Otherwise they can stick to the more simple `on_xyz` callbacks that don't receive any arguments.

qubes/storage/callback.py
```diff
@@ -0,0 +1,442 @@
...
+                args = ' '.join(quote(str(a)) for a in args)
+                cmd = ' '.join(filter(None, [cmd, args]))
+                log.info('callback driver executing (%s, %s %s): %s' % (self._cb_conf_id, cb, cb_args, cmd))
+                res = subprocess.run(cmd, shell=True, check=True, executable='/bin/bash', stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
```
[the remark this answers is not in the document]
3hhh: Your proposal should do the same as mine, cf. https://github.com/python/cpython/blob/3.5/Lib/subprocess.py#L1175-L1178

Anyway yours is more readable, so I'll change that.

qubes/storage/callback.py
3hhh: `self._cb_requires_init` is set to `False` after storage initialization, so the `else`-path should handle that.

3hhh:
I think I ran into another issue with the `qubes.storage.Volume` API in combination with `asyncio` for this particular driver:

`Volume.export()` and `Volume.block_device()` seem to be required to be synchronous.

However with the `callback` driver, they both require storage initialization to provide a reasonable return value (That's why I called `self._assertInitialized()` in my code so far) - the backend driver probably won't be able to say much without having access to the actual data.

Storage initialization may involve calling asynchronous methods such as `setup()` though.

So how can I call asynchronous code from synchronous functions?

If I understand [1] correctly, it simply won't work?

Do you happen to be aware of any workaround or should I just trigger an exception in that case?

[1] https://rokups.github.io/#!pages/python3-asyncio-sync-async.md

qubes/storage/callback.py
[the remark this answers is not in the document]
marmarek: But still - why filter `None`, instead of passing it to the script? It would give the script parameters always at the same positions then.

marmarek:
> Storage initialization may involve calling asynchronous methods such as `setup()` though.

`block_device()` is easy - in any sensible case it is guaranteed to have start() called before. I'm not 100% it covers all the cases, but if `block_device()` was called before `start()`, it is safe to simply return `None`, as its value will not be used in real VM config anyway.

But `export()` indeed may be problematic. I think not allowing async there is an oversight, I'll add it.

3hhh:
I just pushed the discussed changes.

Some notes:
- I added a doc note as I'm not 100% sure whether all backend drivers will work with async code: Calling sync code asynchronously may not be what the backend driver programmer had intended. My manual tests with the file pool worked, but I guess race conditions are not so obvious.
- `Volume.export()` may trigger an exception for now.
- Wrt unit tests I had hoped that code from one of the other pool drivers (preferably lvm or reflink) could be re-used by just injecting a callback pool with the respective backend instead, but unfortunately the code doesn't look trivial to me. :-(
Anyway it would only have tested whether a callback pool behaves exactly like the backend driver, not the added functionality.
- I like long lines, sorry @travis...

3hhh:
P.S.: `_nocoro` is unrelated to the pandemic. ;-)

marmarek:
>     * Calling sync code asynchronously may not be what the backend driver programmer had intended.

For flexibility, see `qubes.utils.coro_maybe`. 

>  Anyway it would only have tested whether a callback pool behaves exactly like the backend driver, not the added functionality.

Yes, I think this is fine for now. You can simply add test callbacks like "touch /tmp/callback-X-called" and then check for existence.

> I like long lines, sorry @travis...

See https://www.qubes-os.org/doc/coding-style/#general-typographic-conventions, this isn't only travis being picky.

marmarek:
Some idea: if there would be a callback allowing to override volume path (returned in `block_device()`, but probably makes more sense to do that in `start()`), then it would be possible to apply per-volume encryption, not only per-pool one. Especially useful for volatile volumes for DispVMs (you just linked) - use random ephemeral key.
So, the idea would be to let lvm/file/reflink pool provide plain volume, then callback driver would take that and build encryption layer (dm-crypt probably) on top - and let the VM use that new device, instead of passing through original one directly.
What do you think?

3hhh:
> Some idea: if there would be a callback allowing to override volume path (returned in `block_device()`, but probably makes more sense to do that in `start()`), then it would be possible to apply per-volume encryption, not only per-pool one. Especially useful for volatile volumes for DispVMs (you just linked) - use random ephemeral key.
> So, the idea would be to let lvm/file/reflink pool provide plain volume, then callback driver would take that and build encryption layer (dm-crypt probably) on top - and let the VM use that new device, instead of passing through original one directly.
> What do you think?

Hmmm....

Assuming the file driver as an example:
If you really only want the volatile part to be encrypted, a post (backend driver) `Volume.start()` callback would indeed be what you want:
You could move the cow file to some path based on an encrypted device with key in `ramfs` (`tmpfs` can be swapped btw) or just move it to `ramfs` entirely, put a symbolic link where the original cow was and then pray that the file driver works with symbolic links (not sure).

Qubes OS shouldn't attempt to write to the cow before as the pool didn't return control yet.

The only thing I don't like about this approach is that it is even more likely to fail open (i.e. leave the user without encryption without him being able to notice it): If you silently fail the `mv cow.img` in the above example, all would go to disk.

Admittedly this is not too different from the example I had put in the code though as the file pool doesn't really seem to care whether its base directories created during `setup()` exist or not and proceeds regardless. It could be different for another backend driver though, one might consider this a bug or one could point it to some path it cannot create without help from the user code.

In total I think I'll add a post start callback.

Is there some way to obtain the template name from within a `Volume` object?
I guess that would be useful inside an external callback to decide whether one wants to do such an on-the-fly swapping or not.

P.S.: I just tested my code with dispVMs and unfortunately it doesn't work yet as dispVMs are apparently created in a different way than regular VMs. I have a bugfix, but I need to test that further.

marmarek:
> Is there some way to obtain the template name from within a `Volume` object?

By design, no. Volume that is supposed to be based on a template's volume (with `snap_on_start=True`) have a reference to that base volume in `source` attributed. 

> The only thing I don't like about this approach is that it is even more likely to fail open (i.e. leave the user without encryption without him being able to notice it): If you silently fail the `mv cow.img` in the above example, all would go to disk.

Yes, that's one of my worries about this approach. But in some cases it may be even more tricky - in case of lvm, the pool driver points at `/dev/` entries, which are managed by `udev` - if you do symlink tricks there, you can be very unlikely.

3hhh:
> > Some idea: if there would be a callback allowing to override volume path (returned in `block_device()`, but probably makes more sense to do that in `start()`), then it would be possible to apply per-volume encryption, not only per-pool one. Especially useful for volatile volumes for DispVMs (you just linked) - use random ephemeral key.
> > So, the idea would be to let lvm/file/reflink pool provide plain volume, then callback driver would take that and build encryption layer (dm-crypt probably) on top - and let the VM use that new device, instead of passing through original one directly.
> > What do you think?
> 
> Hmmm....
> 
> Assuming the file driver as an example:
> If you really only want the volatile part to be encrypted, a post (backend driver) `Volume.start()` callback would indeed be what you want:
> You could move the cow file to some path based on an encrypted device with key in `ramfs` (`tmpfs` can be swapped btw) or just move it to `ramfs` entirely, put a symbolic link where the original cow was and then pray that the file driver works with symbolic links (not sure).
> 
> Qubes OS shouldn't attempt to write to the cow before as the pool didn't return control yet.

Another alternative for a separate volatile part:

Stick with the `on_sinit` approach of my example, but eliminate its single disadvantage, namely the need to re-create the disposable template inside the ephemeral pool on reboot with the related increased storage requirements leading to longer startup delays (larger encrypted storage required to be initialised on boot).
(The RAM usage is irrelevant as it only holds the 100 bit key.)

For the file driver you can do that by keeping an unencrypted copy of the disposable template and just `mount --rebind` it to the right place during `on_sinit`.

For the lvm driver you might be able to do something similar by backing the private lvm pool from some static pre-saved data on your disk and backing the volatile pool from a cryptsetup device (assuming lvm supports that; I'm no lvm expert).

In total the concept should be applicable to all backend drivers where using/including pre-saved data & encryption of a single volume is somehow possible.

Side note if anyone thinks one key per disposable VM start would increase security:
Yes it does, but 16 keys will increase your security by 4 bit... You can also just choose a 150 Bit key and get the same as 2^50 such additional keys. Even more so since a relatively small pool will also make it more likely that your older data is overwritten before forensic disk analysis anyway. Don't fall for security snake oil. ;-)

3hhh:
P.S.: With the above approach, one could also create a root template inside such a pool to avoid the root-cow and volatile volume to be on non-encrypted storage (even with https://github.com/QubesOS/qubes-issues/issues/5933 fixed it would still apply for at least the root snapshots going to unencrypted storage otherwise, cf. https://github.com/QubesOS/qubes-issues/issues/904#issuecomment-508295453).

3hhh:
The version I just pushed contains the fix to work with disposable VMs.

qubes/storage/callback.py
```diff
@@ -442,7 +442,7 @@ def _assert_initialized(self, **kwargs):
     def _callback(self, cb, cb_args=None, **kwargs):
         if cb_args is None:
             cb_args = []
-        vol_args = [self.name, self.vid, *cb_args]
+        vol_args = [self.name, self.vid, self.source, *cb_args]
```
marmarek, line 446: And this is yet another example why `filter(None, ...)` on args is a terrible idea. If `self.source` is None for a given volume, then other arguments (like `size` for resize for example) will be in different place.

3hhh: Fully true.

But I had (admittedly somewhat silently) dropped the line in https://github.com/QubesOS/qubes-core-admin/commit/4becf29c342b4ddd8afd6425608054a145d1c5db#diff-271f0ec3445d2fb476cb711090e750b8L252 anyway.

The other `filter()` two lines below is still somewhat required to remove an entirely empty args String `''` / avoid adding an unnecessary space after a command, if there are no arguments (args is a String by then).

qubes/storage/lvm.py
```diff
@@ -528,7 +528,7 @@ def import_volume(self, src_volume):
         # HACK: neat trick to speed up testing if you have same physical thin
         # pool assigned to two qubes-pools i.e: qubes_dom0 and test-lvm
         # pylint: disable=line-too-long
-        if isinstance(src_volume.pool, ThinPool) and \
+        if hasattr(src_volume.pool, 'thin_pool') and \
```
marmarek, line 531: I know it's tempting, but this may bypass some of the callbacks and cause the operation to fail or copy wrong data (if callbacks setup some extra layer like encryption on top).

3hhh:
Fyi:
I'll push a rebased variant (wrt https://github.com/QubesOS/qubes-issues/issues/5935 fixes - thanks again!) and probably some more tests soon.

3hhh:
A question:

All of the existing storage tests use
```
        self.app = qubes.Qubes.create_empty_store(store=xml_path,
            clockvm=None,
            updatevm=None,
            offline_mode=True,
        )
```
and create the pool afterwards.

I'd however prefer to test with a pre-existing pool (i.e. have it inside the `qubes.xml`).

How would I do that?

3hhh:
> A question:
> 
> All of the existing storage tests use
> 
> ```
>         self.app = qubes.Qubes.create_empty_store(store=xml_path,
>             clockvm=None,
>             updatevm=None,
>             offline_mode=True,
>         )
> ```
> 
> and create the pool afterwards.
> 
> I'd however prefer to test with a pre-existing pool (i.e. have it inside the `qubes.xml`).
> 
> How would I do that?

No worries, I found https://github.com/QubesOS/qubes-core-admin/blob/master/qubes/tests/app.py#L387-L440

3hhh:
> I'll push a rebased variant (wrt [QubesOS/qubes-issues#5935](https://github.com/QubesOS/qubes-issues/issues/5935) fixes - thanks again!) and probably some more tests soon.

That was it.

The tests didn't reveal any further issues. So I guess I'm done unless you see any further issues.

qubes/tests/storage_callback.py
```diff
@@ -0,0 +1,396 @@
...
+    bak_pool_conf = None
+    conf_id = None
+
+    @classmethod
+    def setUpClass(cls, conf_id='utest-callback-01'):
+        CallbackBase.bak_pool_class = qubes.tests.storage_lvm.POOL_CLASS
+        CallbackBase.bak_volume_class = qubes.tests.storage_lvm.VOLUME_CLASS
+        CallbackBase.bak_pool_conf = qubes.tests.storage_lvm.POOL_CONF
+        qubes.tests.storage_lvm.POOL_CLASS = POOL_CLASS
+        qubes.tests.storage_lvm.VOLUME_CLASS = VOLUME_CLASS
```
marmarek, line 107: Please, don't. You can move POOL_CLASS and co to a class attribute then override in children classes.

qubes/storage/callback.py
```diff
@@ -0,0 +1,648 @@
...
+        super().__init__(name=name, revisions_to_keep=int(bdriver_args.get('revisions_to_keep', 1)))
+        self._cb_ctor_done = True
+        self._callback_nocoro('post_ctor')
+
+    @asyncio.coroutine
+    def _coro_maybe(self, value):
+        if asyncio.iscoroutine(value):
+            return (yield from value)
+        with (yield from self._cb_sync_lock):
+            return value
```
marmarek, line 242: This doesn't run that function under the lock, it simply takes a lock and instantly releases it (as `return value` doesn't call anything).
I'm not sure what exactly you're trying to protect here  (two synchronous functions can't in python run in parallel, by definition - synchronous function blocks python interpreter until it finishes), but if you want to ensure no sync function is running in parallel with some async one of the same driver (in some unusual case where the driver implements some sync and some async), then you should obtain the `value` (like the call to `self._cb_impl.setup()`) under the lock in the first place.

qubes/storage/lvm.py
marmarek: >  during setup it would immediately max out the pool at 1 GB because I'd put 1 GB random data in the thin volume (otherwise I'd leak the size of the data it keeps).

The assumption here may not be true in all the cases. Not filling free space with random data (or perhaps just encrypted zeroes) indeed leaks info how much space is used, but it may be fine for many use cases. For example you may have a "browsing-work" VM with just firefox - don't leak much when it is known that there is a firefox profile, but you want to protect its content like bookmarks, browsing history, cookies etc.

> Then of course referencing one thin volume from another to remain "thin" doesn't work anymore (I think that's what the "hack" does, isn't it?) because the former may still be encrypted.

Yes, such approach indeed wouldn't work for TemplateVM's root volume of DisposableTemplateVM's private volume. Assuming of course each VM would get a different encryption key.

>  Once again this indicates that encryption per Qubes OS pool is more natural than encryption per Qubes volume.

Yes, but then you loose free disk space sharing between those pools. You need to allocate their sizes statically.

> However if you still want to support the per-volume idea, I guess we'd require `post_volume_import` & `post_volume_create` 

Yes, I think so.

qubes/storage/callback.py
3hhh: I wanted to fix this issue:
https://github.com/QubesOS/qubes-core-admin/pull/354/commits/287a4a0429165f9fa969960dafbd810a29c421d9#diff-271f0ec3445d2fb476cb711090e750b8L50-L53

I.e. I wanted to make my async code block qubesd as the sync code would do it / imitate the behaviour in order not to break any assumptions made by sync code authors.

As you pointed out an `asyncio` lock cannot achieve that though (i.e. that commit was useless regardless of the other mistakes).

I'm not sure whether Qubes OS runs async code in another process or whether sync & async code share the same.

In the former case (one common process for sync & async) this could only be fixed by blocking the Qubes OS sync main loop entirely during execution / run exclusively. Is there such a "global lock"?

In the latter case (one process each) there's probably no issue at all: Sync code cannot run in parallel to async code and sync code will not be interrupted even if run inside a coroutine as there's no awaitable afaik, i.e. the "word of caution" could be removed.

Please let me know what you think or whether all of this was just plain wrong...

3hhh:
> In the latter case (one process each) there's probably no issue at all: Sync code cannot run in parallel to async code and sync code will not be interrupted even if run inside a coroutine as there's no awaitable afaik, i.e. the "word of caution" could be removed.

I simply pushed https://github.com/QubesOS/qubes-core-admin/pull/354/commits/b95339ea273b8bea0e725bafbb919fef1fd4c7b8 as I think this was likely the right answer.

That would be it then.

BR
David

3hhh:
> That would be it then.
> 
> BR
> David

Bump.

dylangerdaly:
Is it possible for me to merge these commits into my local build and test?

I'd like to test this.

3hhh:
> Is it possible for me to merge these commits into my local build and test?
> 
> I'd like to test this.

It looks like the APIs didn't change so far in master [1], i.e. it should work.

However the current master is a 4.1 branch, so you'd have to use the currently unstable 4.1.

[1] https://github.com/3hhh/qubes-core-admin/compare/master...QubesOS:master

3hhh:
@dylangerdaly Did you manage to test it? Any findings?

@marmarek Any chance to see this in 4.1?

marmarek (approved):
This is all fine now for various highly customized configs or prototyping. But since it requires direct access to dom0 to configure it (both pool config itself, and possibly scripts called by it), it won't work with locked down dom0 (like with GUI domain). And if you configure it, then secrecy of config_id isn't really viable protection - if one can execute `admin.pool.Add`, then can probably also `admin.pool.Info` - which will reveal config_id. 
Anyway it is still useful tool for power users.

marmarek:
As for merging, this should go through openQA first to make sure we don't have any unwanted side effects (although very little common code is touched). But there is quite a lot PRs in openQA processing that needs to be (fixed and) merged first.

3hhh:
Alright, just let me know when I have to do anything (openQA probably cannot be triggered by me I guess).

marmarek:
> [](https://openqa.qubes-os.org/tests/13450#dependencies)
> ## Failed tests
> 
> No failures!

Nope. Everything failed, including storing list of failed tests... But unrelated to this PR.

3hhh:
> locked down dom0 (like with GUI domain)

Side note about that one:

Once you lock down dom0 entirely for users and admins, it becomes irrelevant to users, admins _and_ attackers. Since you likely want to retain the former Qubes OS functionality, it'll be provided to the users, admins and attackers inside the GUI domain. So that becomes the new dom0 then.

So in the end you'll either split functionality between the GUI & dom0 domain and thus not lock down dom0 entirely or have a new dom0 with a different name. ;-)

marmarek:
There are different use cases: https://www.qubes-os.org/news/2020/03/18/gui-domain/#why-make-a-gui-domain-at-all
So, yes, in a case when local user of GUI domain is the same as system admin, they by definition can change anything, including harmful actions. But with GUI domain it is realistic to split those roles.

marmarek:
> * StorageLVM: [test_000_volatile](https://openqa.qubes-os.org/tests/13819#step/StorageLVM/1)
> 
> * StorageLVM: [test_001_non_volatile](https://openqa.qubes-os.org/tests/13819#step/StorageLVM/2)
> 
> * StorageLVM: [test_002_read_only](https://openqa.qubes-os.org/tests/13819#step/StorageLVM/3)
> 
> * StorageLVM: [test_003_snapshot](https://openqa.qubes-os.org/tests/13819#step/StorageLVM/4)
> 
> * StorageLVM: [test_004_snapshot_non_persistent](https://openqa.qubes-os.org/tests/13819#step/StorageLVM/5)
> 
> * StorageLVM: [test_005_size_after_clone](https://openqa.qubes-os.org/tests/13819#step/StorageLVM/6)

Those failures seems to be related to this PR

3hhh:
> > * StorageLVM: [test_000_volatile](https://openqa.qubes-os.org/tests/13819#step/StorageLVM/1)
> > * StorageLVM: [test_001_non_volatile](https://openqa.qubes-os.org/tests/13819#step/StorageLVM/2)
> > * StorageLVM: [test_002_read_only](https://openqa.qubes-os.org/tests/13819#step/StorageLVM/3)
> > * StorageLVM: [test_003_snapshot](https://openqa.qubes-os.org/tests/13819#step/StorageLVM/4)
> > * StorageLVM: [test_004_snapshot_non_persistent](https://openqa.qubes-os.org/tests/13819#step/StorageLVM/5)
> > * StorageLVM: [test_005_size_after_clone](https://openqa.qubes-os.org/tests/13819#step/StorageLVM/6)
> 
> Those failures seems to be related to this PR

Yes, apparently I had overlooked the integration test references to `POOL_CONF`.
Anyway the latest commit re-introduced the global variable and should make the tests work as before.

(I guess the CI failures are related to the provider switch).

marmarek:
PipelineRetry

fepitre:
PipelineRetry

fepitre:
PipelineRetry

fepitre:
PipelineRetry

fepitre:
PipelineRetry

fepitre:
PipelineRetry

marmarek:
PipelineRefresh

marmarek:
PipelineRetry

fepitre:
PipelineRetry

PR #355, opened 2020-07-05
Removed unused vm.icon_path property

marmarta:
The property was not used for anything, but caused numerous
problems due to symlinks.

fixes QubesOS/qubes-issues#5934

### DIFF
--- a/qubes/tests/integ/backupcompatibility.py
+++ b/qubes/tests/integ/backupcompatibility.py
@@ -184,16 +184,12 @@ class TC_00_BackupCompatibility(
         os.mkdir(self.fullpath("appvms/test-work"))
         self.create_whitelisted_appmenus(self.fullpath(
             "appvms/test-work/whitelisted-appmenus.list"))
-        os.symlink("/usr/share/qubes/icons/green.png",
-                   self.fullpath("appvms/test-work/icon.png"))
         self.create_private_img(self.fullpath("appvms/test-work/private.img"))
 
         # StandaloneVM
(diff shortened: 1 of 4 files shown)


qubes/tests/integ/backupcompatibility.py
```diff
@@ -108,7 +108,6 @@
 Exec=qvm-run -q --tray -a %VMNAME% '{command} %u'
 Categories=Network;WebBrowser;
 X-Qubes-VmName=%VMNAME%
-Icon=%VMDIR%/icon.png
```
marmarek, line 111: This should stay, because this is how .desktop files looked at R2.0 times.
But I think that isn't a problem for removing the icon. Perhaps the restore code should adjust `Icon=` entry appropriately (but that's out of scope for this PR).

qubes/vm/qubesvm.py
```diff
@@ -954,19 +950,8 @@ def on_domain_init_loaded(self, event):
...
-                pass
-            if hasattr(os, "symlink"):
-                os.symlink(newvalue.icon_path, self.icon_path)
-                subprocess.call(['sudo', 'xdg-icon-resource', 'forceupdate'])
-            else:
-                shutil.copy(newvalue.icon_path, self.icon_path)
-
         # icon is calculated based on label
-        self.fire_event('property-reset:icon', name='icon')
+        self.fire_event('property-set:icon', name='icon')
```
marmarek, line 954: This looks like an unintended change.

PR #356, opened 2020-07-06
Make Volume.export async, and few other changes

marmarek:
- async `Volume.export`
- add `Volume.export_end`
- DispVM volatile volume improvement
- various fixes

Fixes QubesOS/qubes-issues#5935
Fixes QubesOS/qubes-issues#5933


3hhh:
Thanks for picking this up so quickly! :-)

I just noticed the missing doc hint for `Volume.export()` ("can be implemented as coroutine") and the unexpected coroutine annotation of `Volume.export_end()`, which none of the other declarations have (I guess because it's optional).

The rest looked reasonable, but then I'm certainly no experienced eye reviewer.

PR #369, opened 2020-09-22
ext: support for non-service feature advertisement

marmarek:
Add an API for VMs to announce support for non-service features. This is
very similar to supported-service.* features, but applies to non-service
features. This may be also used for announcing support for features that
do not use qvm-features framework itself - for example some VM kernel
features, installed drivers, packages etc.

QubesOS/qubes-issues#6030


marmarek:
@marmarta here is the backend support for `supported-feature.*` features.

PR #371, opened 2020-10-30
tests: fail the test early if VM fails to start

marmarek:
Make an exception in vm.start() actually interrupt the test. The
asyncio.wait() returns list of completed tasks, where exception may be
stored - but is not raised directly. Change to asyncio.gather() that will
propagate the exception by default.
As a side effect, avoid deprecated direct coroutine passing to
asyncio.wait(). This functionality in asyncio.gather() is not
deprecated.


marmarek:
PipelineRetry

PR #374, opened 2020-11-20
Implement changing the template of a DispVM

DemiMarie:
Fixes https://github.com/QubesOS/qubes-issues/issues/5491
Fixes https://github.com/QubesOS/qubes-issues/issues/3576

Please excuse any mistakes ― I made this PR after 1AM :smile:.

### DIFF
--- a/qubes/tests/vm/dispvm.py
+++ b/qubes/tests/vm/dispvm.py
@@ -113,19 +113,19 @@ class TC_00_DispVM(qubes.tests.QubesTestCase):
                 'get_new_unused_dispid': mock.Mock(return_value=42),
                 '__getitem__.side_effect': orig_getitem
             })
-            dispvm = self.app.add_new_vm(qubes.vm.dispvm.DispVM,
+            self.dispvm = self.app.add_new_vm(qubes.vm.dispvm.DispVM,
                 name='test-dispvm', template=self.appvm)
 
-            dispvm.template = self.appvm
-            dispvm.start()
+            self.dispvm.template = self.appvm
+            self.dispvm.start()
             if not self.app.vmm.offline_mode:
                 assert not dispvm.is_halted()
                 with self.assertRaises(qubes.exc.QubesVMNotHaltedError):
-                    dispvm.template = self.appvm
+                    self.dispvm.template = self.appvm
             with self.assertRaises(qubes.exc.QubesValueError):
-                dispvm.template = qubes.property.DEFAULT
-            dispvm.kill()
-            dispvm.template = self.appvm
+                self.dispvm.template = qubes.property.DEFAULT
+            self.dispvm.kill()
+            self.dispvm.template = self.appvm
 
     def test_003_dvmtemplate_template_change(self):
         self.appvm.template_for_dispvms = True
@@ -141,8 +141,7 @@ class TC_00_DispVM(qubes.tests.QubesTestCase):
             self.dispvm = self.app.add_new_vm(qubes.vm.dispvm.DispVM,
                 name='test-dispvm', template=self.appvm)
 
-            with self.assertRaises(qubes.exc.QubesVMNotHaltedError):
-                self.appvm.template = self.template
+            self.appvm.template = self.template
             with self.assertRaises(qubes.exc.QubesValueError):
                 self.appvm.template = qubes.property.DEFAULT
 
@@ -172,9 +171,10 @@ class TC_00_DispVM(qubes.tests.QubesTestCase):
                 'get_new_unused_dispid': mock.Mock(return_value=42),
                 '__getitem__.side_effect': orig_getitem
             })
-            dispvm = self.app.add_new_vm(qubes.vm.dispvm.DispVM,
+            self.dispvm = self.app.add_new_vm(qubes.vm.dispvm.DispVM,
                 name='test-dispvm', template=self.appvm)
             mock_domains.get_new_unused_dispid.assert_called_once_with()
+        dispvm=self.dispvm
         self.assertEqual(dispvm.name, 'test-dispvm')
         self.assertEqual(dispvm.template, self.appvm)
         self.assertEqual(dispvm.label, self.appvm.label)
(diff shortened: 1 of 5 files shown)


qubes/vm/mix/dvmtemplate.py
```diff
@@ -45,14 +45,12 @@ def __on_pre_del_dvmtemplate(self, event, name,
...
+    @qubes.events.handler('property-set:template')
+    def __on_property_set_template(self, event, name, newvalue,
             oldvalue=None):
         # pylint: disable=unused-argument
-        if any(self.dispvms):
-            raise qubes.exc.QubesVMInUseError(self,
-                'Cannot change template '
-                'while there are DispVMs based on this qube')
+        for vm in self.dispvms:
+            vm.on_property_set_template(event, name, newvalue, oldvalue)
```
marmarek, line 53: I think you need also a pre-set handler to check if there is no running DispVM. Perhaps DispVMs with `auto_cleanup=True` could be excluded from this check (as there isn't a risk of starting it again with not switched template), but not fully sure about that - qvm-prefs will tell lies for such not updated DispVMs.

In the end, we want to have some queued template change feature: https://github.com/QubesOS/qubes-issues/issues/5697

marmarek (changes requested):
Conceptually this should be fine. There is one code duplication issue that can be easily solved, but take a look also at tests output: https://gitlab.com/QubesOS/qubes-core-admin/-/jobs/862415892
The pylint job is not yet fixed after migration to gitlab, so ignore that failure for now.

qubes/vm/mix/dvmtemplate.py
DemiMarie: What happens if there is no pre-set handler, and `on_property_set_template` throws?

marmarek: Property remains set, other handlers may not be executed.

marmarek:
PipelineRetry

marmarek:
Tests fail:
```
ERROR: qubes.tests.vm.dispvm/TC_00_DispVM/test_003_dvmtemplate_template_change
  File "/home/gitlab-runner/builds/QubesOS/qubes-core-admin/qubes/tests/vm/dispvm.py", line 139, in test_003_dvmtemplate_template_change
    self.appvm.template = self.template
  File "/home/gitlab-runner/builds/QubesOS/qubes-core-admin/qubes/vm/__init__.py", line 492, in __set__
    super().__set__(instance, vm)
  File "/home/gitlab-runner/builds/QubesOS/qubes-core-admin/qubes/__init__.py", line 259, in __set__
    instance.fire_event('property-pre-set:' + self.__name__,
  File "/home/gitlab-runner/builds/QubesOS/qubes-core-admin/qubes/events.py", line 197, in fire_event
    sync_effects, async_effects = self._fire_event(event, kwargs,
  File "/home/gitlab-runner/builds/QubesOS/qubes-core-admin/qubes/events.py", line 166, in _fire_event
  File "/home/gitlab-runner/builds/QubesOS/qubes-core-admin/qubes/vm/mix/dvmtemplate.py", line 54, in __on_pre_property_set_template
    raise qubes.exc.QubesVMNotHaltedError(self,
qubes.exc.QubesVMNotHaltedError: Cannot change template while there are running DispVMsbased on this DVM template
FAIL: qubes.tests.vm.dispvm/TC_00_DispVM/test_002_template_change
  File "/home/gitlab-runner/builds/QubesOS/qubes-core-admin/qubes/tests/vm/dispvm.py", line 120, in test_002_template_change
    dispvm.template = self.appvm
AssertionError: QubesValueError not raised
```

DemiMarie:
Fixed, along with a test that changing the template works if the dispvm is stopped.

marmarek:
It still fails (see "Details" link)

qubes/vm/appvm.py
```diff
@@ -77,7 +85,7 @@ def __init__(self, app, xml, **kwargs):
                     qubes.property.bool(None, None, node_dispvm_allowed.text)
                 node_dispvm_allowed.getparent().remove(node_dispvm_allowed)
 
-        self.volume_config = copy.deepcopy(self.default_volume_config)
+        self.volume_config = copy.deepcopy(AppVM.default_volume_config)
```
marmarek, line 88: This looks wrong - will break in a surprising way if there would be a class inheriting from AppVM that uses different default_volume_config.

qubes/vm/appvm.py
```diff
@@ -120,10 +128,4 @@ def on_property_set_template(self, event, name, newvalue, oldvalue=None):
...
         on template change.
         '''  # pylint: disable=unused-argument
-
-        for volume_name, conf in self.default_volume_config.items():
-            if conf.get('snap_on_start', False) and \
-                    conf.get('source', None) is None:
-                config = conf.copy()
-                self.volume_config[volume_name] = config
-                self.storage.init_volume(volume_name, config)
+        template_changed_update_storage(self, AppVM.default_volume_config)
```
marmarek, line 131: Same here, should be `self.default_volume_config`.

run-tests
```diff
@@ -1,5 +1,7 @@
-#!/bin/sh
+#!/bin/sh --
 
+set -eu
+sudo dnf -y install lvm2 python3-inotify python3-sphinx python3-docutils python3-PyYAML python3-jinja2 python3-lxml python3-pylint python3-coverage
```
marmarek, line 4: Can we add it to README instead, or make it conditional in some way? `dnf` likes to refresh metadata which takes time and having that every time in the middle of debugging will be annoying.

qubes/tests/vm/dispvm.py
```diff
@@ -116,10 +116,16 @@ def test_002_template_change(self):
...
+            dispvm.template = self.appvm
+            dispvm.start()
+            if not self.app.vmm.offline_mode:
+                assert not dispvm.is_halted()
+                with self.assertRaises(qubes.exc.QubesVMNotHaltedError):
+                    dispvm.template = self.appvm
             with self.assertRaises(qubes.exc.QubesValueError):
                 dispvm.template = qubes.property.DEFAULT
+            dispvm.kill()
+            dispvm.template = self.appvm
```
marmarek, line 128: Since the whole thing was blocked because it caused volumes not being updated, can you add a test that verifies it? I think we need two cases:
1. 2x TemplateVM + 2x AppVM + DispVM, then change template of the DispVM and verify if its root and private volumes were updated.
2. 2x TemplateVM + AppVM + DispVM, then change template of the AppVM and verify if DispVM's root volume was updated.

There are already similar tests for AppVM, you can copy from there.

marmarek (changes requested):

run-tests
DemiMarie: Fixed by checking with `rpm -q --quiet` first.

run-tests
```diff
@@ -1,7 +1,16 @@
...
+#!/bin/bash --
 
+set -eu
+unset applications do_install_deps CLEANUP_LVM name DEFAULT_LVM_POOL retcode
+applications=(lvm2 python3-inotify python3-sphinx python3-docutils
+    python3-pyyaml python3-jinja2 python3-lxml python3-pylint python3-coverage
+    btrfs-progs vim-common)
+do_install_deps=false
+rpm -q --quiet -- "${applications[@]}" ||
+sudo dnf -- install "${applications[@]}"
```
marmarek, line 10: I'd disarm it even further, to not break when running on Debian...

DemiMarie: Fixed by checking that `rpm` and `dnf` are installed in the standard search path before using them.

marmarek (approved):

PR #375, opened 2020-11-20
Always snapshot in the FILE pool

DemiMarie:
We must snapshot a VM’s disk before exporting it.  Otherwise, we will
likely corrupt the VM’s filesystem.

Fixes https://github.com/QubesOS/qubes-issues/issues/4324

### DIFF
--- a/qubes/storage/file.py
+++ b/qubes/storage/file.py
@@ -183,7 +183,7 @@ class FileVolume(qubes.storage.Volume):
         self.dir_path = dir_path
         assert self.dir_path, "dir_path not specified"
         self._revisions_to_keep = 0
-        self._locked = None
+        self._export_lock = None
         super().__init__(**kwargs)
 
         if self.snap_on_start:
@@ -268,17 +268,17 @@ class FileVolume(qubes.storage.Volume):
         return self
 
     def export(self):
-        if self._locked is not None:
-            assert self._locked is FileVolume._marker_running, \
+        if self._export_lock is not None:
+            assert self._export_lock is FileVolume._marker_running, \
                 'nested calls to export()'
             self._not_implemented('exporting a dirty volume')
-        self._locked = FileVolume._marker_exported
+        self._export_lock = FileVolume._marker_exported
         return self.path
 
     def export_end(self, path):
-        assert self._locked is FileVolume._marker_exported, \
-            'cannot end an export that never began'
-        self._locked = None
+        assert self._export_lock is not FileVolume._marker_running, \
+            'ending an export on a running volume?'
+        self._export_lock = None
 
     @asyncio.coroutine
     def import_volume(self, src_volume):
@@ -322,13 +322,13 @@ class FileVolume(qubes.storage.Volume):
         return self
 
     def start(self):
-        if self._locked is not None:
-            assert self._locked is FileVolume._marker_exported, \
+        if self._export_lock is not None:
+            assert self._export_lock is FileVolume._marker_exported, \
                 'nested calls to start()'
             self._not_implemented('starting a VM with an exported volume')
         if self.is_dirty():
             self._not_implemented('exporting a dirty volume')
-        self._locked = FileVolume._marker_running
+        self._export_lock = FileVolume._marker_running
         if not self.save_on_stop and not self.snap_on_start:
             self.reset()
         else:
@@ -346,15 +346,15 @@ class FileVolume(qubes.storage.Volume):
         return self
 
     def stop(self):
-        assert self._locked is FileVolume._marker_running, \
-            'cannot stop a volume that has not been started'
+        assert self._export_lock is not FileVolume._marker_exported, \
+            'trying to stop exported file volume?'
         if self.save_on_stop:
             self.commit()
         elif self.snap_on_start:
             _remove_if_exists(self.path_cow)
         else:
             _remove_if_exists(self.path)
-        self._locked = None
+        self._export_lock = None
         return self
 
     @property
--- a/qubes/storage/file.py
+++ b/qubes/storage/file.py
@@ -265,9 +265,7 @@ class FileVolume(qubes.storage.Volume):
         return self
 
     def export(self):
-        # FIXME: this should rather return snapshot(self.path, self.path_cow)
-        #  if domain is running
-        return self.path
+        return snapshot(self.path, self.path_cow)
 
     @asyncio.coroutine
     def import_volume(self, src_volume):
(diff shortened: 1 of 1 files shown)


qubes/storage/file.py
```diff
@@ -265,9 +265,7 @@ def commit(self):
         return self
 
     def export(self):
-        # FIXME: this should rather return snapshot(self.path, self.path_cow)
-        #  if domain is running
-        return self.path
+        return snapshot(self.path, self.path_cow)
```
marmarek, line 268: The `snapshot` is not a real function. It's a pseudo-code for assembling dm-snapshot target (or rather - dm-snapshot-origin). There are two difficulties with this:
1. The actual device mapper is assembled in in `block-origin` script, which is called via libvirt -> libxl. It (in its current shape) can't be called outside because for example it parameters in xenstore.
2. Such device mapper needs to be cleaned up after use. Until recently, we didn't have API to cleanup device returned from `export` function (now we have it: `export_end`).

Doing this properly requires quite a bit of refactoring - at least around `block-snapshot`/`block-origin` scripts. They should really be a python code in this very file, instead of external shell scripts and libvirt should get already assembled device. But that's quite a bit of work that may not be worth for a storage driver we are going to remove (at least that was initial plan).
Do you see any use case that would warrant actively maintaining this driver? If so, it may be worth doing this refactor. If not, I'd rather prefer a band-aid blocking export on a running VM. If I'm not mistaken, export is used in backup code only.

marmarek (changes requested):

qubes/storage/file.py
DemiMarie: I honestly cannot think of any reason to keep maintaining it, and I suggest formally deprecating it.  The main reasons other virtualization tools support files are portability and not requiring root privileges, neither of which apply here.

rustybird:
Backup calls `export()` for all VMs at the very beginning, and then the backup can run for hours. One of those VMs might start before it's finished. :\

What do you think about adding an "export lock" to the FileVolume object, which `export()` or `start()` would attempt to acquire (non-blocking), and `export_end()` or `stop()` would release?

The `is_dirty()` check could still be kept as a fail-safe in case `qubesd` was restarted after `start()`. (In the opposite case - restarting `qubesd` after `export()` - the backup is aborted anyway.)

qubes/storage/file.py
```diff
@@ -328,12 +344,15 @@ def start(self):
         return self
 
     def stop(self):
+        assert self._locked is FileVolume._marker_running, \
+            'cannot stop a volume that has not been started'
```
rustybird, line 350: This breaks if a VM is started, then `qubesd` is restarted (recreating a FileVolume object with `_locked == None`), then the VM is stopped.

Similar for `export_end()`, I suppose.

qubes/storage/file.py
```diff
@@ -176,11 +176,14 @@ def included_in(self, app):
...
     ''' Parent class for the xen volumes implementation which expects a
         `target_dir` param on initialization.  '''
+    _marker_running = object()
+    _marker_exported = object()
 
     def __init__(self, dir_path, **kwargs):
         self.dir_path = dir_path
         assert self.dir_path, "dir_path not specified"
         self._revisions_to_keep = 0
+        self._locked = None
```
rustybird, line 186: `_export_locked`, to distinguish it from the Volume base class's `_lock`?

qubes/storage/file.py
DemiMarie: Is `is_dirty` guaranteed to be `True` in that situation?

rustybird: Hmm maybe not quite guaranteed, e.g. if `qubesd` is restarted while Xen is still running the `block-origin` script (so that `st_blocks` is not yet `> 0`).

marmarek: In that case, VM wouldn't fully start - at best would remain "paused" (it's only unpaused by qubesd after fully constructing - if interrupted here, it won't unpause). If qubesd gets restarted at this stage, I don't think such VM will boot properly, even if gets unpaused manually later, for example qubesdb won't be there.
For all practical purposes, I think it is ok to assume `is_dirty` is true when VM is running.

qubes/storage/file.py
```diff
@@ -326,6 +326,8 @@ def start(self):
             assert self._locked is FileVolume._marker_exported, \
                 'nested calls to start()'
             self._not_implemented('starting a VM with an exported volume')
+        if self.is_dirty():
+            self._not_implemented('exporting a dirty volume')
```
rustybird, line 330: This chunk slipped into `start()` instead of `export()`.

qubes/storage/file.py
```diff
@@ -265,10 +268,18 @@ def commit(self):
         return self
 
     def export(self):
-        if self.is_dirty():
+        if self._locked is not None:
+            assert self._locked is FileVolume._marker_running, \
+                'nested calls to export()'
             self._not_implemented('exporting a dirty volume')
```
rustybird, line 274: `'exporting a started volume'`

rustybird:
`_not_implemented()` only creates the exception objects, they still need `raise`.

qubes/storage/file.py
```diff
@@ -271,6 +271,8 @@ def export(self):
         if self._export_lock is not None:
             assert self._export_lock is FileVolume._marker_running, \
                 'nested calls to export()'
+            self._not_implemented('exporting a starting volume')
```
rustybird, line 274: Nitpick: start**ed**

rustybird:
> `_not_implemented()` only creates the exception objects, they still need `raise`.

Actually, it might be better to raise `NotImplementedError` directly. The `_not_implemented()` helper expects the passed string to be a method name and adds parentheses at the end, so the generated message looks odd if you pass a freeform string:

```
Volume FileVolume has exporting a dirty volume() not implemented
                                              ^^
```

rustybird:
Or, even better (sorry for the noise): A `StoragePoolException` - those are surfaced to the user as a `qvm-backup` error message, whereas with `NotImplementedError` there's only a generic `Got empty response from qubesd` error message and the real information goes to syslog.

rustybird (approved):
Tested on R4.1. Both CLI and GUI for backup/clone/start work fine!

rustybird:
I just noticed a small bug in Qube Manager: If a VM fails to start, it's handled gracefully with [an error message](https://github.com/QubesOS/qubes-manager/blob/1b0f96a7563bee398d698fbe0bb16c9b809a0752/qubesmanager/qube_manager.py#L580) but the VM is stuck in starting state (yellow). @marmarta

marmarek:
PipelineRetry

DemiMarie:
@marmarek can you merge this?

marmarek (approved):

PR #382, opened 2021-01-15
Do not activate unused LVM volumes

marmarek:
This prevents udev from accessing not currently used volumes, and also 
makes LVM snapshot operations faster (less locking to do).

There is at least one unsolved problem here: inactive volumes do not have
size listed on `lvs` output. There is a band-aid by caching the usage from
the last time the volume was active, but it isn't enough for many cases (at
the very least - reporting disk usage of a VM that wasn't started
recently).

There may be also other places that assume existing volumes always have 
relevant /dev/ node present.

Fixes QubesOS/qubes-issues#6184

### DIFF
--- a/qubes/storage/lvm.py
+++ b/qubes/storage/lvm.py
@@ -223,16 +223,21 @@ def _parse_lvm_cache(lvm_output):
         line = line.decode().strip()
         pool_name, pool_lv, name, size, usage_percent, attr, \
             origin, metadata_size, metadata_percent = line.split(';', 8)
-        if '' in [pool_name, name, size, usage_percent]:
+        if '' in [pool_name, name, size]:
             continue
         name = pool_name + "/" + name
         size = int(size[:-1])  # Remove 'B' suffix
-        usage = int(size / 100 * float(usage_percent))
+        if usage_percent:
+            usage = int(size / 100 * float(usage_percent))
+        else:
+            # keep cached usage for non-active volumes
+            usage = size_cache.get(name, {}).get('usage')
         if metadata_size:
             metadata_size = int(metadata_size[:-1])
+        if metadata_percent:
             metadata_usage = int(metadata_size / 100 * float(metadata_percent))
         else:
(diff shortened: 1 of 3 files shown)


DemiMarie:
Is it possible to avoid using the LVM2 CLI and instead invoke a library and/or kernel interface?

marmarek:
Last time I checked, python bindings to LVM were very limited (including many missing pieces for thin pools). As for using kernel interface directly - I definitely don't want to re-implement myself all the metadata processing, locking, udev synchronization etc. In any case, see https://github.com/QubesOS/qubes-issues/issues/6184#issuecomment-760565957, likely I won't spent any more time on this PR right now.

DemiMarie:
> Last time I checked, python bindings to LVM were very limited (including many missing pieces for thin pools). As for using kernel interface directly - I definitely don't want to re-implement myself all the metadata processing, locking, udev synchronization etc. In any case, see [QubesOS/qubes-issues#6184 (comment)](https://github.com/QubesOS/qubes-issues/issues/6184#issuecomment-760565957), likely I won't spent any more time on this PR right now.

Turns out that the Python bindings for LVM2 (and indeed the LVM2 C API) are deprecated!  The recommended replacement is the D-Bus API.

DemiMarie:
> > Last time I checked, python bindings to LVM were very limited (including many missing pieces for thin pools). As for using kernel interface directly - I definitely don't want to re-implement myself all the metadata processing, locking, udev synchronization etc. In any case, see [QubesOS/qubes-issues#6184 (comment)](https://github.com/QubesOS/qubes-issues/issues/6184#issuecomment-760565957), likely I won't spent any more time on this PR right now.
> 
> Turns out that the Python bindings for LVM2 (and indeed the LVM2 C API) are deprecated! The recommended replacement is the D-Bus API.

To elaborate: since LVM2 needs to support modifying the device on which it itself is located, it needs to lock all of its memory into RAM at various points.  Otherwise, it could deadlock.

PR #384, opened 2021-01-28
storage/file: fix is_dirty() false positive

rustybird:
Marking this as a draft in case someone has an idea how to fix the underlying problem instead, which is the false positive in FileVolume's `is_dirty()` (see commit message). @DemiMarie

### DIFF
--- a/linux/system-config/block-snapshot
+++ b/linux/system-config/block-snapshot
@@ -4,6 +4,8 @@
 #
 # This creates dm-snapshot device on given arguments
 
+SNAPSHOT_CHUNKSIZE=256  # same as in file.py
+
 dir=$(dirname "$0")
 if [ "$1" = "prepare" ] || [ "$1" = "cleanup" ]; then
   # shellcheck disable=SC1090,SC1091
@@ -80,7 +82,7 @@ create_dm_snapshot() {
     base_dev=$(get_dev "$base")
     cow_dev=$(get_dev "$cow")
     base_sz=$(blockdev --getsz "$base_dev")
-    do_or_die dmsetup create "$dm_devname" --table "0 $base_sz snapshot $base_dev $cow_dev P 256"
+    do_or_die dmsetup create "$dm_devname" --table "0 $base_sz snapshot $base_dev $cow_dev P $SNAPSHOT_CHUNKSIZE"
   fi
 
 }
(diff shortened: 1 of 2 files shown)


DemiMarie:
Will this lead to corruption?

marmarek:
I think instead of removing the check, the export lock should be removed. So, exporting volume of a running VM would fail (IMO more important case), but starting a VM while its being exported would not be prevented.

rustybird:
@DemiMarie:

> Will this lead to corruption?

If this PR removing the buggy fallback check is merged, then there's nothing preventing a started volume from being exported after qubesd is restarted. So e.g. a backup after a qubes-core-dom0 package update could be corrupted. But _with_ the buggy fallback check, 'file' pool users typically won't be able to back up some of their (stopped) templates at all.

@marmarek:
> In fact, just removing this check does not solve the issue - the volume will remain locked after export.

Are you referring to the [R4.0 backport bug](https://github.com/QubesOS/qubes-core-admin/commit/0eb95044dd937857581a22c13a692eff5d92c70b#r46447802)? That's unrelated to this PR.

marmarek:
> Are you referring to the [R4.0 backport bug](https://github.com/QubesOS/qubes-core-admin/commit/0eb95044dd937857581a22c13a692eff5d92c70b#r46447802)? That's unrelated to this PR.

Ah, yes.

marmarek:
What about extending is_dirty  check to look into device mapper status to check if there are actually any uncomitted changes? or Maybe just slightly smarter file size check would be enough (if there is some file size that still says no blocks are written)?

rustybird:
`dmsetup status` shows the same false positive as `stat()`ing the `-cow.img` file: 512 blocks used, that's one [256 KiB chunk](https://github.com/QubesOS/qubes-core-admin/blob/v4.1.18/linux/system-config/block-snapshot#L83).

I guess it's possible for this one chunk to be empty except for the header (= not really dirty) or to also contain some other data (= legitimately dirty), so here's a very ugly commit to distinguish these cases...

rustybird:
The fixup commit hardcodes the EMPTY_SNAPSHOT byte sequence, with a comment explaining how to generate it.

If it looks lightly less disgusting to you too, I'll squash it onto the previous commit.

marmarek:
Yes, I think precomputed version is slightly better, it doesn't introduce several mostly unused variables, while still enough to understand its origin.

Thanks!

rustybird:
Okay, squashed.

rustybird:
And force-pushed again with "Fixes ..." line in the commit message.

marmarek (approved):

PR #386, opened 2021-02-04
Don't log to qubes.log or vm-*.log files on disk, only stderr

rustybird:


rustybird:
Still need to think about how to deal with `/var/log/qubes/vm-*.log`, which need logrotate and thus a qubesd restart.

marmarek:
What about ensuring log messages in journalctl include originating VM name, and then drop logging to files? On the other hand, logging to files (instead of journalctl) makes it easier to clean logs on DispVM destroy (see https://github.com/QubesOS/qubes-issues/issues/4972).
As for qubesd restart, this should be easy to avoid with [`logging.handlers.WatchedFileHandler`](https://stackoverflow.com/a/28333560)

rustybird:
`WatchedFileHandler` ftw! Nice.

Oh man, anti-forensic logging is a can of worms. I've backed out of removing `/var/log/qubes/qubes.log` for now: Given that VM-related messages _also_ go to the main qubesd log, the user might want to override `qubesd.service` so that stdout/stderr is sent to `/dev/null` - because `/var/log/qubes/qubes.log` is easier to purge from storage.

rustybird:
> the user might want to override `qubesd.service` so that stdout/stderr is sent to `/dev/null` - because `/var/log/qubes/qubes.log` is easier to purge from storage

On second thought, that's a bad idea, because it would also throw away stdout/stderr of `qubesd.service` subprocesses, which are normally captured in the journal but not the `qubes.log` file.

@marmarek:
> What about ensuring log messages in journalctl include originating VM name, and then drop logging to files?

Done.

> On the other hand, logging to files (instead of journalctl) makes it easier to clean logs on DispVM destroy

Currently it doesn't seem to make much of a difference, because plenty of VM related log messages don't even use the VM logger - e.g. storage drivers printing a volume ID. There are also messages involving multiple VMs - stuff like `Copying vm1/root.img to vm2/root.img`.

marmarek (approved):

PR #396, opened 2021-03-20
Encrypt volatile volumes

DemiMarie:
This uses cryptsetup to handle the encryption.  cryptsetup transparently creates loop devices where needed.


marmarek (changes requested):
There are a couple design-level remarks:
1. The `Storage` class (as well as `Pool` and `Volume`) should handle volumes uniformly; if you need anywhere `if name == '<some-specific-value>'`, then something went wrong.
2. Similarly, changes in the `QubesVM` class have the same issue.
3. As noted below - the actual block device may not exist at this stage yet, if "block script" is in use. But in fact, it would be good to get rid of the block script (the thing that is called by libxl when starting the VM) and call whatever is needed in the storage driver here (*).

I can think of two alternative designs:
1. Create a wrapper storage pool that takes block device from the actual pool and wraps it with encryption layer. For now, it would work just for volatile volume (save_on_stop=False, snap_on_start=False), but it could be trivially extended later for a persistent encryption too (#1293). We have already some storage pool driver that wraps another ("callback" driver), although this one could be way simpler.
2. Add some property to the Volume config (`ephemeral`?) and let the storage drivers implement it how they like. To make it reasonably robust, it would need:
   - a helper function for simple case (take existing block device and wrap it with cryptsetup) - should be usable by most drivers with minimal boilerplate
   - a class property whether "ephemera" functionality is supported - and if not, refuse (in the base class) to create a volume with `ephemeral` property set - this way we avoid pitfalls of `ephemeral` property being no-op on some drivers

The first option feels a bit more flexible, but may be harder to handle on upper levels (like you'd probably need separate "volatile" pool for each actual pool, inspecting free space may also be not very obvious etc).

(*) This will allow to better parallelize the calls. But also, we could get rid of the libvirt patch that passes through the script parameter to libxl - this one was rejected upstream.

DemiMarie:
> There are a couple design-level remarks:
> 
> 1. The `Storage` class (as well as `Pool` and `Volume`) should handle volumes uniformly; if you need anywhere `if name == '<some-specific-value>'`, then something went wrong.

Fixed before I even read this :laughing:

> 2. Similarly, changes in the `QubesVM` class have the same issue.

Ditto

> 3. As noted below - the actual block device may not exist at this stage yet, if "block script" is in use. But in fact, it would be good to get rid of the block script (the thing that is called by libxl when starting the VM) and call whatever is needed in the storage driver here (*).

It looks like none of our current pools use block scripts for ephemeral volumes.  The only one that uses them at all is the file pool, and it explicitly returns `None` in that case.  Since block scripts are deprecated, I would be fine just asserting that ephemeral volumes do not use them.

> I can think of two alternative designs:
> 
> 1. Create a wrapper storage pool that takes block device from the actual pool and wraps it with encryption layer. For now, it would work just for volatile volume (save_on_stop=False, snap_on_start=False), but it could be trivially extended later for a persistent encryption too (#1293). We have already some storage pool driver that wraps another ("callback" driver), although this one could be way simpler.
> 2. Add some property to the Volume config (`ephemeral`?) and let the storage drivers implement it how they like. To make it reasonably robust, it would need:
>    
>    * a helper function for simple case (take existing block device and wrap it with cryptsetup) - should be usable by most drivers with minimal boilerplate
>    * a class property whether "ephemera" functionality is supported - and if not, refuse (in the base class) to create a volume with `ephemeral` property set - this way we avoid pitfalls of `ephemeral` property being no-op on some drivers
> 
> The first option feels a bit more flexible, but may be harder to handle on upper levels (like you'd probably need separate "volatile" pool for each actual pool, inspecting free space may also be not very obvious etc).

I agree that the first option is too complex.  Is a default implementation using cryptsetup (which Just Works with every existing driver) okay?

marmarek:
> I agree that the first option is too complex. Is a default implementation using cryptsetup (which Just Works with every existing driver) okay?

Mostly yes. It needs to handle the case where underlying "device" is in fact a file (btrfs pool). I think a cleaner option would be to require `volume.block_device().path` to really return a block device path (`loop` device in this case), instead of a file and letting (the xen default) block script creating the loop device later.

> Similarly, changes in the `QubesVM` class have the same issue.

You still calculate device path there (`vm.volatile_volume_path()`). IMO this whole modification should be contained within storage classes.

> Since block scripts are deprecated, I would be fine just asserting that ephemeral volumes do not use them.

+1

DemiMarie:
> > I agree that the first option is too complex. Is a default implementation using cryptsetup (which Just Works with every existing driver) okay?
> 
> Mostly yes. It needs to handle the case where underlying "device" is in fact a file (btrfs pool). I think a cleaner option would be to require `volume.block_device().path` to really return a block device path (`loop` device in this case), instead of letting (the xen default) block script creating the loop device later.

Not a problem, `cryptsetup` handles that automatically.  

> > Similarly, changes in the `QubesVM` class have the same issue.
> 
> You still calculate device path there (`vm.volatile_volume_path()`). IMO this whole modification should be contained within storage classes.

Got it.

DemiMarie:
> > Similarly, changes in the `QubesVM` class have the same issue.
> 
> You still calculate device path there (`vm.volatile_volume_path()`). IMO this whole modification should be contained within storage classes.

The problem is finding a name for the encrypted volume.  It needs to be unique *and* not conflict with any other devicemapper volumes.  I could use something like a SHA-256 hash of the volume path, but that is both ugly and user-hostile.  So my current approach is to compute the name based on the (qube name, volume name) pair, which is guaranteed to be unique across the entire system.  The result is an implementation that requires no per-driver boilerplate whatsoever, and provides friendly names to boot.

marmarek:
You can build similar name inside storage class, you don't need to have it in VM class. But also, the current algorithm will collide names if you create a VM with two encrypted volumes. You need to use both VM name and volume name.

DemiMarie:
> You can build similar name inside storage class, you don't need to have it in VM class. But also, the current algorithm will collide names if you create a VM with two encrypted volumes. You need to use both VM name and volume name.

I have some local changes that fix this, but forgot to push them.  Oops!

DemiMarie:
@marmarek would you mind re-reviewing?  My only worry is the lack of code coverage in the tests.

qubes/tools/qubesd.py
```diff
@@ -15,6 +15,22 @@
...
+    """
+    import ctypes
+    libc = ctypes.CDLL('libc.so.6', use_errno=True)
+    getentropy = libc['getentropy']
+    getentropy.restype = ctypes.c_int
+    getentropy.argtypes = (ctypes.c_void_p, ctypes.c_size_t)
+    buffer = ctypes.create_string_buffer(1)
+    if getentropy(buffer, 1):
+        err = ctypes.get_errno()
+        raise OSError(err, os.strerror(err))
```
rustybird, line 31:
```suggestion
    with open('/dev/random', 'rb') as random:
        random.read(1)
```

Or `os.getrandom(1)` for Python 3.6+.

You could even skip this whole commit if the `cryptsetup` invocation is changed to use `/dev/random`, assuming [kernel 5.6+](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=30c08efec8884fb106b8e57094baa51bb4c44e32).

DemiMarie: I can’t assume kernel 5.6+, sadly.  But I *can* assume Python 3.6+ in R4.1.

qubes/storage/__init__.py
```diff
@@ -158,6 +158,48 @@ def __xml__(self):
...
+        if self.script is not None:
+            self._not_implemented("starting an encrypted volume with a script")
+        await qubes.utils.coro_maybe(self.start())
+        await qubes.utils.cryptsetup(
+            '--key-file=/dev/urandom',
+            '--cipher=aes-xts-plain64',
+            '--type=plain',
+            '--',
+            'open',
+            self.path,
```
marmarek, line 187: `self.path` may be not the final device for the VM to use, you need to extract it from `self.block_device()` (after `self.start()`). Coincidentally for all the pools currently in core-admin, both paths are the same for volatile volumes (but for example they do differ in some other cases).

qubes/storage/__init__.py
```diff
@@ -339,6 +381,43 @@ def size(self, size):
...
+            is available to help construct the device path.
+        '''
+        assert device.domain is None, "Volatile volume must be in dom0"
+        assert device.script is None, \
+            "Block device scripts for ephemeral volumes not supported"
+        assert device.devtype == 'disk'
+        assert device.rw, 'Encrypting read-only volumes makes no sense'
+        path = self.volatile_volume_path(qube_name, device.name)
+        return qubes.storage.BlockDevice(
+            path='/dev/mapper/' + path,
```
marmarek, line 414: In LVM pool you already return the full path from `volatile_volume_path`. And given the `path` in the function name, I'd expect it this way.

qubes/storage/__init__.py
```diff
@@ -339,6 +381,43 @@ def size(self, size):
         # pylint: disable=attribute-defined-outside-init
         self._size = int(size)
 
+    # pylint: disable=no-self-use
+    def volatile_volume_path(self, qube_name, device_name):
```
marmarek, line 385: Maybe rather name it with `encrypted` or `ephemeral` instead of `volatile`? Otherwise it can be easily confused with the underlying `volatile` volume.

marmarek (changes requested):
Besides the few comments, generally it looks good. But please add unit tests. You can take inspiration from `tests/storage_lvm.py`.

qubes/tools/qubesd.py
marmarek: Yes, `os.getrandom(1)` would be better.

qubes/storage/__init__.py
```diff
@@ -134,8 +134,12 @@ def __init__(self, name, pool, vid,
         self.source = source
         #: Volume unique (inside given pool) identifier
         self.vid = vid
+        #: Asynchronous lock for cryptsetup operations
+        self._crypt_lock = asyncio.Lock()
         #: Asynchronous lock for @Volume.locked decorator
         self._lock = asyncio.Lock()
+        #: Name of the encrypted version of the volume
+        self.encrypted_name = None
```
marmarek, line 142: You set this only in `start_encrypted`, and then condition cleanup based on this variable. Restarting qubesd in the meantime will make it skip the cleanup :(

qubes/storage/__init__.py
```diff
@@ -134,8 +134,12 @@ def __init__(self, name, pool, vid,
         self.source = source
         #: Volume unique (inside given pool) identifier
         self.vid = vid
+        #: Asynchronous lock for cryptsetup operations
+        self._crypt_lock = asyncio.Lock()
```
marmarek, line 138: What this lock guards against? Since it is per-volume lock, it won't prevent parallel cryptsetup calls for different volumes. And for a single volume, there is already VM-level lock that guarantee only a single start or stop is running at a time.

qubes/storage/__init__.py
```diff
@@ -158,6 +162,58 @@ def __xml__(self):
...
+        '''
+        assert name.startswith('/dev/mapper/'), 'Invalid path %r passed to cryptsetup' % name
+        async with self._crypt_lock:
+            if self.script is not None:
+                self._not_implemented("starting an encrypted volume with a script")
+            if self.encrypted_name is not None:
+                await qubes.utils.cryptsetup('--', 'close', self.encrypted_name)
+                self.encrypted_name = None
+            await qubes.utils.coro_maybe(self.start())
+            self.encrypted_name = name = name[12:]
```
marmarek, line 190: Maybe move the `name = name[12:]` just after the assert, so it's clear what this 12 is about.

marmarek (changes requested):

qubes/utils.py
```diff
@@ -246,3 +246,30 @@ async def void_coros_maybe(values):
...
+    '''
+    Run cryptsetup with the given arguments.  This method returns a coroutine.
+    '''
+    return run_program(
+        'sudo',
+        'cryptsetup',
+        *args,
+        executable='/usr/bin/sudo',
+        # otherwise cryptsetup tries to mlock() the entire locale archive :(
+        env={'LC_ALL':'C'},
```
marmarek, line 271: Please don't throw away the whole rest of the environment.

qubes/storage/__init__.py
DemiMarie: Good catch!  The purpose of this variable is so that the cleanup code knows what name to use, but that is only an issue when restarting a volume, and only the LVM pool needs it.  The LVM pool can calculate that info on its own, though.

qubes/storage/lvm.py
```diff
@@ -339,6 +339,12 @@ async def _reset(self):
         assert not self.snap_on_start and not self.save_on_stop, \
             "Not a volatile volume"
         self.log.debug('Resetting volatile %s', self.vid)
+        if self.encrypted_name is not None:
+            # self.encrypted_volume_path doesn’t actually use either argument,
+            # so just pass empty strings 🙂
+            await qubes.utils.cryptsetup(
+                    '--', 'close', self.encrypted_volume_path('', ''))
+            self.encrypted_name = None
```
marmarek, line 347: This isn't needed. `_reset()` is called only from `start()`, and with `start_encrypted()` already has this very call before calling `start()`.
And indeed tests complain about failing (the second) cryptsetup call.

qubes/storage/__init__.py
```diff
@@ -158,6 +160,60 @@ def __xml__(self):
         config = _sanitize_config(self.config)
         return lxml.etree.Element('volume', **config)
 
+    def ephemeral(self):
```
marmarek, line 163: Shouldn't this be a property?

marmarek (changes requested):

qubes/storage/__init__.py
```diff
@@ -158,6 +160,61 @@ def __xml__(self):
         config = _sanitize_config(self.config)
         return lxml.etree.Element('volume', **config)
 
+    @property
+    def ephemeral(self):
+        '''Should this volume be encrypted with an ephemeral key in dom0?
```
fepitre, line 165: Can use please replace triple single quotes by triple double quotes everywhere please? (possibly as a separate commit) We should add such checks into lint test.

DemiMarie: That would be a separate PR ― this is consistent with the rest of the code, IIUC.

fepitre: Separate commit on files you edited is fine. Generally this is what I do but as you prefer :)

DemiMarie:
This is strange: the code _always_ works when I run it locally, but _never_ on CI.

marmarek:
Oh!
```
  Volatile volume test ... sudo: cryptsetup: command not found
```

fepitre:
> Oh!
> 
> ```
>   Volatile volume test ... sudo: cryptsetup: command not found
> ```

We can add it to the base image if needed?

marmarek:
Yes, please :)

qubes/utils.py
```diff
@@ -246,3 +246,30 @@ async def void_coros_maybe(values):
...
+
+def cryptsetup(*args):
+    """
+    Run cryptsetup with the given arguments.  This method returns a future.
+    """
+    return run_program(
+        'sudo',
+        'cryptsetup',
+        *args,
+        executable='/usr/bin/sudo',
```
marmarek, line 269: Please spare the logs and use sudo only when necessary - as you did in #399. Actually, it makes sense to use the same mechanism (and I consider #399 a prerequisite, because otherwise this PR will outright fail to start a VM in a file pool). So, lets wait for merging #399, and then fix this while doing rebase.

marmarek:
One more thing I wonder - does it make sens to make this feature configurable (opt-out)? I'm mainly wondering about the performance, but given the swap should not be that heavily used, maybe it doesn't matter?

qubes/utils.py
DemiMarie: Why will this PR otherwise fail to start a VM in a file pool?  Volatile volumes do not use a custom block script even with the current file pool, IIRC.  Non-volatile volumes do, but this PR does not alter how those are handled.

marmarek: Ah, indeed volatile volumes don't use script. Ok, lets see what openQA says (it doesn't supersedes the need for unit tests).

marmarek:
> Installing updates failed, skipping rest of the report!

```
Aug 17 20:00:29 dom0 qubesd[32976]: Traceback (most recent call last):
Aug 17 20:00:29 dom0 qubesd[32976]:   File "/usr/lib/python3.8/site-packages/qubes/api/__init__.py", line 286, in respond
Aug 17 20:00:29 dom0 qubesd[32976]:     response = await self.mgmt.execute(
Aug 17 20:00:29 dom0 qubesd[32976]:   File "/usr/lib/python3.8/site-packages/qubes/api/admin.py", line 846, in vm_start
Aug 17 20:00:29 dom0 qubesd[32976]:     await self.dest.start()
Aug 17 20:00:29 dom0 qubesd[32976]:   File "/usr/lib/python3.8/site-packages/qubes/vm/dispvm.py", line 241, in start
Aug 17 20:00:29 dom0 qubesd[32976]:     await super().start(**kwargs)
Aug 17 20:00:29 dom0 qubesd[32976]:   File "/usr/lib/python3.8/site-packages/qubes/vm/qubesvm.py", line 1159, in start
Aug 17 20:00:29 dom0 qubesd[32976]:     await self.storage.start()
Aug 17 20:00:29 dom0 qubesd[32976]:   File "/usr/lib/python3.8/site-packages/qubes/storage/__init__.py", line 714, in start
Aug 17 20:00:29 dom0 qubesd[32976]:     await qubes.utils.void_coros_maybe(
Aug 17 20:00:29 dom0 qubesd[32976]:   File "/usr/lib/python3.8/site-packages/qubes/utils.py", line 244, in void_coros_maybe
Aug 17 20:00:29 dom0 qubesd[32976]:     coros = [val for val in values if asyncio.iscoroutine(val)]
Aug 17 20:00:29 dom0 qubesd[32976]:   File "/usr/lib/python3.8/site-packages/qubes/utils.py", line 244, in <listcomp>
Aug 17 20:00:29 dom0 qubesd[32976]:     coros = [val for val in values if asyncio.iscoroutine(val)]
Aug 17 20:00:29 dom0 qubesd[32976]:   File "/usr/lib/python3.8/site-packages/qubes/storage/__init__.py", line 718, in <genexpr>
Aug 17 20:00:29 dom0 qubesd[32976]:     for name, vol in self.vm.volumes)
Aug 17 20:00:29 dom0 qubesd[32976]: ValueError: too many values to unpack (expected 2)
```

qubes/storage/__init__.py
```diff
@@ -607,15 +697,33 @@ async def remove(self):
...
+                    yield block_dev
+
     async def start(self):
         ''' Execute the start method on each volume '''
         await qubes.utils.void_coros_maybe(
-            vol.start() for vol in self.vm.volumes.values())
+            # pylint: disable=line-too-long
+            (vol.start_encrypted(vol.encrypted_volume_path(self.vm.name, name))
+             if vol.ephemeral else vol.start())
+            for name, vol in self.vm.volumes)
```
marmarek, line 718: You need this:
```suggestion
            for name, vol in self.vm.volumes.items())
```
And same in `stop()`.

marmarek (changes requested):

qubes/storage/__init__.py
DemiMarie: :woman_facepalming: Oops!

marmarek:
Now the tests failed with:
```
Aug 18 06:14:47 dom0 qubesd[6685]: vm.fedora-34-xfce: Starting fedora-34-xfce
Aug 18 06:14:48 dom0 qubesd[61398]: Device qubes_dom0-vm--fedora--34--xfce--volatile@crypt is not active.
Aug 18 06:14:48 dom0 qubesd[6685]: vm.fedora-34-xfce: Start failed: Command 'cryptsetup' returned non-zero exit status 4.
Aug 18 06:14:48 dom0 qubesd[6685]: unhandled exception while calling src=b'dom0' meth=b'admin.vm.Start' dest=b'fedora-34-xfce' arg=b'' len(untrusted_payload)=0
Aug 18 06:14:48 dom0 qubesd[6685]: Traceback (most recent call last):
Aug 18 06:14:48 dom0 qubesd[6685]:   File "/usr/lib/python3.8/site-packages/qubes/api/__init__.py", line 286, in respond
Aug 18 06:14:48 dom0 qubesd[6685]:     response = await self.mgmt.execute(
Aug 18 06:14:48 dom0 qubesd[6685]:   File "/usr/lib/python3.8/site-packages/qubes/api/admin.py", line 846, in vm_start
Aug 18 06:14:48 dom0 qubesd[6685]:     await self.dest.start()
Aug 18 06:14:48 dom0 qubesd[6685]:   File "/usr/lib/python3.8/site-packages/qubes/vm/qubesvm.py", line 1159, in start
Aug 18 06:14:48 dom0 qubesd[6685]:     await self.storage.start()
Aug 18 06:14:48 dom0 qubesd[6685]:   File "/usr/lib/python3.8/site-packages/qubes/storage/__init__.py", line 714, in start
Aug 18 06:14:48 dom0 qubesd[6685]:     await qubes.utils.void_coros_maybe(
Aug 18 06:14:48 dom0 qubesd[6685]:   File "/usr/lib/python3.8/site-packages/qubes/utils.py", line 248, in void_coros_maybe
Aug 18 06:14:48 dom0 qubesd[6685]:     task.result()  # re-raises exception if task failed
Aug 18 06:14:48 dom0 qubesd[6685]:   File "/usr/lib/python3.8/site-packages/qubes/storage/__init__.py", line 188, in start_encrypted
Aug 18 06:14:48 dom0 qubesd[6685]:     await qubes.utils.cryptsetup('--', 'close', self.encrypted_name)
Aug 18 06:14:48 dom0 qubesd[6685]:   File "/usr/lib/python3.8/site-packages/qubes/utils.py", line 257, in run_program
Aug 18 06:14:48 dom0 qubesd[6685]:     raise subprocess.CalledProcessError(p.returncode,
Aug 18 06:14:48 dom0 qubesd[6685]: subprocess.CalledProcessError: Command 'cryptsetup' returned non-zero exit status 4.
```
But not for all the cases, so it's either a race condition, or some other factor is involved, like if the VM was started before.

DemiMarie:
The failure happens below:

https://github.com/QubesOS/qubes-core-admin/blob/1768880e7b5d1f2e482debcbffc52dcd28ffebc3/qubes/storage/__init__.py#L187-L189

If the volume has been started and previously stopped, `self.encrypted_name` will not be `None`, but it will name an encrypted device that doesn’t exist.  That would cause the error seen above; I will add a test for this situation.

marmarek:
Udev rules need an update: https://openqa.qubes-os.org/tests/20638#step/qui_widgets_devices/5

marmarek:
Integration test verifies if volatile image is clean by comparing it to `/dev/zero`...
https://openqa.qubes-os.org/tests/20644#step/StorageFile/1

This actually will be an issue outside of tests too, if/when we'd like to extend per-volume encryption for persistent volumes too. VM startup scripts decide to initialize private volume (mkfs etc) by comparing it to zero. But lets not worry about it just yet.

DemiMarie:
> Udev rules need an update: https://openqa.qubes-os.org/tests/20638#step/qui_widgets_devices/5

Good catch!  This is also a legitimate security issue (these volumes would be scanned by udev when they should not).

DemiMarie:
The most recent revision requires (as yet unreleased) 4.1.13 of `qubes-core-dom0-linux` to provide the needed udev rules.

DemiMarie:
CI is failing because of the RPM dependency changes.  Since merging this patch without the changes would create a security vulnerability, I decided to be cautious.

marmarek:
PipelineRetry

marmarek:
I've found this one while reviewing test logs:
```
ERROR:vm.test-inst-backend:Failed to stop storage for domain test-inst-backend
  File "/usr/lib/python3.8/site-packages/qubes/vm/qubesvm.py", line 1285, in on_domain_stopped
    await self.storage.stop()
  File "/usr/lib/python3.8/site-packages/qubes/storage/__init__.py", line 712, in stop
    await qubes.utils.void_coros_maybe(
  File "/usr/lib/python3.8/site-packages/qubes/utils.py", line 276, in void_coros_maybe
    task.result()  # re-raises exception if task failed
  File "/usr/lib/python3.8/site-packages/qubes/storage/__init__.py", line 218, in wrapper
    return await method(self, *args, **kwargs)
  File "/usr/lib/python3.8/site-packages/qubes/storage/lvm.py", line 662, in stop
    await self._commit()
  File "/usr/lib/python3.8/site-packages/qubes/storage/lvm.py", line 408, in _commit
    await qubes_lvm_coro(cmd, self.log)
  File "/usr/lib/python3.8/site-packages/qubes/storage/lvm.py", line 798, in qubes_lvm_coro
    return _process_lvm_output(p.returncode, out, err, log)
  File "/usr/lib/python3.8/site-packages/qubes/storage/lvm.py", line 776, in _process_lvm_output
    raise qubes.storage.StoragePoolException(err)
qubes.storage.StoragePoolException:   Device open /dev/mapper/qubes_dom0-vm--test--inst--backend--volatile@crypt 253:68 failed errno 2
  Device open /dev/mapper/qubes_dom0-vm--test--inst--backend--volatile@crypt 253:68 failed errno 2
  WARNING: Scan ignoring device 253:68 with no paths.
```

marmarek:
And there is also:
```
device-mapper: remove ioctl on qubes_dom0-vm--test--inst--frontend--volatile@crypt  failed: Device or resource busy
device-mapper: remove ioctl on qubes_dom0-vm--test--inst--frontend--volatile@crypt  failed: Device or resource busy
```
(note two spaces before "failed", not sure if relevant)
Unfortunately, I don't have a traceback for this one.

DemiMarie:
I think I figured out the problem.  The tricky part is that this PR is actually correct ― it merely unmasks a preexisting race condition.

The problem is in `_kill_locked`:

https://github.com/QubesOS/qubes-core-admin/blob/792437bf0946500447211948ca0d6d70515058e5/qubes/vm/qubesvm.py#L1350-L1362

This destroys the VM and then calls `_ensure_shutdown_handled` (below) to wait for post-shutdown tasks to complete.  

https://github.com/QubesOS/qubes-core-admin/blob/792437bf0946500447211948ca0d6d70515058e5/qubes/vm/qubesvm.py#L1075-L1101

However, `_ensure_shutdown_handled` does not actually wait for libvirt to send `VIR_DOMAIN_EVENT_STOPPED`.   This creates a race condition: it is possible for the storage to be stopped before the VM has actually shut down.  The storage will still be in use at this time.

Prior to this PR, this bug did not actually cause problems.  In the case of the file and reflink pools, the actual loop device is cleaned up by the block script, and the file on disk is deleted when the last reference to it goes away. 
 The LVM2 pool, on the other hand, calls `blkdiscard` before removing the storage.  `blkdiscard` does _not_ require the domain to have stopped, and on thin pools is slow due to a Linux kernel bug.  Therefore, by the time LVM2 is actually told to remove the volume, the VM will likely have actually stopped down, releasing the volume.

However, `cryptsetup close` is called _before_ `blkdiscard`, and it _does_ require the volume to be unused.  This makes the problem much more likely to show up.  This is sometimes hidden because `cryptsetup close` has built-in retry logic, so if the VM stops soon enough, the only symptom will be a warning printed to stderr.  However, it is also possible for it to fail outright, causing problems.

The fix is to ensure that the storage does not get stopped while it is still in use, by waiting for `VIR_DOMAIN_EVENT_STOPPED` from libvirt.  I will make a PR for this, which will also avoid a wait-loop in `shutdown`:

https://github.com/QubesOS/qubes-core-admin/blob/792437bf0946500447211948ca0d6d70515058e5/qubes/vm/qubesvm.py#L1316-L1327

marmarek:
> However, `_ensure_shutdown_handled` does not actually wait for libvirt to send `VIR_DOMAIN_EVENT_STOPPED`. This creates a race condition: it is possible for the storage to be stopped before the VM has actually shut down. The storage will still be in use at this time.

The libvirt destroy call is synchronous - it returns only after all the cleanup is completed. Specifically, libvirt (at least libxl driver) sends `VIR_DOMAIN_EVENT_STOPPED` event before returning from the destroy function: https://github.com/libvirt/libvirt/blob/master/src/libxl/libxl_driver.c#L1384

> This is sometimes hidden because `cryptsetup close` has built-in retry logic, so if the VM stops soon enough, the only symptom will be a warning printed to stderr.

This indeed may explain some `Device or resource busy` messages, not not the lvm commit failing with ENOENT...

> The fix is to ensure that the storage does not get stopped while it is still in use, by waiting for `VIR_DOMAIN_EVENT_STOPPED` from libvirt.

This indeed may avoid double-stop issue (I think it isn't the only way, but probably the best one). Be sure to not release the startup lock between domain destroy call, and (receiving and) handling the shutdown cleanup, including storage stop. It is important to finish shutdown cleanup (including possibly DispVM remove!) before allowing subsequent start.

qubes/vm/qubesvm.py
```diff
@@ -1359,6 +1361,8 @@ async def _kill_locked(self):
             raise
 
         # make sure all shutdown tasks are completed
+        self.__waiter = asyncio.get_running_loop().create_future()
```
marmarek, line 1364: I think this should be before `destroy()` call.

qubes/vm/qubesvm.py
```diff
@@ -1359,6 +1361,8 @@ async def _kill_locked(self):
...
+        await self.__waiter
```
marmarek, line 1365: Please add some timeout. If the event is lost for any reason (including libvirt restart or crash for example), this would wait indefinitely, *while holding the self.startup_lock*.

marmarek (changes requested):

qubes/vm/qubesvm.py
```diff
@@ -1359,6 +1362,10 @@ async def _kill_locked(self):
             raise
 
         # make sure all shutdown tasks are completed
+        try:
+            await asyncio.wait_for(self.__waiter, timeout=10_000_000)
```
marmarek, line 1366: Timeout is in seconds, so this is way too long.

DemiMarie: oops… I thought it was in microseconds :woman_facepalming:

qubes/storage/__init__.py
```diff
@@ -40,6 +40,8 @@
 
 STORAGE_ENTRY_POINT = 'qubes.storage'
 
+_am_root = os.getuid() == 0
+_big_buffer = b'\0' * 4096
```
marmarek, line 44: Can you make it bigger? For example ISO9660 has its header at 32KiB offset. I don't mean one wants to keep ISO9660 on volatile volume, but that 4KiB isn't universally enough to ensure no collision with some filesystem (detection tool).
Maybe 1MiB is indeed too excessive, but lets make it at least 64KiB.

marmarek (changes requested):

marmarek:
pylint complains about too long line

In the meantime, I can (again?) confirm the "Device or resource busy" message is harmless:
```
device-mapper: remove ioctl on qubes_dom0-vm--test--inst--backend--volatile@crypt  failed: Device or resource busy
device-mapper: remove ioctl on qubes_dom0-vm--test--inst--backend--volatile@crypt  failed: Device or resource busy
DEBUG:qubes.storage.lvm.vm-pool:Invoked with arguments ['remove', 'qubes_dom0/vm-test-inst-backend-volatile']
DEBUG:qubes.storage.lvm.vm-pool:Invoked with arguments ['lvm', 'lvremove', '-f', 'qubes_dom0/vm-test-inst-backend-volatile']
DEBUG:qubes.storage.lvm.vm-pool:b'  Logical volume "vm-test-inst-backend-volatile" successfully removed\n'
```

DemiMarie:
> pylint complains about too long line
> 
> In the meantime, I can (again?) confirm the "Device or resource busy" message is harmless:
> 
> ```
> device-mapper: remove ioctl on qubes_dom0-vm--test--inst--backend--volatile@crypt  failed: Device or resource busy
> device-mapper: remove ioctl on qubes_dom0-vm--test--inst--backend--volatile@crypt  failed: Device or resource busy
> DEBUG:qubes.storage.lvm.vm-pool:Invoked with arguments ['remove', 'qubes_dom0/vm-test-inst-backend-volatile']
> DEBUG:qubes.storage.lvm.vm-pool:Invoked with arguments ['lvm', 'lvremove', '-f', 'qubes_dom0/vm-test-inst-backend-volatile']
> DEBUG:qubes.storage.lvm.vm-pool:b'  Logical volume "vm-test-inst-backend-volatile" successfully removed\n'
> ```

I do want to figure out the cause of it, since it should not happen.  But it should not block merging either.

marmarek:
PipelineRetry

marmarek (approved):

HW42:
@marmarek wrote:
> One more thing I wonder - does it make sens to make this feature configurable (opt-out)? I'm mainly wondering about the performance, but given the swap should not be that heavily used, maybe it doesn't matter?

If I'm not missing something this was never addressed, right? Given that this is only really useful for a, in my opinion, niche use case I'm wonder if having this always on is really a good idea. It has been a while since I looked at numbers, but IIRC dm-crypt had a quite noticeable impact on fast storage like SSDs. And on volatile we even have swap, so something particular performance sensitive. In my experience quite a bit of space if freed when I shutdown all my VMs, so it seems volatile gets used regularly (Didn't investigate if it's swap or rw for rootfs).

marmarek:
Indeed, I missed this one before merging. Logically, it should be a volume's property, but that isn't inherited when creating new DispVM - which is the main use case for the feature. Maybe VM's property (`encrypt_volatile"`? )? Or a "feature"?

DemiMarie:
Sorry about that, @HW42!

PR #397, opened 2021-04-02
qubesd: at startup, stop storage for stopped domains

DemiMarie:
If qubesd is not cleanly shut down, it is possible that storage was also
not properly stopped.  Therefore, we need to stop storage for all
domains that are not currently running.

### DIFF
--- a/qubes/app.py
+++ b/qubes/app.py
@@ -20,6 +20,7 @@
 # License along with this library; if not, see <https://www.gnu.org/licenses/>.
 #
 
+import asyncio
 import collections.abc
 import copy
 import functools
@@ -65,7 +66,6 @@ import qubes.vm
 import qubes.vm.adminvm
 import qubes.vm.qubesvm
 import qubes.vm.templatevm
-
 
 # pylint: enable=wrong-import-position
 
@@ -1403,6 +1403,26 @@ class Qubes(qubes.PropertyHolder):
         except KeyError:
             raise qubes.exc.QubesException('No driver %s for pool %s' %
                                            (driver, name))
+
+    async def stop_storage(self):
+        """
+        Stop the storage of all domains that are not running.
+        """
+        async def stop(i):
+            async with i.startup_lock:
+                if not i.is_running():
+                    await i.storage.stop()
+        future = tuple(stop(i) for i in self.domains if i.klass != 'AdminVM')
+        finished = ()
+        while future:
+            qubes.utils.systemd_extend_timeout()
+            finished, future = await asyncio.wait(future, timeout=30000000)
+            for i in finished:
+                try:
+                    await i
+                except Exception:  # pylint: disable=broad-except
+                    self.log.exception(
+                        'Stopping storage for a qube raised an exception')
 
     def register_event_handlers(self, old_connection=None):
         """Register libvirt event handlers, which will translate libvirt
--- a/qubes/storage/lvm.py
+++ b/qubes/storage/lvm.py
@@ -767,6 +767,9 @@ async def qubes_lvm_coro(cmd, log=logging.getLogger('qubes.storage.lvm')):
     ''' Call :program:`lvm` to execute an LVM operation '''
     environ={'LC_ALL': 'C.UTF-8', **os.environ}
     if cmd[0] == "remove":
+        if not os.path.exists('/dev/' + cmd[1]):
+            # ignore errors if the file does not exist
+            return
         pre_cmd = ['blkdiscard', '-p', '1G', '/dev/'+cmd[1]]
         p = await asyncio.create_subprocess_exec(*pre_cmd,
             stdout=subprocess.DEVNULL,
--- a/qubes/storage/lvm.py
+++ b/qubes/storage/lvm.py
@@ -340,7 +340,9 @@ class ThinVolume(qubes.storage.Volume):
             "Not a volatile volume"
         self.log.debug('Resetting volatile %s', self.vid)
         try:
-            cmd = ['remove', self.vid]
+            assert self.path.startswith('/dev/'), \
+                'bad value for self.path: %r' % self.path
+            cmd = ['remove', self.path[5:]]
             await qubes_lvm_coro(cmd, self.log)
         except qubes.storage.StoragePoolException:
             pass
@@ -456,7 +458,9 @@ class ThinVolume(qubes.storage.Volume):
         await self._remove_revisions(self.revisions.keys())
         if not os.path.exists(self.path):
             return
-        cmd = ['remove', self.path]
+        assert self.path.startswith('/dev/'), \
+            'bad value for self.path: %r' % self.path
+        cmd = ['remove', self.path[5:]]
         await qubes_lvm_coro(cmd, self.log)
         await reset_cache_coro()
         # pylint: disable=protected-access
@@ -627,6 +631,8 @@ class ThinVolume(qubes.storage.Volume):
         try:
             cmd = ['remove', self._vid_snap]
             await qubes_lvm_coro(cmd, self.log)
+        except AssertionError:
+            raise
         except:  # pylint: disable=bare-except
             pass
 
@@ -723,6 +729,7 @@ def _get_lvm_cmdline(cmd):
     '''
     action = cmd[0]
     if action == 'remove':
+        assert not cmd[1].startswith('/'), 'absolute path to ‘remove’???'
         lvm_cmd = ['lvremove', '-f', cmd[1]]
     elif action == 'clone':
         lvm_cmd = ['lvcreate', '-kn', '-ay', '-s', cmd[1], '-n', cmd[2]]
--- a/qubes/storage/reflink.py
+++ b/qubes/storage/reflink.py
@@ -222,7 +222,8 @@ class ReflinkVolume(qubes.storage.Volume):
     @_coroutinized
     def stop(self):  # pylint: disable=invalid-overridden-method
         if self.save_on_stop:
-            self._commit(self._path_dirty)
+            if os.path.exists(self._path_dirty):
+                self._commit(self._path_dirty)
         else:
             if not self.snap_on_start:
                 self._size = self.size  # preserve manual resize of image
--- a/qubes/tools/qubesd.py
+++ b/qubes/tools/qubesd.py
@@ -7,6 +7,7 @@ import signal
 import libvirtaio
 
 import qubes
+import qubes.app
 import qubes.api
 import qubes.api.admin
 import qubes.api.internal
@@ -36,6 +37,9 @@ def main(args=None):
         raise
 
     args.app.register_event_handlers()
+
+    # Stop storage for domains not currently running
+    loop.run_until_complete(args.app.stop_storage())
 
     if args.debug:
         qubes.log.enable_debug()
--- a/qubes/storage/file.py
+++ b/qubes/storage/file.py
@@ -539,9 +539,10 @@ def copy_file(source, destination):
 
 def _remove_if_exists(path):
     ''' Removes a file if it exist, silently succeeds if file does not exist '''
-    if os.path.exists(path):
+    try:
         os.remove(path)
-
+    except FileNotFoundError:
+        pass
 
 def _check_path(path):
     ''' Raise an StoragePoolException if ``path`` does not exist'''
--- a/qubes/utils.py
+++ b/qubes/utils.py
@@ -168,7 +168,20 @@ def systemd_notify():
         nofity_socket = '\0' + nofity_socket[1:]
     sock = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
     sock.connect(nofity_socket)
-    sock.sendall(b'READY=1')
+    sock.sendall(b'READY=1\nSTATUS=qubesd online and operational\n')
+    sock.close()
+
+def systemd_extend_timeout():
+    """Extend systemd startup timeout by 60s"""
+    notify_socket = os.getenv('NOTIFY_SOCKET')
+    if not notify_socket:
+        return
+    if notify_socket.startswith('@'):
+        notify_socket = '\0' + notify_socket[1:]
+    sock = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
+    sock.connect(notify_socket)
+    sock.sendall(b'EXTEND_TIMEOUT_USEC=60000000\n'
+                 b'STATUS=Cleaning up storage for stopped qubes\n')
     sock.close()
 
 def match_vm_name_with_special(vm, name):


rustybird:
Conceptually, this looks good: E.g. volatile volumes would be cleaned up after a system crash sooner than whenever the the VM happens to be started and stopped again.

Two general thoughts before I bother you with line comments:

- If a system with let's say 20 running VMs crashes, then the next reboot is going to frontload 60 `Volume.stop()` invocations. Which, in light of QubesOS/qubes-issues#5426, could take a long time especially on lvm_thin, and (because it happens in `register_event_handlers()`, before signalling readiness to systemd) it could exceed qubesd.service's 90s default `TimeOutStartSec`.

- Is the `is_running()` check really comprehensive enough? I'm never sure about startup/shutdown/exotic Xen states. @marmarek

DemiMarie:
> * If a system with let's say 20 running VMs crashes, then the next reboot is going to frontload 60 substantial `Volume.stop()` invocations, plus the insubstantial invocations for all other VMs. Which, in light of [QubesOS/qubes-issues#5426](https://github.com/QubesOS/qubes-issues/issues/5426), could take a long time especially on lvm_thin, and (because it happens in `register_event_handlers()`, before signalling readiness to systemd) it could exceed qubesd.service's 90s default `TimeOutStartSec`.

This is indeed a potential problem.  The operations *should* all be done in parallel (if not, that’s a bug), which should help substantially.  qubesd could do this after it starts to receive events, but this would require additional synchronization to prevent races.

marmarek:
>     * Is the `is_running()` check really comprehensive enough?

Yes, this should be fine. 

As for parallelism, I'm not sure how well dm-thin parallelize DISCARD calls, but LVM does take a lock for updating its metadata when adding/removing volumes. So, it isn't perfect.

qubes/storage/lvm.py
```diff
@@ -818,6 +818,11 @@ def qubes_lvm_coro(cmd, log=logging.getLogger('qubes.storage.lvm')):
     environ = os.environ.copy()
     environ['LC_ALL'] = 'C.utf8'
     if cmd[0] == "remove":
+        try:
+            os.stat('/dev/' + cmd[1])
+        except FileNotFoundError:
+            # ignore errors if the file does not exist
```
rustybird, line 824:
```suggestion
        if not os.path.exists('/dev/' + cmd[1]):
```

qubes/storage/reflink.py
```diff
@@ -227,14 +227,23 @@ def stop(self):
...
-            _remove_file(self._path_dirty)
-            _remove_file(self._path_clean)
+            try:
+                _remove_file(self._path_dirty)
+            except FileNotFoundError:
+                pass
+            try:
+                _remove_file(self._path_clean)
+            except FileNotFoundError:
+                pass
```
rustybird, line 237: This isn't necessary,  the reflink module's `_remove_file()` already suppresses `FileNotFoundError`.

qubes/storage/reflink.py
```diff
@@ -227,14 +227,23 @@ def stop(self):
...
         return self
 
     def _commit(self, path_from):
         self._add_revision()
         self._prune_revisions()
-        qubes.utils.fsync_path(path_from)
+        try:
+            qubes.utils.fsync_path(path_from)
+        except FileNotFoundError:
+            return
```
rustybird, line 246: Better just prepend

```
if os.path.exists(self._path_dirty):
```

before the `_commit()` call in `stop()` instead of this. Otherwise the revision stuff at the beginning of `_commit()` would make it non-idempotent.

rustybird (commented):
Some line comments before I forget. Feel free to ignore them while we bang our heads against the performance issue

qubes/storage/lvm.py
DemiMarie: Good catch, thanks!

qubes/storage/reflink.py
DemiMarie: Indeed it does.

qubes/storage/reflink.py
DemiMarie: Good point!  Fixed.

DemiMarie (commented):
Thanks for the review, @rustybird!

DemiMarie:
As far as the performance problem is concerned, one approach is for qubesd to be configured to automatically restart.  My understanding is that qubesd will make forward progress on each failed startup, until it eventually has cleared out the queue and starts normally.

marmarek:
> As far as the performance problem is concerned, one approach is for qubesd to be configured to automatically restart. My understanding is that qubesd will make forward progress on each failed startup, until it eventually has cleared out the queue and starts normally.

Failed qubesd startup is bad. Until it will (finally) successfully start, all the requests made to it will fail. Basically, this makes startup dependencies useless. You'll likely end up with no autostarted VMs, which may be "inconvenient" if your input devices live in sys-usb...

So, any design assuming qubesd would fail to start as "a normal thing" is bad.
Maybe have some separate systemd service (with much longer timeout) doing the cleanup, before qubesd start? It could use the same code base, and since qubesd is not running then yet, it should be safe to construct `Qubes()` object and load qubes.xml.

DemiMarie:
> > As far as the performance problem is concerned, one approach is for qubesd to be configured to automatically restart. My understanding is that qubesd will make forward progress on each failed startup, until it eventually has cleared out the queue and starts normally.
> 
> Failed qubesd startup is bad. Until it will (finally) successfully start, all the requests made to it will fail. Basically, this makes startup dependencies useless. You'll likely end up with no autostarted VMs, which may be "inconvenient" if your input devices live in sys-usb...
> 
> So, any design assuming qubesd would fail to start as "a normal thing" is bad.
> Maybe have some separate systemd service (with much longer timeout) doing the cleanup, before qubesd start? It could use the same code base, and since qubesd is not running then yet, it should be safe to construct `Qubes()` object and load qubes.xml.

What about doing cleanup asynchronously in the background?  That would solve a bunch of other problems too.  Otherwise, the system would be (very) slow to start, which is annoying, to say the least.

@iamhuman has an extension that does this for R4.0.

marmarek:
> What about doing cleanup asynchronously in the background? That would solve a bunch of other problems too. Otherwise, the system would be (very) slow to start, which is annoying, to say the least.

IIUC this just needs to ensure the cleanup for a given VM will complete before starting that VM. Are there any other constrains?

DemiMarie:
> > What about doing cleanup asynchronously in the background? That would solve a bunch of other problems too. Otherwise, the system would be (very) slow to start, which is annoying, to say the least.
> 
> IIUC this just needs to ensure the cleanup for a given VM will complete before starting that VM. Are there any other constrains?

We need to clean up eventually or we will leak disk space.

DemiMarie:
> > As far as the performance problem is concerned, one approach is for qubesd to be configured to automatically restart. My understanding is that qubesd will make forward progress on each failed startup, until it eventually has cleared out the queue and starts normally.
> 
> Failed qubesd startup is bad. Until it will (finally) successfully start, all the requests made to it will fail. Basically, this makes startup dependencies useless. You'll likely end up with no autostarted VMs, which may be "inconvenient" if your input devices live in sys-usb...

Agreed.  The problem will go away after enough reboots, but it is still a horrible user experience.

> So, any design assuming qubesd would fail to start as "a normal thing" is bad.
> Maybe have some separate systemd service (with much longer timeout) doing the cleanup, before qubesd start? It could use the same code base, and since qubesd is not running then yet, it should be safe to construct `Qubes()` object and load qubes.xml.

systemd actually has built-in support for this via `EXTEND_TIMEOUT_USEC`.  qubesd can tell systemd that it needs more time to start, and can do so as many times as is necessary.  So long as stopping storage doesn’t block the event loop, qubesd can set a timer that sends `EXTEND_TIMEOUT_USEC` periodically.  This is implemented in 2863f68e3ca9bd40b7cb2516b2e844aec729c7b0.

marmarek:
> systemd actually has built-in support for this via `EXTEND_TIMEOUT_USEC`. qubesd can tell systemd that it needs more time to start, and can do so as many times as is necessary. So long as stopping storage doesn’t block the event loop, qubesd can set a timer that sends `EXTEND_TIMEOUT_USEC` periodically. This is implemented in [2863f68](https://github.com/QubesOS/qubes-core-admin/commit/2863f68e3ca9bd40b7cb2516b2e844aec729c7b0).

Ok, this sounds like an acceptable solution. A minor remark: add some safety margin to the timeout - send the `EXTEND_TIMEOUT_USEC` just before the timeout would expire (or maybe even half way there?), instead of risking missing the timeout.

Just to be sure: this cleanup will take a long time _only_ after unclean shutdown, right? After clean shutdown it will be a no-op in practice, right?

marmarek:
Also, see CI failures.

DemiMarie:
> > systemd actually has built-in support for this via `EXTEND_TIMEOUT_USEC`. qubesd can tell systemd that it needs more time to start, and can do so as many times as is necessary. So long as stopping storage doesn’t block the event loop, qubesd can set a timer that sends `EXTEND_TIMEOUT_USEC` periodically. This is implemented in [2863f68](https://github.com/QubesOS/qubes-core-admin/commit/2863f68e3ca9bd40b7cb2516b2e844aec729c7b0).
> 
> Ok, this sounds like an acceptable solution. A minor remark: add some safety margin to the timeout - send the `EXTEND_TIMEOUT_USEC` just before the timeout would expire (or maybe even half way there?), instead of risking missing the timeout.

There is already such a call at the beginning of the cleanup loop, so the code should always have 60s of safety margin.  Is that enough?  I could also try to actually calculate how much time is left.

> Just to be sure: this cleanup will take a long time _only_ after unclean shutdown, right? After clean shutdown it will be a no-op in practice, right?

Yes, after a clean shutdown it should be a no-op (if not, that’s a bug).

marmarek:
> There is already such a call at the beginning of the cleanup loop, so the code should always have 60s of safety margin. Is that enough? I could also try to actually calculate how much time is left.

My reading of the doc is it moves the timeout X time from _now_, not add to the existing time left:
> The value specified is a time in microseconds during which the service must send a new message

So, extending by 60s, every 60s is risky.

marmarek:
CI complains: pylint about one too long line, but tests fail too.

DemiMarie:
> > There is already such a call at the beginning of the cleanup loop, so the code should always have 60s of safety margin. Is that enough? I could also try to actually calculate how much time is left.
> 
> My reading of the doc is it moves the timeout X time from _now_, not add to the existing time left:
> 
> > The value specified is a time in microseconds during which the service must send a new message
> 
> So, extending by 60s, every 60s, is risky.

I will extend it by 60s every 30s.  That would be a bad idea if it added to the existing time left, but it is very simple and works with what systemd actually does.

qubes/app.py
```diff
@@ -1424,6 +1425,20 @@ def register_event_handlers(self, old_connection=None):
...
 
+        # Stop storage for domains not currently running
+        loop = asyncio.get_event_loop()
+        async def stop_storage():
+            future = (qubes.utils.void_coros_maybe(
+                i.storage.stop() for i in self.domains if not i.is_running()),)
+            finished = ()
+            while future:
+                qubes.utils.systemd_extend_timeout()
+                # pylint: disable=line-too-long
```
marmarek, line 1436: Please wrap the line instead

qubes/app.py
```diff
@@ -1424,6 +1425,20 @@ def register_event_handlers(self, old_connection=None):
                 self._domain_event_callback,
                 None))
 
+        # Stop storage for domains not currently running
+        loop = asyncio.get_event_loop()
+        async def stop_storage():
+            future = (qubes.utils.void_coros_maybe(
+                i.storage.stop() for i in self.domains if not i.is_running()),)
```
marmarek, line 1432: This should take `i.startup_lock`, to avoid starting a VM when `storage.stop()` is still running.
(even if you think it couldn't happen for other reasons, better to take the lock anyway - see an example above why such assumptions are dangerous...)

qubes/app.py
```diff
@@ -1424,6 +1425,20 @@ def register_event_handlers(self, old_connection=None):
                 self._domain_event_callback,
                 None))
 
+        # Stop storage for domains not currently running
```
marmarek, line 1428: `register_event_handlers` is not a good place for this. If nothing else, then because it is called on reconnection to (possibly restarted) libvirt - even if qubesd hasn't been restarted.
I'd rather prefer another call in `qubesd.py`.

marmarek (changes requested):

qubes/tools/qubesd.py
```diff
@@ -55,6 +55,22 @@ def main(args=None):
...
+                    await i.storage.stop()
+        future = (qubes.utils.void_coros_maybe(stop(i) for i in self.domains),)
+        finished = ()
+        while future:
+            qubes.utils.systemd_extend_timeout()
+            finished, future = await asyncio.wait(
+                future, loop=loop, timeout=30000000)
+        for i in finished:
+            await i
+    loop.run_until_complete(stop_storage())
```
marmarek, line 72: I meant to keep the `stop_storage()` function in app.py and just call it from qubesd.py.

marmarek (changes requested):

qubes/tools/qubesd.py
DemiMarie: Ah okay!  Easy fix :slightly_smiling_face:

qubes/app.py
```diff
@@ -1404,6 +1404,24 @@ def _get_pool(**kwargs):
...
+                if not i.is_running():
+                    await i.storage.stop()
+        future = (qubes.utils.void_coros_maybe(stop(i)
+            for i in self.domains if i.klass != 'AdminVM'),)
+        finished = ()
+        while future:
+            qubes.utils.systemd_extend_timeout()
+            finished, future = await asyncio.wait(
+                future, loop=loop, timeout=30000000)
+        for i in finished:
```
marmarek, line 1422: This awaits only tasks from the last iteration (you override `finished` variable).

qubes/app.py
```diff
@@ -1404,6 +1404,24 @@ def _get_pool(**kwargs):
...
 
+    async def stop_storage(self, loop):
+        """
+        Stop the storage of all domains that are not running.
+        """
+        async def stop(i):
+            async with i.startup_lock:
+                if not i.is_running():
+                    await i.storage.stop()
+        future = (qubes.utils.void_coros_maybe(stop(i)
```
marmarek, line 1415: Since `stop()` is always an async function, I think you don't need `void_coros_maybe` helper here.

marmarek (changes requested):

qubes/app.py
DemiMarie: `asyncio.wait` splits the futures passed to it between its two return values.  Here, `future` starts as a 1-tuple, so when `finished` becomes nonempty, `future` will become empty and the loop will exit.  Therefore, I believe this code is currently correct as is, but it should certainly be made clearer, and removing the use of `void_coros_maybe` will require it to be fixed.

qubes/app.py
DemiMarie: Good idea; this also makes the code cleaner, since `asyncio.wait` takes an iterable already.

qubes/app.py
marmarek: Ah, indeed the `void_coros_maybe` was called on a generator, not on each `stop(i)` call separately.

qubes/app.py
```diff
@@ -1404,6 +1404,22 @@ def _get_pool(**kwargs):
...
+            async with i.startup_lock:
+                if not i.is_running():
+                    await i.storage.stop()
+        future = tuple(stop(i) for i in self.domains if i.klass != 'AdminVM')
+        finished = ()
+        while future:
+            qubes.utils.systemd_extend_timeout()
+            finished, future = await asyncio.wait(future, timeout=30000000)
+            for i in finished:
+                await i
```
marmarek, line 1421: This will interrupt processing if _any_ stop fails. Two important questions here:
1. Should it keep waiting for other calls to finish instead (currently it wont, if they take more time than the current iteration)? I think yes.
2. Should the exception prevent qubesd startup (it will now)? I think it shouldn't. Generally failing qubesd startup will make the system mostly unusable. If you use USB keyboard+mouse and sys-usb, then literally unusable. On the other hand, the very VM that has failed storage stop, most likely be in a weird state. In any case, having one broken VM should not prevent using others.

DemiMarie: Perhaps `.start()` should include an implicit `.stop()`?  I really don’t want to be able to start a qube if it is in a broken state.

marmarek: >  I really don’t want to be able to start a qube if it is in a broken state.

That isn't really a concern. `start()` should be able to deal with a semi-broken state (like - interrupted stop), or raise an exception if it cannot. After all, before this PR, `start()` was a normal path to recovery from a hard system reset.

marmarek (approved):

PR #398, opened 2021-04-08
api: improve handling destination removed just before the call

3hhh:
If the destination domain doesn't exist anymore when the call gets
to qubesd, inform the client that it may try again later with more
success rather than giving them a generic PermissionDenied error.
This enables client applications to handle such edge cases by e.g.
waiting for a short time before trying again.

Fixes QubesOS/qubes-issues#5105

---------------------------------------

I manually tested this a bit and it seems to work.
Only drawback is that GUI clients would need to explicitly handle this exception (I saw it for one of the widgets so far) or otherwise annoying popups may randomly appear.

If the general approach would be accepted, I'd look deeper into the GUI code.

### DIFF
--- a/qubes/api/__init__.py
+++ b/qubes/api/__init__.py
@@ -118,20 +118,21 @@ class AbstractQubesAPI:
         #: :py:class:`qubes.Qubes` object
         self.app = app
 
-        #: source qube
-        self.src = self.app.domains[src.decode('ascii')]
-
         try:
+            vm = src.decode('ascii')
+            #: source qube
+            self.src = self.app.domains[vm]
+
+            vm = dest.decode('ascii')
             #: destination qube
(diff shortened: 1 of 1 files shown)


marmarek:
> This enables client applications to handle such edge cases by e.g.
> waiting for a short time before trying again.

If domain was removed, waiting wont help - it won't appear on its own, unless recreated. What I think you had in mind, is to refresh domains list and try again. I think more appropriate exception would be QubesVMNotFoundError.

3hhh:
> I think more appropriate exception would be QubesVMNotFoundError.

You were right - I was too focused on the list applications.

I updated the PR accordingly and also added PRs for the various components that explicitly required handling of the exception.

marmarek (approved):

PR #399, opened 2021-04-14
Improved file pool

DemiMarie:


3hhh:
I had a short look at this:
- `losetup --nooverlap` is not available in 4.0 dom0 at least. It was introduced in losetup 2.29 [afaik](https://github.com/3hhh/blib/blob/e33501fb5a3ce9b2a4473beb2fd010ebfa4793cd/lib/fs#L129).
- Not sure about removing the symlink functionality. Maybe someone is using it? (not me anyway)

DemiMarie:
> I had a short look at this:
> 
> * `losetup --nooverlap` is not available in 4.0 dom0 at least. It was introduced in losetup 2.29 [afaik](https://github.com/3hhh/blib/blob/e33501fb5a3ce9b2a4473beb2fd010ebfa4793cd/lib/fs#L129).

Not being available in R4.0 is fine.  The main purpose of this PR is to make the code more understandable, as I would like to get rid of block scripts entirely.

DemiMarie:
Looks like the tests need to run as root, which makes sense since the file pool cannot create a block device unless the underlying files actually exist.

marmarek:
> Looks like the tests need to run as root, which makes sense since the file pool cannot create a block device unless the underlying files actually exist.

You can run the actual script as root - see for example https://github.com/QubesOS/qubes-core-admin/blob/master/qubes/storage/lvm.py#L777-L780

I see currently tests fail on some mkdir. You can either skip `create_on_disk()` and call `create()` on individual volumes (this should work, as the pool is created in /tmp). Or patch basedir, like [here](https://github.com/QubesOS/qubes-core-admin/blob/master/qubes/tests/storage.py#L98)

qubes/storage/file.py
```diff
@@ -358,13 +381,30 @@ def start(self):
             if hasattr(self, 'path_source_cow'):
                 if not os.path.exists(self.path_source_cow):
                     create_sparse_file(self.path_source_cow, self.size)
+
+        async with _lock:
+            if self.save_on_stop:
+                assert not self.snap_on_start, 'unsupported configuration'
+                await qubes.utils.run_program('sudo', CREATE_SCRIPT, self.path,
```
marmarek, line 388: Better add sudo only when needed (otherwise it will pollute logs...) - maybe add it inside run_program function?

marmarek:
PipelineRetry

marmarek:
The above tests report is bogus, but file pool exploded (https://openqa.qubes-os.org/tests/17156). The overall test run timed out, so the logs are very limited, I'll restart it and maybe pause at the right time, or increase the timeout.

DemiMarie:
> The above tests report is bogus, but file pool exploded (https://openqa.qubes-os.org/tests/17156). The overall test run timed out, so the logs are very limited, I'll restart it and maybe pause at the right time, or increase the timeout.

Does the original file pool work?  I will try it on the test laptop.

marmarek:
> Does the original file pool work? I will try it on the test laptop.

I can retry, but I'm pretty sure it does. 

As for the logs, there is some FD leak. After waiting some more, tests crashed with "Too many open files". We do have also explicit check for FD leaks and object leaks after each test, and I believe "ERROR" messages I've seen were about that (sadly - without the full message - that's at the end). But I do have one more detail - a graph of leaked objects. See `var/log/objgraph-*` files in https://openqa.qubes-os.org/tests/17159/file/system_tests-var_log.tar.gz

linux/system-config/create-snapshot
```diff
@@ -0,0 +1,92 @@
...
+
+  dm_devname=$1
+  base_dev=$2
+  cow_dev=$3
+
+  if [ ! -e "/dev/mapper/$dm_devname" ]; then
+    # prepare new snapshot device
+    base_sz=$(blockdev --getsz "$base_dev")
+    3< "$base_dev" 4< "$cow_dev" dmsetup create "$dm_devname" \
+      --table "0 $base_sz snapshot /dev/fd/3 /dev/fd/4 P $SNAPSHOT_CHUNKSIZE" ||
```
marmarek, line 57: Using `/dev/fd` here doesn't help with anything (the device name is resolved by the shell just before calling `dmsetup`), but makes it harder to read. Please don't do that.

marmarek (changes requested):
Besides the comments below, some commits needs squashing into those they fix.

qubes/storage/file.py
```diff
@@ -24,20 +24,29 @@
...
 # in linux/drivers/md/dm-snap-persistent.c
 EMPTY_SNAPSHOT = b'SnAp\x01\x00\x00\x00\x01\x00\x00\x00\x00\x01\x00\x00' \
                  + bytes(262128)
 
+def _dev_path(file):
+    res = os.stat(file)
+    assert stat.S_ISREG(res.st_mode), 'not a regular file?'
+    return '%x:%d' % (res.st_dev, res.st_ino)
+
+_lock = asyncio.Lock()
```
marmarek, line 49: Missing `asyncio` import (see CI failures).

marmarek (changes requested):

marmarek (approved):

marmarek:
https://openqa.qubes-os.org/tests/20414#step/StorageFile/2:
```
  File "/usr/lib64/python3.8/contextlib.py", line 75, in inner
    return func(*args, **kwds)
  File "/usr/lib64/python3.8/unittest/case.py", line 636, in _callTearDown
    self.tearDown()
  File "/usr/lib/python3.8/site-packages/qubes/tests/integ/storage.py", line 54, in tearDown
    self.loop.run_until_complete(self.vm1.kill())
  File "/usr/lib64/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "/usr/lib/python3.8/site-packages/qubes/vm/qubesvm.py", line 1350, in kill
    await self._kill_locked()
  File "/usr/lib/python3.8/site-packages/qubes/vm/qubesvm.py", line 1366, in _kill_locked
    await self._ensure_shutdown_handled()
  File "/usr/lib/python3.8/site-packages/qubes/vm/qubesvm.py", line 1104, in _ensure_shutdown_handled
    await self.fire_event_async('domain-stopped')
  File "/usr/lib/python3.8/site-packages/qubes/events.py", line 233, in fire_event_async
    effect = task.result()
  File "/usr/lib/python3.8/site-packages/qubes/vm/qubesvm.py", line 1292, in on_domain_stopped
    await self.storage.stop()
  File "/usr/lib/python3.8/site-packages/qubes/storage/__init__.py", line 616, in stop
    await qubes.utils.void_coros_maybe(
  File "/usr/lib/python3.8/site-packages/qubes/utils.py", line 263, in void_coros_maybe
    task.result()  # re-raises exception if task failed
  File "/usr/lib/python3.8/site-packages/qubes/storage/file.py", line 411, in stop
    await self._destroy_blockdev()
  File "/usr/lib/python3.8/site-packages/qubes/storage/file.py", line 242, in _destroy_blockdev
    await qubes.utils.run_program(DESTROY_SCRIPT,
  File "/usr/lib/python3.8/site-packages/qubes/utils.py", line 249, in run_program
    raise subprocess.CalledProcessError(p.returncode,
subprocess.CalledProcessError: Command '/usr/lib/qubes/destroy-snapshot' returned non-zero exit status 1.
```
And the stderr contains:
```
device-mapper: remove ioctl on origin-fd03:930429  failed: Device or resource busy
```

DemiMarie:
Not exactly sure what the problem is, but I suspect that `destroy-snapshot` is trying to remove a `snapshot-origin` device that still has snapshots, and that the kernel is rejecting this.  Is there any way to confirm that hypothesis?

marmarek:
> Not exactly sure what the problem is, but I suspect that `destroy-snapshot` is trying to remove a `snapshot-origin` device that still has snapshots, and that the kernel is rejecting this. Is there any way to confirm that hypothesis?

I think removing snapshot-origin is legal even when there are still snapshots (but worth checking, not 100% sure). Anyway, you can try to reproduce it by following the steps this very test does. The use of kill instead of clean shutdown may be important here.

marmarek:
There is [another failure](https://openqa.qubes-os.org/tests/20453#step/TC_06_AppVM_fedora-34-pool/1):
```
  File "/usr/lib64/python3.8/contextlib.py", line 75, in inner
    return func(*args, **kwds)
  File "/usr/lib64/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/usr/lib/python3.8/site-packages/qubes/tests/integ/basic.py", line 624, in test_121_start_standalone_with_cdrom_vm
    self.loop.run_until_complete(self.vm.create_on_disk())
  File "/usr/lib64/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "/usr/lib/python3.8/site-packages/qubes/vm/qubesvm.py", line 1707, in create_on_disk
    await self.storage.create()
  File "/usr/lib/python3.8/site-packages/qubes/storage/__init__.py", line 534, in create
    await qubes.utils.void_coros_maybe(
  File "/usr/lib/python3.8/site-packages/qubes/utils.py", line 259, in void_coros_maybe
    coros = [val for val in values if asyncio.iscoroutine(val)]
  File "/usr/lib/python3.8/site-packages/qubes/utils.py", line 259, in <listcomp>
    coros = [val for val in values if asyncio.iscoroutine(val)]
  File "/usr/lib/python3.8/site-packages/qubes/storage/__init__.py", line 535, in <genexpr>
    vol.create() for vol in self.vm.volumes.values())
  File "/usr/lib/python3.8/site-packages/qubes/storage/file.py", line 236, in create
    create_sparse_file(self.path, self.size, permissions=0o664)
  File "/usr/lib/python3.8/site-packages/qubes/storage/file.py", line 509, in create_sparse_file
    raise IOError("Volume %s already exists" % path)
OSError: Volume /var/lib/qubes-pool/appvms/test-inst-appvm/volatile.img already exists
```
I think it's related to this change:
```patch
     async def remove(self):
-        await self.stop()
+        await self._destroy_blockdev()
```

marmarek:
>     * StorageFile: [test_001_non_volatile](https://openqa.qubes-os.org/tests/20441#step/StorageFile/2)

And one more:
```
  File "/usr/lib64/python3.8/contextlib.py", line 75, in inner
    return func(*args, **kwds)
AssertionError: nested calls to start()
```
This could be also a missing `await` somewhere?

DemiMarie:
I will need to take a look at this later.

marmarek:
PipelineRetry

qubes/storage/file.py
```diff
@@ -344,24 +375,42 @@ def start(self):  # pylint: disable=invalid-overridden-method
...
         return self
 
-    def stop(self):  # pylint: disable=invalid-overridden-method
+    async def stop(self):
         assert self._export_lock is not FileVolume._marker_exported, \
             'trying to stop exported file volume?'
+        if self.save_on_stop or self.snap_on_start:
+            await self._destroy_blockdev()
         if self.save_on_stop:
+            assert self.rw, 'Attempted to commit read-only volume'
```
marmarek, line 413: While usefulness of `rw=False` + `save_on_stop=True` configuration is dubious, it is a valid configuration. Especially since the user can change the `rw` property but cannot change `save_on_stop`.
Maybe in the `not self.rw` case assert that `self.path_cow` is empty (maybe with some exception for dm-snapshot header/metadata - if anything like that exists?) and then remove it?

DemiMarie: What does the code do now?

marmarek: Currently it refuses to stop `not rw and save_on_stop` volume at all. In fact, I think the old code had the same issue, but the tests were missing testing that (and you fixed this part of the test).

DemiMarie: That explains why the CI is failing!

PR #400, opened 2021-04-19
WIP: Loop device pool

DemiMarie:
This is woefully incomplete.  I am mostly asking for feedback on the design at this point, in particular on the split of work between Python and C and on device lifetime management.  The latter will likely require somehow synchronizing with udev, so that qubesd knows when a loop device has gone away.


DemiMarie:
Closing as the file pool is going away and loop devices for other pools will not be pooled.

PR #403, opened 2021-05-12
Remove some uses of @asyncio.coroutine

DemiMarie:
It is deprecated and will be removed in Python 3.10.

### DIFF
--- a/doc/qubes-events.rst
+++ b/doc/qubes-events.rst
@@ -152,9 +152,8 @@ Asynchronous event handling
 Event handlers can be defined as coroutine. This way they can execute long
 running actions without blocking the whole qubesd process. To define
 asynchronous event handler, annotate a coroutine (a function defined with
-`async def`, or decorated with :py:func:`asyncio.coroutine`) with
-:py:func:`qubes.events.handler` decorator. By definition, order of
-such handlers is undefined.
+`async def`) with :py:func:`qubes.events.handler` decorator. By definition,
+order of such handlers is undefined.
 
 Asynchronous events can be fired using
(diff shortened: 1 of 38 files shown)


doc/qubes-events.rst
```diff
@@ -170,11 +169,10 @@ handler (a coroutine) for synchronous event (the one fired with
 
    class MyClass(qubes.events.Emitter):
        @qubes.events.handler('event1', 'event2')
-       @asyncio.coroutine
        def event_handler(self, event):
```
marmarek, line 172:
```suggestion
       async def event_handler(self, event):
```

doc/qubes-events.rst
```diff
@@ -192,17 +190,15 @@ yield individual values (because of python limitation):
 
    class MyClass(qubes.events.Emitter):
        @qubes.events.handler('event1')
-       @asyncio.coroutine
-       def event_handler(self, event):
+       async def event_handler(self, event):
             print('Got event, starting long running action')
-            yield from asyncio.sleep(10)
+            asyncio.sleep(10)
```
marmarek, line 195:
```suggestion
            await asyncio.sleep(10)
```

marmarek (changes requested):
Those "Remove more uses of @asyncio.coroutine" commits are not very useful separately, especially since some changes are split across multiple of them (removing `@asyncio.coroutine` in one, but changing to `async def` in another), and some are reverted too. Please squash them. Other changes in this PR may (or rather: should) stay in separate commits.

marmarek (approved):

PR #404, opened 2021-05-12
Update lvm.py

brendanhoar:
Break the ioctl calls from blkdiscard into max 1GB worth of blocks per call. This ensures that the amount of pending discards (both for the virtual lvm thin volume as well as any lower level device recipients) do not dominate the IO streams to the devices. Before this change, IO blocking was leading to both UI and processing freezes in Qubes while very large LVM volumes were being removed. The change will allow other device IOs to be interleaved between the internal ioctl calls from blkdiscard.


brendanhoar:
I believe I've replaced the original commit with a **signed** replacement based on the instructions here: https://www.qubes-os.org/doc/code-signing/

DemiMarie (approved):
Looks good, thanks!

marmarek (approved):

PR #405, opened 2021-05-13
Test that quotes and ampersand are XML escaped

DemiMarie:
This ensures that we are safe if we interpolate into an XML attribute.

### DIFF
--- a/qubes/tests/vm/qubesvm.py
+++ b/qubes/tests/vm/qubesvm.py
@@ -961,7 +961,7 @@ class TC_90_QubesVM(QubesVMTestsMixin, qubes.tests.QubesTestCase):
             <loader type="rom">hvmloader</loader>
             <boot dev="cdrom" />
             <boot dev="hd" />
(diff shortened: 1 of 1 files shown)


qubes/tests/vm/qubesvm.py
```diff
@@ -961,7 +961,7 @@ def test_600_libvirt_xml_hvm_dom0_kernel_kernelopts(self):
             <loader type="rom">hvmloader</loader>
             <boot dev="cdrom" />
             <boot dev="hd" />
-            <cmdline>kernel &lt;text&gt; specific options</cmdline>
+            <cmdline>kernel &lt;text&gt;&#39;&#34;&amp specific options</cmdline>
```
marmarek, line 964:
```suggestion
            <cmdline>kernel &lt;text&gt;&#39;&#34;&amp; specific options</cmdline>
```

DemiMarie: oops…

PR #406, opened 2021-05-17
restart automatically started vm on failures

jevank:
This allow to workaround errors like [QI#6203](https://github.com/QubesOS/qubes-issues/issues/6203)

### DIFF
--- a/linux/systemd/qubes-vm@.service
+++ b/linux/systemd/qubes-vm@.service
@@ -9,6 +9,7 @@ Environment=DISPLAY=:0
 ExecStart=/usr/bin/qvm-start --skip-if-running %i
 Group=qubes
 RemainAfterExit=yes
+Restart=on-failure
 
 [Install]
 WantedBy=multi-user.target


linux/systemd/qubes-vm@.service
```diff
@@ -9,6 +9,7 @@ Environment=DISPLAY=:0
 ExecStart=/usr/bin/qvm-start --skip-if-running %i
 Group=qubes
 RemainAfterExit=yes
+Restart=on-failure
```
marmarek, line 12: This may loop forever, if the failure reason is persistent (like configuration error).

jevank: We can set additional configuration.
> StartLimitIntervalSec=interval, StartLimitBurst=burst
>    Configure unit start rate limiting. Units which are started more than burst times within an interval time span are not permitted to start any more. Use StartLimitIntervalSec= to configure the checking interval and StartLimitBurst= to configure how many starts per interval are allowed.
>
>    _interval_ is a time span with the default unit of seconds, but other units may be specified, see systemd.time(5). Defaults to DefaultStartLimitIntervalSec= in manager configuration file, and may be set to 0 to disable any kind of rate limiting. _burst_ is a number and defaults to DefaultStartLimitBurst= in manager configuration file.
>
>    These configuration options are particularly useful in conjunction with the service setting Restart= (see systemd.service(5)); however, they apply to all kinds of starts (including manual), not just those triggered by the Restart= logic.
>
>    Note that units which are configured for Restart=, and which reach the start limit are not attempted to be restarted anymore; however, they may still be restarted manually or from a timer or socket at a later point, after the interval has passed. From that point on, the restart logic is activated again. systemctl reset-failed will cause the restart rate counter for a service to be flushed, which is useful if the administrator wants to manually start a unit and the start limit interferes with that. Rate-limiting is enforced after any unit condition checks are executed, and hence unit activations with failing conditions do not count towards the rate limit.

I see Fedora defaults is 10 seconds and 5 bursts. What about 60s and 5 bursts for qubes-vm@ unit?

marmarek: One of the failure modes is startup timeout (60s). With 60s interval, it will never get limited in such case. But with increasing limit to 300s, you get 5 minutes of system doing something, when you will have hard time to intervene.
IMO a better workaround is the one proposed in https://github.com/OpenXT/xenclient-oe/commit/f60e3a0b4ea0cdb3f354600302d34c918352439f

jevank: Oh, I got you, qrexec_timeout default value is 60s so its possible to get infinitive restarting. In that case looks better setting interval 200s and burst 3, but I don`t insist. We are facing a problem with sys-usb failed on boot about one time in 10  with symptoms like QI#6203 so I thought to share the idea.
As for the patch from OpenXT, I was trying, but it require completely rewritten because of async qmp communication in latest libxl.

PR #407, opened 2021-05-18
Partial fix for PCI devices reordering. See https://github.com/QubesO…

v6ak:
…S/qubes-issues/issues/6587 .

### DIFF
--- a/qubes/devices.py
+++ b/qubes/devices.py
@@ -56,6 +56,8 @@ Extension may use QubesDB watch API (QubesVM.watch_qdb_path(path), then handle
 `device-list-change:class` event.
 '''
 import asyncio
+from collections import OrderedDict
+from typing import Optional
 
 import qubes.utils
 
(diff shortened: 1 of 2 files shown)


marmarek:
>     * They are rather conservative – I haven't touched the original method assignments, but rather copied and created a new one. It doesn't seem to be hard to change other occurrences, but I am not sure about testing.

I think it should be ok to replace original `assignments` method.

> PCI devices seem to reorder less frequently. I'd like to investigate this deeper, but this is all I know ATM.

If you replace the original `assignments` method, it will be used in libvirt xml template too (`templates/libvirt/xen.xml`) - which I guess is another source of reordering.

v6ak:
Aha, good point. So, IIUC, create_config_file generates randomized XML and it is cached in `/var/lib/xen/userdata-d.*` files. So, it can survive some reboots, and thus it is reordered far less often than in 50 % cases. Can I somehow speed the testing up, e.g. by `sudo rm /var/lib/xen/userdata-d.*`? (Is it safe?)

DemiMarie:
@marmarek can we get this going again?  It would be a really nice fix for some problems I have had.

marmarek:
@v6ak can you rebase this PR on master, and sign your commits?

PR #411, opened 2021-06-09
DispVMs cannot be DispVM templates

DemiMarie:
This avoids a traceback from qubesd if one tries to use a DispVM as the
template for a DispVM.

Fixes QubesOS/qubes-issues#6685.

### DIFF
--- a/qubes/vm/dispvm.py
+++ b/qubes/vm/dispvm.py
@@ -201,7 +201,7 @@ class DispVM(qubes.vm.qubesvm.QubesVM):
         This method modifies :file:`qubes.xml` file.
         The qube returned is not started.
         '''
-        if not appvm.template_for_dispvms:
+        if not getattr(appvm, 'template_for_dispvms', False):
             raise qubes.exc.QubesException(
                 'Refusing to create DispVM out of this AppVM, because '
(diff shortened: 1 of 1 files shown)


qubes/vm/dispvm.py
```diff
@@ -41,6 +41,8 @@ class DispVM(qubes.vm.qubesvm.QubesVM):
                                 setter=_setter_template,
                                 doc='AppVM, on which this DispVM is based.')
 
+    template_for_dispvms = False
```
marmarek, line 44: I'd rather replace accessing the property with `getattr(vm, `template_for_dispvm`, False)`. In fact most of the places are this way already, there are just few that are not.

marmarek (changes requested):

qubes/vm/dispvm.py
DemiMarie: What is the disadvantage of my approach?  Serious question.  Using `getattr` might cause difficulties if and when we add static types to our code.

DemiMarie: Nevermind, I forgot that `template_for_dispvms` is no ordinary Python attribute.

marmarek (approved):

PR #417, opened 2021-08-08
Fix timezone handling in HVMs

DemiMarie:
Non-Windows HVMs always want the “hardware” (emulated) clock to be UTC,
and Windows HVMs can be made to expect this by a registry change.

Tested by setting the `timezone` feature to the empty string in the template of `sys-net`, which has the same effect as this change.

### DIFF
--- a/templates/libvirt/xen.xml
+++ b/templates/libvirt/xen.xml
@@ -76,7 +76,7 @@
 
     {% block clock %}
         {% if vm.virt_mode == 'hvm' %}
-            {% set timezone = vm.features.check_with_template('timezone', 'localtime').lower() %}
+            {% set timezone = vm.features.check_with_template('timezone', '').lower() %}
             {% if timezone == 'localtime' %}
                 <clock offset="variable" adjustment="0" basis="localtime" />
             {% elif timezone.isdigit() %}
--- a/qubes/tests/vm/qubesvm.py
+++ b/qubes/tests/vm/qubesvm.py
@@ -848,7 +848,7 @@ class TC_90_QubesVM(QubesVMTestsMixin, qubes.tests.QubesTestCase):
             <apic/>
             <viridian/>
         </features>
-        <clock offset="variable" adjustment="0" basis="localtime" />
+        <clock offset="variable" adjustment="0" basis="utc" />
         <on_poweroff>destroy</on_poweroff>
         <on_reboot>destroy</on_reboot>
         <on_crash>destroy</on_crash>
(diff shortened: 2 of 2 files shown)


templates/libvirt/xen.xml
```diff
@@ -76,14 +76,7 @@
...
         {% if vm.virt_mode == 'hvm' %}
-            {% set timezone = vm.features.check_with_template('timezone', 'localtime').lower() %}
-            {% if timezone == 'localtime' %}
-                <clock offset="variable" adjustment="0" basis="localtime" />
-            {% elif timezone.isdigit() %}
-                <clock offset="variable" adjustment="{{ timezone }}" basis="utc" />
-            {% else %}
-                <clock offset="variable" adjustment="0" basis="utc" />
-            {% endif %}
+            <clock offset="variable" adjustment="0" basis="utc" />
```
marmarek, line 79: No, this basically removes the timezone setting feature (which is here to make Windows happy). I'd rather prefer changing the default to "0" or similar

marmarek (changes requested):

templates/libvirt/xen.xml
marmarek: This would work only if the `os` feature is set - which is done only when QWT are installed inside. This is not always the case, especially given their state. Additionally, this would mean changing the clock behavior by just installing QWT. In fact, this feature is more useful for Windows _without_ QWT, as otherwise we could configure it to expect UTC. And finally, there exist other OSes, that may or may not need this feature - having it conditionally supported only for Windows only will be confusing.

marmarek: > My worry is that having the clock be in local time _will_ lead to breakage on Linux and *BSD. 

That's why I propose to change the default. And keep the possibility to set it manually for Windows to something else (like `locatime`).

> Furthermore, I am worried that the `timezone` feature could wind up in QubesDB, 

No, that value is taken from dom0's timezone (`/etc/localtime`).

> so a value of `0` would cause confusion.

`utc` then?

DemiMarie: It’s also worth noting that we might be able to set up our (emulated) firmware such that QWT is (by default) automatically installed into _any_ Windows qube created on the system, or at least Windows is told to expect the hardware clock to be UTC.

marmarek: Maybe, or maybe not. In any case, that's totally out of scope for this fix.

marmarek (approved):

PR #416, opened 2021-08-05
'async def' fixes

rustybird:
reflink looks broken in master since c74a03a3 (luckily no version bump yet - ⚠️ @marmarek). The first commit in the series should unbreak it.

Marked as draft because it'll take me a few days until I can actually test anything on R4.1.

PR #418, opened 2021-08-08
tests: detect never awaited coroutine objects

rustybird:


rustybird:
The remaining test failures are already fixed as part of #416.

rustybird:
`warnings.catch_warnings()` wasn't quite right, so the latest force-pushed commits retvrn to tradition (overriding `warnings.showwarning()`).

Also added some benchmarks to the commit message.

marmarek (approved):

rustybird:
Gah. I thought I had checked that the decorator didn't mess up tracebacks for unrelated test failures, but what I had actually checked were tracebacks for test _errors_ (from raising something that's not an AssertionError).

-> #421

PR #420, opened 2021-08-14
Add missing await in clone_disk_files() and other fixes

marmarek:


marmarek:
Thanks @rustybird for https://github.com/QubesOS/qubes-core-admin/pull/418 !

DemiMarie:
Could this be the reason for #399 having problems?

marmarek:
> Could this be the reason for #399 having problems?

The most recent run was with this already included.

DemiMarie:
> > Could this be the reason for #399 having problems?
> 
> The most recent run was with this already included.

Gotcha, thanks!

marmarek:
> The most recent run was with this already included.

Hint: if you click into the openQA job's settings, there is `PULL_REQUESTS` option, with a list of included PRs. It is not easily clickable (the whole list is one big link :/) but it's useful anyway.

PR #426, opened 2021-09-06
tests: add test success detector during tearDown()

deeplow:
**Context:** there are non-easily-reproducible test failures in the network tests (ping failed). This fixes the extra logging that should happen upon failure so hopefully when the issue happens again we can detect it.

---

Previously the test success detection during `tearDown()` didn't work as it depended on self._outcome which had not yet received the correct success state of the test just ran.

See a POC of this solution [in this gist](https://gist.github.com/deeplow/143a5636ba16ea5245314b46bc54793e).

The solution essentially does X things:

  1. Runs the code that would otherwise be run after the tearDown
     see:https://github.com/python/cpython/blob/7e246a3/Lib/unittest/case.py#L601

  2. Interprets the results following the internal structure of
     `unittest.TestResult` and returns the respective success state

Solution inspired by the following exchange:
https://stackoverflow.com/questions/4414234/getting-pythons-unittest-results-in-a-teardown-method#answer-39606065

---

Since this method should only be ran from inside `tearDown` (for fear of the results not being set or already having been erased) I wanted to through an exception when that was the case:

```python
        if "tearDown" != sys._getframe(1).f_code.co_name:
            raise Exception("success() can only be called from tearDown")
```

But I am not satisfied with the above code as it uses some low-level functions that should be only used for internal or specialized purposes... So I did not include it. Instead the method's docstring includes that disclaimer.

### DIFF
--- a/qubes/tests/__init__.py
+++ b/qubes/tests/__init__.py
@@ -537,6 +537,15 @@ class QubesTestCase(unittest.TestCase):
 
         del self.loop
 
+    def success(self):
+        """Check if test was successful during tearDown """
+
+        result = self.defaultTestResult()
+        self._feedErrorsToResult(result, self._outcome.errors)
+
+        unsuccessful_tests = [test for (test,_) in (result.errors + result.failures)]
+        return self not in unsuccessful_tests
+
     def assertNotRaises(self, excClass, callableObj=None, *args, **kwargs):
         """Fail if an exception of class excClass is raised
            by callableObj when invoked with arguments args and keyword
--- a/qubes/tests/integ/network.py
+++ b/qubes/tests/integ/network.py
@@ -98,7 +98,7 @@ class VmNetworkingMixin(object):
 
     def tearDown(self):
         # collect more info on failure
-        if self._outcome and not self._outcome.success:
+        if not self.success():
             for vm in (self.testnetvm, self.testvm1, getattr(self, 'proxy', None)):
                 if vm is None:
                     continue


deeplow:
Weird. How does the coverage increase anything with this?

marmarek:
> Weird. How does the coverage increase anything with this?

Weird indeed. If you look at the detailed report, it makes even less sense: https://app.codecov.io/gh/QubesOS/qubes-core-admin/compare/426/changes ...

deeplow:
Didn't seem to break anything. But also none of the network tests failed, so there wasn't really a chance for the code to be executed.

marmarek:
I'm not sure why this is necessary, Looking at https://github.com/python/cpython/blob/7e246a3/Lib/unittest/case.py#L586-L593, the `self._outcome` should be set already, and its [`testPartExecutor()` method](https://github.com/python/cpython/blob/7e246a3a7b43762480ee4fe0cfb859e8e997a8c8/Lib/unittest/case.py#L72) does set the "success" property. Are you sure this is the reason why the messages weren't logged?

In recent tests I've added `--log-level debug` to `nose2` call, which produces a lot of output. This may make the tests a bit slower... Maybe that's why it doesn't fail now?

deeplow:
> I'm not sure why this is necessary, Looking at https://github.com/python/cpython/blob/7e246a3/Lib/unittest/case.py#L586-L593, the `self._outcome` should be set already, and its [`testPartExecutor()` method](https://github.com/python/cpython/blob/7e246a3a7b43762480ee4fe0cfb859e8e997a8c8/Lib/unittest/case.py#L72) does set the "success" property. Are you sure this is the reason why the messages weren't logged?

I was also confused by that. I didn't fully get to the bottom of why exactly `self._outcome.success` was not set on the teardown. I followed through with `pudb` but got lost. I concluded it had probably to do with the context manager for the outcomes in the lines you highlighted. It set the `success` on the `with` context but when it left the function that outcome was not set on the object's `_outcome`.

But the POC demonstrates this and shows a working solution: https://gist.github.com/deeplow/143a5636ba16ea5245314b46bc54793e#file-unittest-detect-fail-in-teardown-py

If you're not convinced, I can try to understand exactly why this is. But I didn't think it justfied the effort. Let me know otherwise.
 
> In recent tests I've added `--log-level debug` to `nose2` call, which produces a lot of output. This may make the tests a bit slower... Maybe that's why it doesn't fail now?

Oh, no. It didn't fail just because of the instability. In order for this code to run it needs to fail. However in the job above none of the network tests happened to fail. I think it din't affect the likelyhood of failure of tests (and shouldn't).

marmarek:
In https://openqa.qubes-os.org/tests/24040#step/VmIPv6Networking_debian-11/14 this did fixed the logging (it can be found in journalctl), although I'm still not sure about the failure reason.

PR #428, opened 2021-09-29
Make encrypted volatile volume optional

marmarek:
Make 'ephemeral' volume property accessible read-write via Admin API, and set
the default value to False. This allows the feature to be enabled for specific
VMs where it is needed. This can be changed via `qvm-volume` tool.
Furthermore, volumes of a new DispVM gets the value from its template's
matching volumes. Do the same for 'rw' property.

QubesOS/qubes-issues#904

### DIFF
--- a/qubes/storage/__init__.py
+++ b/qubes/storage/__init__.py
@@ -84,7 +84,7 @@ class Volume:
 
     def __init__(self, name, pool, vid,
             revisions_to_keep=0, rw=False, save_on_stop=False, size=0,
-            snap_on_start=False, source=None, **kwargs):
+            snap_on_start=False, source=None, ephemeral=None, **kwargs):
         ''' Initialize a volume.
 
             :param str name: The name of the volume inside owning domain
@@ -98,6 +98,7 @@ class Volume:
                 vm.stop(), otherwise - discard
             :param Volume source: other volume in same pool to make snapshot
                 from, required if *snap_on_start*=`True`
+            :param ephemeral: encrypt volume with an ephemeral key
             :param str/int size: Size of the volume
 
         '''
@@ -129,6 +130,9 @@ class Volume:
         #: Should volume state be saved or discarded at :py:meth:`stop`
         self.save_on_stop = save_on_stop
         self._size = int(size)
+        #: Should the volume be encrypted with an ephemeral key;
+        #  None means the default value
+        self._ephemeral = ephemeral
         #: Should the volume state be initialized with a snapshot of
         #: same-named volume of domain's template.
         self.snap_on_start = snap_on_start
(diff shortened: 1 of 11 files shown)


qubes/storage/__init__.py
```diff
@@ -162,10 +166,25 @@ def __xml__(self):
...
+
+    @ephemeral.setter
+    def ephemeral(self, value):
+        if not value:
+            self._ephemeral = False
+            return
+        if self.snap_on_start or self.save_on_stop or \
+                self.domain is not None or not self.rw:
+            raise ValueError(
+                'Cannot enable ephemeral on snap_on_star or save_on_stop or '
```
DemiMarie, line 185:
```suggestion
                'Cannot enable ephemeral on snap_on_start or save_on_stop or '
```

qubes/storage/__init__.py
```diff
@@ -162,10 +166,25 @@ def __xml__(self):
...
+        # default value
+        return False
+
+    @ephemeral.setter
+    def ephemeral(self, value):
+        if not value:
+            self._ephemeral = False
+            return
+        if self.snap_on_start or self.save_on_stop or \
+                self.domain is not None or not self.rw:
```
DemiMarie, line 183: Will that prevent DispVM templates from starting?

qubes/tests/vm/dispvm.py
```diff
@@ -320,3 +321,24 @@ def test_022_storage_app_change(self):
...
+                as mock_domains:
+            mock_domains.configure_mock(**{
+                'get_new_unused_dispid': mock.Mock(return_value=42),
+                '__getitem__.side_effect': orig_getitem
+            })
+            dispvm = self.app.add_new_vm(qubes.vm.dispvm.DispVM,
+                name='test-dispvm', template=self.appvm)
+            self.loop.run_until_complete(dispvm.create_on_disk())
+        self.assertIs(dispvm.template, self.appvm)
+        self.assertTrue(dispvm.volumes['volatile'].ephemeral)
```
DemiMarie, line 344:
```suggestion
        self.assertTrue(dispvm.volumes['volatile'].ephemeral)
```

qubes/storage/__init__.py
marmarek: It should not. It's about `volatile` volume only, which (currently) always have the same properties regardless of qube type. Furthermore, this is done on setting the property, not starting the qube.

DemiMarie: Ah, good catch :woman_facepalming:

qubes/tests/vm/dispvm.py
marmarek: Where is the difference?

DemiMarie: Missing newline at EOF

marmarek:
> Instead of the `ephemeral` property for volatile volumes defaulting to `False` unconditionally, I strongly believe this should be configurable system-wide via `qubes-prefs` or the Qubes Global Settings GUI. 

I don't exclude adding such option later. But as discussed before, the use case (besides DispVM) is rather niche, so I'm not going to spend too much time on this now (delaying R4.1 even more).

> It should also be configurable on TemplateVMs.

For DispVMs, I specifically added inheriting the setting from their templates. For others: see above.

PR #431, opened 2021-10-08
tests/network: avoid race conditions with NetworkManager

marmarek:
The "netvm" in the tests is configured manually, using "ip" command, and
by manually writing to /etc/resolv.conf. Do not let NetworkManager
interfere - simply disable it.


marmarek:
@deeplow this is what happened in debian-11 network tests. I finally found it by decorating all the tests with `@qubes.tests.wait_on_fail` and running them in a loop. I didn't got extra logs, but thanks to the decorator I could interact with broken VMs. In `test-inst-netvm1` I've found `/etc/resolv.conf` with just one line: `# Generated by NetworkManager` (and no actual DNS servers).

PR #432, opened 2021-10-13
Add pool-level ephemeral property

DemiMarie:
This allows pool-wide control over whether volatile volumes are encrypted with an ephemeral key.

Fixes QubesOS/qubes-issues#6958.


DemiMarie:
This still needs the test for saving and reloading from `qubes.xml`.  Are there any existing similar tests?

PR #441, opened 2021-12-14
improve QWT advertise handling

jevank:
This enables Windows supported features:
 - USB support
 - audio support
 - timezone fix

### DIFF
--- a/qubes/ext/windows.py
+++ b/qubes/ext/windows.py
@@ -51,13 +51,14 @@ class WindowsFeatures(qubes.ext.Extension):
             vm.features['rpc-clipboard'] = True
             setattr(vm, 'maxmem', 0)
             setattr(vm, 'qrexec_timeout', 6000)
-            if not vm.features.check_with_template('stubdom-qrexec', False):
+            if vm.features.check_with_template('stubdom-qrexec', None) is None:
                 vm.features['stubdom-qrexec'] = True
-            if not vm.features.check_with_template('audio-model', False):
+            if vm.features.check_with_template('audio-model', None) is None:
                 vm.features['audio-model'] = 'ich6'
-            if not vm.features.check_with_template('timezone', False):
+            if vm.features.check_with_template('timezone', None) is None:
                 vm.features['timezone'] = 'localtime'
-            if not vm.features.check_with_template('no-monitor-layout', False):
+            if vm.features.check_with_template('no-monitor-layout',
+                                               None) is None:
                 vm.features['no-monitor-layout'] = True
 
     @qubes.ext.handler('domain-create-on-disk')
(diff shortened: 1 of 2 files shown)


qubes/ext/windows.py
```diff
@@ -49,6 +49,16 @@ def qubes_features_request(self, vm, event, untrusted_features):
             vm.features['os'] = guest_os
         if guest_os == 'Windows' and qrexec:
             vm.features['rpc-clipboard'] = True
+            setattr(vm, 'maxmem', 0)
+            setattr(vm, 'qrexec_timeout', 6000)
```
marmarek, line 53: I realize 1min may be too little, but over an hour isn't excessive?

jevank: Booting after system updates can be time consuming. 

qvm-create-windows-qube default value is 999999
https://github.com/elliotkillick/qvm-create-windows-qube/blob/master/qvm-create-windows-qube#L284

DemiMarie: In that case we should fix QubesOS/qubes-issues#7116.

jevank: perhaps it makes sense to disable this timeout (value 0, for example), since even 6000 does not guarantee anything and at the same time does not solve the problems of a hung VM?

marmarek: > In that case we should fix https://github.com/QubesOS/qubes-issues/issues/7116.

Since the above is fixed, I think we can go ahead with merging it.
Later we can adjust to other versions, like no timeout at all. But for that, I'd prefer  better handling for it in various tools (for example `qvm-start` would basically hang).

marmarek:
PipelineRetry

marmarek:
a test needs an update:
```
FAIL: qubes.tests.ext/TC_10_WindowsFeatures/test_000_notify_tools_full
  File "/home/gitlab-runner/builds/QubesOS/qubes-core-admin/qubes/tests/ext.py", line 220, in test_000_notify_tools_full
    self.assertEqual(self.vm.mock_calls, [
AssertionError: [call.features.__setitem__('os', 'Windows'[482 chars]__()] != [('features.__setitem__', ('os', 'Windows'[57 chars] {})]
```
it basically needs duplicating what your change does...

jevank:
I didn't verify, but looks obvious..

jevank:
Not so obvious.

marmarek:
Ouch, I think it wants all the `check_with_template` calls too. This is getting ridiculous. Let me change the tests to something less insane.

jevank:
Oh, thanks, but I realized that I need to deploy the environment to run the tests

marmarek:
For those unit tests, you can run `run-tests` script directly from sources, in a VM.

marmarek (approved):
Approved as of 76e5b63262906ccd8f3bcba6931b05cd24af0a3b

PR #442, opened 2021-12-16
Support nanosecond precision for dates

DemiMarie:
Along with several other useful changes.  See individual commit messages for details.

### DIFF
--- a/qvm-tools/qvm-sync-clock
+++ b/qvm-tools/qvm-sync-clock
@@ -41,8 +41,8 @@ def main():
             clockvm.name))
         sys.exit(0)
 
-    p = clockvm.run_service('qubes.GetDate+nanoseconds')
-    untrusted_date_out = p.stdout.read(36)
+    with clockvm.run_service('qubes.GetDate+nanoseconds') as p:
+        untrusted_date_out = p.stdout.read(36)
     try:
         untrusted_date_out.decode('ascii', 'strict')
     except UnicodeDecodeError:
--- a/qvm-tools/qvm-sync-clock
+++ b/qvm-tools/qvm-sync-clock
@@ -26,6 +26,7 @@ import subprocess
 from qubesadmin import Qubes
 
 def main():
+    os.environ['LC_ALL'] = 'C'
     if os.geteuid() != 0:
         sys.stderr.write('This program must be run as root to set the date, aborting!\n')
         sys.exit(1)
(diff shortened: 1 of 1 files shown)


marmarek:
> * TC_00_AppVM_debian-11: [test_210_time_sync](https://openqa.qubes-os.org/tests/26800#step/TC_00_AppVM_debian-11/13)
> 
> * TC_00_AppVM_fedora-34: [test_210_time_sync](https://openqa.qubes-os.org/tests/26800#step/TC_00_AppVM_fedora-34/13)

It doesn't work.
Anyway, it is an API change, and at this point it's too late to include it in R4.1.

DemiMarie:
> > * TC_00_AppVM_debian-11: [test_210_time_sync](https://openqa.qubes-os.org/tests/26800#step/TC_00_AppVM_debian-11/13)
> > * TC_00_AppVM_fedora-34: [test_210_time_sync](https://openqa.qubes-os.org/tests/26800#step/TC_00_AppVM_fedora-34/13)
> 
> It doesn't work. Anyway, it is an API change, and at this point it's too late to include it in R4.1.

:woman_facepalming: Silly syntax error (extra colon on line 45, now fixed), but the point about the API change stands.

DemiMarie:
@marmarek worth merging to `master` for R4.2?

marmarek:
I'm confused by recent force-push, which undoes change requested in https://github.com/QubesOS/qubes-core-agent-linux/pull/347

DemiMarie:
> I'm confused by recent force-push, which undoes change requested in [QubesOS/qubes-core-agent-linux#347](https://github.com/QubesOS/qubes-core-agent-linux/pull/347)

The advantage of a number is that it makes future evolution easier: dom0 passes the version it supports, and the ClockVM replies with data corresponding to the version it understands.  Is this worthwhile @marmarek?

marmarek:
I don't think you need full ordering for this. `qubes.GetDate` service either knows requested format - and then responds with it, or it doesn't - and then responds with latest it knows. So, you can get the same with descriptive argument.

qvm-tools/qvm-sync-clock
```diff
@@ -40,19 +41,45 @@ def main():
...
 
-    p = clockvm.run_service('qubes.GetDate')
-    untrusted_date_out = p.stdout.read(25).decode('ascii', errors='strict')
-    untrusted_date_out = untrusted_date_out.strip()
-
-    if not re.match(r'^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\+00:?00$', untrusted_date_out):
+    with clockvm.run_service('qubes.GetDate+nanoseconds') as p:
+        untrusted_date_out = p.stdout.read(36)
+    try:
+        untrusted_date_out.decode('ascii', 'strict')
```
marmarek, line 47: Please save result of this decode and operate on str further down. It makes sense to use `bytes` only for binary data (that can't be converted to `str`) or if you are not going to convert it to `str` at all. Otherwise `str` is preferred.

marmarek (changes requested):

qvm-tools/qvm-sync-clock
DemiMarie: Saving the result and passing it to the `date` command later on makes sense, but I am curious why the validation should be done on the `str` rather than on the `bytes`.  All of the needed operations work on both, and I suspect the `bytes` regex code is probably a bit simpler (less attack surface) than its `str` counterpart.  I could also just skip this, on the grounds that a non-ASCII date will fail to match any of the regexps and will thus rejected.

marmarek: > Saving the result and passing it to the `date` command later on makes sense, but I am curious why the validation should be done on the `str` rather than on the `bytes`.

Validation should be done on the thing you will actually use - so, since you'll need str, you should validate that str.
Theoretically you could do validation on bytes and convert only at the end, but there is no point in doing this (and surely forgetting some `b` at some point in the future).

DemiMarie: I decided to go with a different approach: take advantage of `subprocess` supporting `bytes` for command-line arguments.  The decode is now purely an input validation step; its output is not used.

DemiMarie:
PipelineRetry

marmarek (approved):

PR #445, opened 2021-12-22
Remove `logrotate` configuration for no longer present `qmemman.log`

Rot127:
This PR fixes https://github.com/QubesOS/qubes-issues/issues/7132.

`logrotate` tried to access the `qmemman.log` file although it is no longer present. Hence it failed to start on boot.

Since `qmemman.log` is no longer in use, the configuration for it should be removed as well.

### DIFF
--- a/linux/system-config/Makefile
+++ b/linux/system-config/Makefile
@@ -6,6 +6,3 @@ install:
 	cp vif-route-qubes $(DESTDIR)/etc/xen/scripts
 	install -m 0755 create-snapshot destroy-snapshot $(DESTDIR)/usr/lib/qubes
 	install -m 0644 -D tmpfiles-qubes.conf $(DESTDIR)/usr/lib/tmpfiles.d/qubes.conf
-	install -d $(DESTDIR)/etc/logrotate.d
-	install -m 0644 logrotate-qubes \
-		$(DESTDIR)/etc/logrotate.d/qubes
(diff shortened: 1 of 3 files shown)


rustybird:
Thanks!

Looks like this file could even be deleted (and dropped from the `Makefile` and the `.spec.in`), now that it's empty?

linux/system-config/Makefile
```diff
@@ -7,5 +7,3 @@ install:
 	install -m 0755 create-snapshot destroy-snapshot $(DESTDIR)/usr/lib/qubes
 	install -m 0644 -D tmpfiles-qubes.conf $(DESTDIR)/usr/lib/tmpfiles.d/qubes.conf
 	install -d $(DESTDIR)/etc/logrotate.d
```
rustybird, line 9:
```suggestion
```

Rot127: Why shouldn't `logrotate.d` be installed? It does contain other configs. Sorry if I miss a point here.

rustybird: I thought you had forgotten to push this patch hunk, because the commit message already said "Remove [...] empty parent directory".

> Why shouldn't `logrotate.d` be installed? It does contain other configs.

Those all come from other packages though. They'll take care of creating the directory themselves.

> Sorry if I miss a point here.

Admittedly it's kind of pointless OCD on my part :)

Rot127: > Those all come from other packages though. They'll take care of creating the directory themselves.

Ah I see, thanks for the explanation.

Rot127:
Aren't you allow squashing commits? Sorry for making the git history so cluttered. I create a new PR after this is green.

rustybird (approved):

marmarek (approved):

PR #449, opened 2022-01-25
reflink performance, correctness, and style improvements

rustybird:
Can you trigger an openQA run please?

I've been using this for a few days but [backported to R4.0](https://github.com/rustybird/qubes-core-admin/commit/reflink4).

### DIFF
--- a/qubes/storage/reflink.py
+++ b/qubes/storage/reflink.py
@@ -438,7 +438,7 @@ def _copy_file(src, dst):
                 raise qubes.storage.StoragePoolException(str(result))
     return reflinked
 
-def is_supported(dst_dir, src_dir=None):
+def is_supported(dst_dir, *, src_dir=None):
     ''' Return whether destination directory supports reflink copies
         from source directory. (A temporary file is created in each
         directory, using O_TMPFILE if possible.)
--- a/qubes/storage/reflink.py
+++ b/qubes/storage/reflink.py
@@ -271,13 +271,14 @@ class ReflinkVolume(qubes.storage.Volume):
         if not self.rw:
             raise qubes.storage.StoragePoolException(
                 'Cannot resize: {} is read-only'.format(self.vid))
-        for path in (self._path_dirty, self._path_clean):
-            with suppress(FileNotFoundError):
-                _resize_file(path, size)
-                break
-        self._size = size
-        if path == self._path_dirty:
+        try:
+            _resize_file(self._path_dirty, size)
+        except FileNotFoundError:
+            if self.save_on_stop and not self.snap_on_start:
+                _copy_file(self._path_clean, self._path_clean, dst_size=size)
+        else:
             _update_loopdev_sizes(self._path_dirty)
+        self._size = size
         return self
 
     async def export(self):
--- a/qubes/storage/reflink.py
+++ b/qubes/storage/reflink.py
@@ -214,10 +214,7 @@ class ReflinkVolume(qubes.storage.Volume):
             if self.snap_on_start or self.save_on_stop:
                 _copy_file(self._path_clean, self._path_dirty)
             else:
-                # Preferably use the size of a leftover image, in case
-                # the volume was previously resized - but then a crash
-                # prevented qubes.xml serialization of the new size.
-                _create_sparse_file(self._path_dirty, self.size)
+                _create_sparse_file(self._path_dirty, self._size)
         return self
 
     @qubes.storage.Volume.locked
--- a/qubes/storage/reflink.py
+++ b/qubes/storage/reflink.py
@@ -348,9 +348,6 @@ class ReflinkVolume(qubes.storage.Volume):
 
     @property
     def usage(self):
-        ''' Return volume disk usage from the VM's perspective. It is
-            usually much lower from the host's perspective due to CoW.
-        '''
         for path in (self._path_dirty, self._path_clean):
             with suppress(FileNotFoundError):
                 return os.stat(path).st_blocks * 512
--- a/qubes/storage/reflink.py
+++ b/qubes/storage/reflink.py
@@ -425,20 +425,32 @@ def _attempt_ficlone(src_io, dst_io):
             raise
     return ficloned
 
-def _copy_file(src, dst):
-    ''' Copy src to dst as a reflink if possible, sparse if not. '''
-    with _replace_file(dst) as tmp_io:
-        with open(src, 'rb') as src_io:
-            reflinked = _attempt_ficlone(src_io, tmp_io)
-        if reflinked:
-            LOGGER.info('Reflinked file: %r -> %r', src, tmp_io.name)
+def _copy_file(src, dst, *, dst_size=None, copy_mtime=False):
+    ''' Transfer the data at src (and optionally its modification
+        time) to a new inode at dst, using a reflink if possible or a
+        sparsifying copy if not. Optionally, the new dst will have
+        been resized to dst_size bytes.
+    '''
+    with open(src, 'rb') as src_io, _replace_file(dst) as tmp_io:
+        if dst_size == 0:
+            reflinked = None
         else:
-            LOGGER.info('Copying file: %r -> %r', src, tmp_io.name)
-            result = subprocess.run(
-                ['cp', '--sparse=always', '--', src, tmp_io.name],
-                stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False)
-            if result.returncode != 0:
-                raise qubes.storage.StoragePoolException(str(result))
+            reflinked = _attempt_ficlone(src_io, tmp_io)
+            if reflinked:
+                LOGGER.info('Reflinked file: %r -> %r', src, tmp_io.name)
+            else:
+                LOGGER.info('Copying file: %r -> %r', src, tmp_io.name)
+                result = subprocess.run(
+                    ['cp', '--sparse=always', '--', src, tmp_io.name],
+                    stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False)
+                if result.returncode != 0:
+                    raise qubes.storage.StoragePoolException(str(result))
+            if dst_size is not None:
+                tmp_io.truncate(dst_size)
+        if copy_mtime:
+            mtime_ns = os.stat(src_io.fileno()).st_mtime_ns
+            atime_ns = mtime_ns  # Python doesn't support UTIME_OMIT
+            os.utime(tmp_io.fileno(), ns=(atime_ns, mtime_ns))
     return reflinked
 
 def is_supported(dst_dir, *, src_dir=None):
--- a/qubes/storage/reflink.py
+++ b/qubes/storage/reflink.py
@@ -369,15 +369,17 @@ _remove_file = functools.partial(
     qubes.utils.remove_file, log_level=logging.INFO)
 
 def _make_dir(path):
-    ''' mkdir path, ignoring FileExistsError; return whether we
-        created it.
-    '''
-    with suppress(FileExistsError):
+    try:
+        created = False
         os.mkdir(path)
+        created = True
+    except FileExistsError:
+        if not os.path.isdir(path):
+            raise
+    if created:
         qubes.utils.fsync_path(os.path.dirname(path))
         LOGGER.info('Created directory: %r', path)
-        return True
-    return False
+    return created
 
 def _remove_empty_dir(path):
     try:
--- a/qubes/storage/reflink.py
+++ b/qubes/storage/reflink.py
@@ -195,9 +195,10 @@ class ReflinkVolume(qubes.storage.Volume):
     def is_outdated(self):
         if self.snap_on_start:
             with suppress(FileNotFoundError):
-                # pylint: disable=protected-access
-                return (os.path.getmtime(self.source._path_clean) >
-                        os.path.getmtime(self._path_clean))
+                return not _eq_files(
+                    # pylint: disable=protected-access
+                    os.stat(self.source._path_clean),
+                    os.stat(self._path_clean))
         return False
 
     def is_dirty(self):
@@ -411,6 +412,12 @@ def _create_sparse_file(path, size):
     with _replace_file(path) as tmp_io:
         tmp_io.truncate(size)
         LOGGER.info('Created sparse file: %r', tmp_io.name)
+
+def _eq_files(stat1, stat2, *, by_attrs=('st_ino', 'st_dev')):
+    for attr in by_attrs:
+        if getattr(stat1, attr) != getattr(stat2, attr):
+            return False
+    return True
 
 def _update_loopdev_sizes(img):
     ''' Resolve img; update the size of loop devices backed by it. '''
--- a/qubes/storage/reflink.py
+++ b/qubes/storage/reflink.py
@@ -256,11 +256,7 @@ class ReflinkVolume(qubes.storage.Volume):
         if self.is_dirty():
             raise qubes.storage.StoragePoolException(
                 'Cannot revert: {} is not cleanly stopped'.format(self.vid))
-        if revision is None:
-            revision, timestamp = list(self.revisions.items())[-1]
-        else:
-            timestamp = None
-        path_revision = self._path_revision(revision, timestamp)
+        path_revision = self._path_revision(revision)
         self._add_revision()
         _rename_file(path_revision, self._path_clean)
         return self
@@ -321,9 +317,12 @@ class ReflinkVolume(qubes.storage.Volume):
                 await self._import_data_end_unlocked(success)
         return self
 
-    def _path_revision(self, revision, timestamp=None):
+    def _path_revision(self, revision=None, timestamp=None):
         if timestamp is None:
-            timestamp = self.revisions[revision]
+            if revision is None:
+                revision, timestamp = list(self.revisions.items())[-1]
+            else:
+                timestamp = self.revisions[revision]
         return self._path_clean + '.' + revision + '@' + timestamp + 'Z'
 
     @property
--- a/qubes/storage/reflink.py
+++ b/qubes/storage/reflink.py
@@ -238,10 +238,11 @@ class ReflinkVolume(qubes.storage.Volume):
     def _add_revision(self):
         if self.revisions_to_keep == 0:
             return
-        ctime = os.path.getctime(self._path_clean)
-        timestamp = qubes.storage.isodate(int(ctime))
-        _copy_file(self._path_clean,
-                   self._path_revision(self._next_revision, timestamp))
+        timestamp = qubes.storage.isodate(
+            int(os.path.getctime(self._path_clean)))
+        _copy_file(
+            self._path_clean,
+            self._path_revision(self._next_revision, timestamp))
 
     def _prune_revisions(self, keep=None):
         if keep is None:
@@ -382,12 +383,16 @@ def _create_dir(path):
 
 def _remove_empty_dir(path):
     try:
+        removed = False
         os.rmdir(path)
-        qubes.utils.fsync_path(os.path.dirname(path))
-        LOGGER.info('Removed empty directory: %r', path)
+        removed = True
     except OSError as ex:
         if ex.errno not in (errno.ENOENT, errno.ENOTEMPTY):
             raise
+    if removed:
+        qubes.utils.fsync_path(os.path.dirname(path))
+        LOGGER.info('Removed empty directory: %r', path)
+    return removed
 
 def _resize_file(path, size):
     ''' Resize an existing file. '''
@@ -414,13 +419,14 @@ def _update_loopdev_sizes(img):
 
 def _attempt_ficlone(src_io, dst_io):
     try:
+        ficloned = False
         fcntl.ioctl(dst_io.fileno(), FICLONE, src_io.fileno())
-        return True
+        ficloned = True
     except OSError as ex:
         if ex.errno not in (errno.EBADF, errno.EINVAL,
                             errno.EOPNOTSUPP, errno.EXDEV):
             raise
-        return False
+    return ficloned
 
 def _copy_file(src, dst):
     ''' Copy src to dst as a reflink if possible, sparse if not. '''
@@ -445,7 +451,7 @@ def is_supported(dst_dir, *, src_dir=None):
     '''
     if src_dir is None:
         src_dir = dst_dir
-    with tempfile.TemporaryFile(dir=src_dir) as src, \
-         tempfile.TemporaryFile(dir=dst_dir) as dst:
-        src.write(b'foo')  # don't let any fs get clever with empty files
-        return _attempt_ficlone(src, dst)
+    with tempfile.TemporaryFile(dir=src_dir) as src_io, \
+         tempfile.TemporaryFile(dir=dst_dir) as dst_io:
+        src_io.write(b'foo')  # don't let any fs get clever with empty files
+        return _attempt_ficlone(src_io, dst_io)
--- a/qubes/storage/reflink.py
+++ b/qubes/storage/reflink.py
@@ -333,7 +333,7 @@ class ReflinkVolume(qubes.storage.Volume):
     @property
     def revisions(self):
         prefix = self._path_clean + '.'
-        paths = glob.iglob(glob.escape(prefix) + '*@*Z')
+        paths = glob.iglob(glob.escape(prefix) + '*?@????-??-??T??:??:??Z')
         items = (path[len(prefix):-1].split('@') for path in paths)
         return collections.OrderedDict(
             sorted(items, key=lambda item: int(item[0])))
--- a/qubes/storage/reflink.py
+++ b/qubes/storage/reflink.py
@@ -223,10 +223,9 @@ class ReflinkVolume(qubes.storage.Volume):
     @qubes.storage.Volume.locked
     @_coroutinized
     def stop(self):  # pylint: disable=invalid-overridden-method
-        if self.save_on_stop:
-            if os.path.exists(self._path_dirty):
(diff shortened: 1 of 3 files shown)


rustybird:
`_copy_file()` TODO: 

- don't copy the modification time if called with a `dst_size` argument
- optimize `dst_size==0` case

DemiMarie:
@rustybird would you mind adding tests for the tricky migration cases?

rustybird:
@DemiMarie you mean [`_remove_stale_precache()`](https://github.com/rustybird/qubes-core-admin/blob/5045da8eea73a324bfd017d221e51ac0b1b6cc32/qubes/storage/reflink.py#L161-L171)? Not sure how to write a realistic test for that Qubes upgrade->downgrade->upgrade scenario

DemiMarie:
> @DemiMarie you mean [`_remove_stale_precache()`](https://github.com/rustybird/qubes-core-admin/blob/5045da8eea73a324bfd017d221e51ac0b1b6cc32/qubes/storage/reflink.py#L161-L171)?

Yes, and for all of the other code you added not covered by tests.

> Not sure how to write a realistic test for that Qubes upgrade->downgrade->upgrade scenario

Thanks for asking!  I have had the same difficulty myself in the past.  In this case, you can manually create the state that is expected after an upgrade and subsequent downgrade, and test that the upgraded code handles this correctly.

rustybird:
Yea it definitely needs some more unit tests! At least for the meaty parts like the souped up  `_copy_file()`, and also `_eq_files()`.

> > Not sure how to write a realistic test for that Qubes upgrade->downgrade->upgrade scenario
 
> In this case, you can manually create the state that is expected after an upgrade and subsequent downgrade, and test that the upgraded code handles this correctly.

But the `_remove_stale_precache()` *code* is trivial, once `_eq_files()` is unit tested. Everything tricky is in the *assumptions* that it makes about how Qubes version changes could affect the pool state. If I "duplicated" those in unit tests it would be boxticking rather than substantially testing anything.

I was thinking about setting up integration tests that actually upgrade/downgrade qubes module code. Oof...

rustybird:
> I was thinking about setting up integration tests that actually upgrade/downgrade qubes module code. Oof...

Oof indeed. That sort of integration testing looks like it would be a huge project, so I went with unit testing after all. `_update_precache()` is nulled out to simulate a Qubes downgrade, which should be realistic enough.

3hhh:
Thanks for working on this, much appreciated!

rustybird:
@3hhh you're welcome, and thank you for suggesting the precache idea.

qubes/storage/reflink.py
```diff
@@ -352,17 +383,14 @@ def size(self):
 
     @property
     def usage(self):
-        ''' Return volume disk usage from the VM's perspective. It is
-            usually much lower from the host's perspective due to CoW.
-        '''
```
DemiMarie, line 357: Any reason to remove this comment?

qubes/storage/reflink.py
```diff
@@ -372,25 +400,38 @@ def _replace_file(dst):
 _remove_file = functools.partial(
     qubes.utils.remove_file, log_level=logging.INFO)
 
-def _make_dir(path):
-    ''' mkdir path, ignoring FileExistsError; return whether we
-        created it.
-    '''
```
DemiMarie, line 378: Ditto

qubes/storage/reflink.py
```diff
@@ -372,25 +400,38 @@ def _replace_file(dst):
...
+        created = True
+    except FileExistsError:
+        if not os.path.isdir(path):
+            raise
+    if created:
         qubes.utils.fsync_path(os.path.dirname(path))
         LOGGER.info('Created directory: %r', path)
-        return True
-    return False
+    return created
```
DemiMarie, line 421: Would it be better to use `mkdirat()` instead?  Not sure if Python supports it.  In any case this can be done in a separate PR.

qubes/storage/reflink.py
```diff
@@ -147,16 +147,38 @@ def __init__(self, *args, **kwargs):
...
+    def _remove_stale_precache(self):
+        ''' Defuse the following situation: We created a precache.
+            After that, the pool was used on an older Qubes version
+            without precache support - causing staleness. Now the
+            pool is here again on new Qubes.
+        '''
+        stat_clean = os.stat(self._path_clean)
+        with suppress(FileNotFoundError):
+            if not _eq_files(stat_clean, os.stat(self._path_precache),
+                             by_attrs=['st_mtime_ns', 'st_size']):
```
DemiMarie, line 170: This is very fragile.  Timekeeping in Qubes OS is handled by `sys-net`, which is not trusted.  I would prefer for precache support to be a separate PR.

DemiMarie (changes requested):
Would you mind moving the precache work to a separate PR, for reviewability?

rustybird:
> Would you mind moving the precache work to a separate PR, for reviewability? It’s okay if it requires this PR.

GitHub is dreadful IME when one PR's HEAD is an ancestor of another...

I've reordered the commits (and squashed some unit test commits into their related feature implementation commits), so that the series now ends with a single self-contained precache commit. Does this work for you?

qubes/storage/reflink.py
rustybird: See [commit](https://github.com/QubesOS/qubes-core-admin/pull/449/commits/f482916999ee4d2868281a9a5d07e340a1d9185d) message

qubes/storage/reflink.py
rustybird: That one seemed pointless after polishing the function

qubes/storage/reflink.py
rustybird: Python would [support it](https://docs.python.org/3/library/os.html#dir-fd), however none of the relink functions attempt to cater to a user doing wild stuff like replacing a directory inside a pool *in the middle of a function call* that's operating on that directory.

> In any case this can be done in a separate PR.

Hmm I generally try to avoid submitting a very small refactor (that modifies the same file) as a separate PR, if the marginal cost of reviewing one more commit in a series would be dwarfed by the overhead (silly merge conflicts, context switching, CI) of another PR.

qubes/storage/reflink.py
rustybird: `_remove_stale_precache()` doesn't attempt to sort the two modification times into newer/older, it triggers if they are not precisely the same. How could a malicious ClockVM on the old Qubes installation (before precache support) attack this?

To be clear this method doesn't have a load-bearing role in normal installations. Its only job is to address the corner case that some people might upgrade->downgrade->upgrade the Qubes version on which they use their (e.g. secondary) pool. Once that is no longer plausible, the code can be deleted.

rustybird: > What happens if they are precisely the same? If the result is merely a missed optimization that is fine, but if it is data corruption, not so much.

Obsolete volume data would be used in this case, so data corruption in a sense. Which I'm not saying is acceptable. It just seems like a negligible risk that the malicious ClockVM (on old Qubes OS) could arrange this, with nanosecond precision - even if it were somehow able to guess within one second when the attacked volume was last shut down and when it will be shut down next.

> Could something more robust (perhaps based on xattrs) be used instead? One could add a cryptographically random xattr to each file, copy the xattrs when making a reflink, and check that the xattrs match.

It's a cool idea, and Python has xattr support. OTOH it's more code on top of the size check code (to detect old Qubes OS resizing the shut down volume, which [would not change](https://github.com/QubesOS/qubes-core-admin/pull/449/commits/0f72c3e7f788b3e9a2e2a3fc6be8c31d2d42a1a3) its xattr) and also still the modification time copy/check code (to detect old Qubes OS shrinking the volume, then expanding it to the previous size again).

:shrug:

qubes/storage/reflink.py
DemiMarie: > Python would [support it](https://docs.python.org/3/library/os.html#dir-fd), however none of the relink functions attempt to cater to a user doing wild stuff like replacing a directory inside a pool _in the middle of a function call_ that's operating on that directory.

I would still prefer to use `mkdirat()`, but that might be me being overly pedantic.  Theoretically `mkdirat()` should be a teeny bit faster but it almost certainly does not matter.  That said, `mkdirat()` might even lead to cleaner code.

> > In any case this can be done in a separate PR.
> 
> Hmm I generally try to avoid submitting a very small change (that modifies the same file) as a separate PR, if the marginal cost of reviewing one more commit in a series would be dwarfed by the overhead (silly merge conflicts, context switching, CI) of another PR.

:+1:

qubes/storage/reflink.py
DemiMarie: > > What happens if they are precisely the same? If the result is merely a missed optimization that is fine, but if it is data corruption, not so much.
> 
> Obsolete volume data would be used in this case, so data corruption in a sense. Which I'm not saying is acceptable. It just seems like a negligible risk that the malicious ClockVM (on old Qubes OS) could arrange this, with nanosecond precision - even if it were somehow able to guess within one second when the attacked volume was last shut down and when it will be shut down next.

This still worries me, though it is ultimately up to @HW42 and @marmarek as the members of the Qubes OS Security Team.  Is this something that can be exploited without a downgrade?

> > Could something more robust (perhaps based on xattrs) be used instead? One could add a cryptographically random xattr to each file, copy the xattrs when making a reflink, and check that the xattrs match.
> 
> It's a cool idea, and Python has xattr support. OTOH it's more code on top of the size check code (to detect old Qubes OS resizing the shut down volume, which [would not change](https://github.com/QubesOS/qubes-core-admin/pull/449/commits/0f72c3e7f788b3e9a2e2a3fc6be8c31d2d42a1a3) its xattr) and also still the modification time copy/check code (to detect old Qubes OS shrinking the volume, then expanding it to the previous size again).

What would happen in that case?

qubes/storage/reflink.py
```diff
@@ -321,9 +317,12 @@ async def import_volume(self, src_volume):
...
         return self
 
-    def _path_revision(self, revision, timestamp=None):
+    def _path_revision(self, revision=None, timestamp=None):
         if timestamp is None:
-            timestamp = self.revisions[revision]
+            if revision is None:
+                revision, timestamp = list(self.revisions.items())[-1]
+            else:
+                timestamp = self.revisions[revision]
```
DemiMarie, line 361: This is not equivalent to the old code; consider the case where timestamp is not `None` but revision is.

DemiMarie (changes requested):
cc268bdc1c3624cc1a089f6dc645f87ec3b491ff needs to be fixed; either the code needs to be changed or the commit message does.  With that exception, I would be fine with including 069970e6f0273c67c488c1e4525b2311a2696650...16ac1bde37dc4a463e6352e3f3c8a308a2dc3720 in an R4.1 update if @marmarek is, but 3f4dfd64880f10fccba9552e4129c99aee247e0c should wait for R4.2.

DemiMarie:
> I've reordered the commits (and squashed some unit test commits into their related feature implementation commits), so that the series now ends with a single self-contained precache commit. Does this work for you?

That is definitely an improvement (thanks!), but the precache change appears to be by far the riskiest of the series.  I would rather it wait for R4.2.  The export bugfix should definitely be included in an R4.1 update, though.

qubes/storage/reflink.py
rustybird: > Is this something that can be exploited without a downgrade?

No without a downgrade the mtime is just an opaque integer that's copied along with the file, and the ClockVM is out of the picture.

>> OTOH it's more code on top of the size check code (to detect old Qubes OS resizing the shut down volume, which [would not change](https://github.com/QubesOS/qubes-core-admin/pull/449/commits/0f72c3e7f788b3e9a2e2a3fc6be8c31d2d42a1a3) its xattr) and also still the modification time copy/check code (to detect old Qubes OS shrinking the volume, then expanding it to the previous size again).

> What would happen in that case?

Shrinking and expanding again would be like punching a hole at the end of the file, so that part of the data would be replaced by zeros (in contrast to whatever it is in the now stale precache).

HW42: > My understanding (correct me if I am wrong @marmarek @HW42) is that a compromise of the dom0 CSPRNG is considered out of scope for the Qubes OS threat model.

"Out of scope" is a strange way to put it, but I think I know what you mean (and that's correct in my opinion): If someone is able to predict dom0's RNG output it's already game over. That being said in general I prefer if relying on a secret can be avoided , not the least given all the side channel issues (but this is irrelevant for this proposal).

> This still worries me, though it is ultimately up to @HW42 and @marmarek as the members of the Qubes OS Security Team.

I just had a quick look after your ping on this PR so I might be missing something, but given what I have read I tend towards @rustybird's argument, because:

 - Probably tricky to exploit at all (due to high res)
 - Only reachable at all in a specific scenario (down- and then upgrade, IIUC)
 - Current solution keeps the code simpler
 - Even if exploited it would be a one time rollback. While this isn't great it's something that you have in general think about anyway (unclean shutdown, backups, etc.)

Independently improving the clock sync situation is something on my "if I have time" wishlist but this is getting off-topic.

HW42:
> GitHub is dreadful IME when one PR's HEAD is an ancestor of another...

+1, it's really annoying that GitHub can't handle dependencies between PRs better. (If this specific PR should be split, I will leave to others)

qubes/storage/reflink.py
DemiMarie: > > My understanding (correct me if I am wrong @marmarek @HW42) is that a compromise of the dom0 CSPRNG is considered out of scope for the Qubes OS threat model.
> 
> "Out of scope" is a strange way to put it, but I think I know what you mean (and that's correct in my opinion): If someone is able to predict dom0's RNG output it's already game over. That being said in general I prefer if relying on a secret can be avoided , not the least given all the side channel issues (but this is irrelevant for this proposal).

You understood me correctly.  I 100% agree on not relying on a secret if it can be helped, but in this case the only requirement is uniqueness.  Is that what you meant by “but this is irrelevant for this proposal”?

> > This still worries me, though it is ultimately up to @HW42 and @marmarek as the members of the Qubes OS Security Team.
> 
> I just had a quick look after your ping on this PR so I might be missing something, but given what I have read I tend towards @rustybird's argument, because:
> 
> * Probably tricky to exploit at all (due to high res)

I agree with this assessment.

> * Only reachable at all in a specific scenario (down- and then upgrade, IIUC)

That does make me feel much better, but it is the kind of subtle fact that should definitely be documented in the code itself, along with a proof (and tests) that it is not normally reachable.

> * Current solution keeps the code simpler

That it does.

> * Even if exploited it would be a one time rollback. While this isn't great it's something that you have in general think about anyway (unclean shutdown, backups, etc.)

Agreed.  I could see it being used to revert a critical security patch but that is starting to stretch the limits of practicality.  In any case if this winds up being delayed until R4.2 (as I consider likely), this won’t be an issue (downgrades from R4.2 to R4.1 will most likely be unsupported).

> Independently improving the clock sync situation is something on my "if I have time" wishlist but this is getting off-topic.

That it is.

qubes/storage/reflink.py
rustybird: It's intentional, I've added a note to the commit message

qubes/storage/reflink.py
rustybird: > `mkdirat()` might even lead to cleaner code.

Nah it gets really convoluted in Python because you have to switch to lower level code patterns, e.g. look at the side-by-side view of https://github.com/rustybird/qubes-core-admin/commit/dir_fd_example

rustybird:
Jesus, R4.2? So far I had thought of it as suitable for R4.0 even.

Anyway not my call to make ofc, and I'm fine with using it in my own dom0 until whenever.

qubes/storage/reflink.py
DemiMarie: That makes sense.  I suggest including this in the function’s docstring, and throwing a better exception than `TypeError: can only concatenate str (not "NoneType") to str`.

qubes/storage/reflink.py
```diff
@@ -438,7 +438,7 @@ def _copy_file(src, dst):
                 raise qubes.storage.StoragePoolException(str(result))
     return reflinked
 
-def is_supported(dst_dir, src_dir=None):
+def is_supported(dst_dir, *, src_dir=None):
```
DemiMarie, line 504: Should both arguments be keyword-only?  That would be even easier to read at the expense of some verbosity and a bigger diff.  Obviously this is optional.

DemiMarie (commented):
Same comments as last time, but I agree with @HW42’s assessment security-wise.

qubes/storage/reflink.py
rustybird: I'm sad to report that I had bikeshedded these alternatives (and a third: deleting `src_dir`) in my own head for a silly amount of time. This one seemed fine I guess lol

DemiMarie:
> Jesus, R4.2? So far I had thought of it as suitable for R4.0 even.

That was my opinion for the precache change (only).  The bug fix should definitely be backported to R4.0 as it could (if I understand correctly) cause data corruption.

> Anyway not my call to make ofc, and I'm fine with using it in my own dom0 until whenever.

Not my call to make either.

rustybird:
> The bug fix should definitely be backported to R4.0 as it could (if I understand correctly) cause data corruption.

I'm not sure what the impact is. Looks like all storage drivers are currently getting this wrong...

Theoretically it should cause a nuisance like https://github.com/QubesOS/qubes-issues/issues/7198 if a shut down volume is resized during a backup, but I haven't been able to reproduce it

qubes/storage/reflink.py
HW42: @DemiMarie:
> Is that what you meant by “but this is irrelevant for this proposal”?

Exactly, the value would not have to be secret, so leaks are not relevant in this specific case.

DemiMarie:
> I'm not sure what the impact is. Looks like all storage drivers are currently getting this wrong...

I think the file driver actually gets this right.  Not sure about the LVM2 driver.

rustybird:
> I think the file driver actually gets this right.

How so? Its `resize()` truncates `self.path` in place, and the latter is returned by `export()`. Like file-reflink before the "don't mutate _path_clean file contents" commit

DemiMarie:
> > I think the file driver actually gets this right.
> 
> How so? Its `resize()` truncates `self.path` in place, and the latter is returned by `export()`. Like file-reflink before the "don't mutate _path_clean file contents" commit

Good catch!  Since the file driver is deprecated I suggest throwing an exception if one tries to resize an exported volume.  The LVM driver will need to take a snapshot, which brings the question of how to handle cleanup.  Whereas with reflink the file will automatically be cleaned up by the kernel, with dm-thin there is no automatic cleanup, so qubesd needs to handle this manually.  One approach would be to use the `export_end` hook to do the cleanup.  That hook may not get called, though (consider a system reboot), so qubesd would also need to remove any bogus export volumes during system startup.

rustybird:
> Since the file driver is deprecated I suggest throwing an exception if one tries to resize an exported volume.

:+1:

> The LVM driver will need to take a snapshot, which brings the question of how to handle cleanup. Whereas with reflink the file will automatically be cleaned up by the kernel, with dm-thin there is no automatic cleanup

I wonder if there's some undocumented way to make LVM2 play nice with a deferred removal à la `dmsetup remove --deferred`. Then it would be straightforward: The `export()` API could be changed to return an open file handle instead of a path, `export_end()` would close it, and the kernel would hopefully garbage collect the volume.

marmarek:
Thanks @rustybird!

Can you add few more tests for the precache feature? There is [a test](https://github.com/QubesOS/qubes-core-admin/blob/master/qubes/tests/integ/storage.py#L180) for the case of TemplateVM shutdown and then AppVM, but not the other way (and I could see a failure mode where shutting down TemplateVM after AppVM would make subsequent AppVM start use the old volume).
Ideally, those would be unit tests, but since we have integration tests for this, I'm fine with new test being that type too.

rustybird:
@marmarek: I can add such an integration test, but because `snap_on_start` volumes don't have a precache (and they don't use their source's precache either) it would actually exercise the [hardlink feature](https://github.com/QubesOS/qubes-core-admin/pull/449/commits/6515cae2e1f2243d177b57b6c74d399fed4288f9).

rustybird:
> worth adding a test anyway

Done.

FWIW installing R4.1 with the latest `reflink.py` (updated in `/mnt/sysroot` during the installation) went fine.

qubes/storage/reflink.py
DemiMarie: Good point, although utility functions would make that easier.

DemiMarie:
> > Since the file driver is deprecated I suggest throwing an exception if one tries to resize an exported volume.
> 
> +1
> 
> > The LVM driver will need to take a snapshot, which brings the question of how to handle cleanup. Whereas with reflink the file will automatically be cleaned up by the kernel, with dm-thin there is no automatic cleanup
> 
> I wonder if there's some undocumented way to make LVM2 play nice with a deferred removal à la `dmsetup remove --deferred`. Then it would be straightforward: The `export()` API could be changed to return an open file handle instead of a path, `export_end()` would close it, and the kernel would hopefully garbage collect the LVM2 volume.

Another option would be to have an `export` snapshot, which qubesd automatically garbage-collects when it starts up for the first time on a system.  That said, there is still a problem: it is not safe (as in: one can open the wrong device) to destroy a device-mapper device (including a logical volume) before one is certain that nobody else will be trying to open it.  A handle-based API would be safe, as calls to `open()` would be serialized (by qubesd) against concurrent calls to `lvremove`.

@marmarek: is it okay to use change `export()` to use a handle-based API instead of a file-based one?  At least the backup code should hopefully not be too hard to modify.

marmarek:
> That said, there is still a problem: it is not safe (as in: one can open the wrong device) to destroy a device-mapper device (including a logical volume) before one is certain that nobody else will be trying to open it.

No, there is explicit serialization here, using `export_end()`. You must cease to use the path you get from `export()`  when you call `export_end()`.
So, a snapshot for a backup may be a good idea, and no API change is necessary.

DemiMarie:
> > That said, there is still a problem: it is not safe (as in: one can open the wrong device) to destroy a device-mapper device (including a logical volume) before one is certain that nobody else will be trying to open it.
> 
> No, there is explicit serialization here, using `export_end()`. You must cease to use the path you get from `export()` when you call `export_end()`. So, a snapshot for a backup may be a good idea, and no API change is necessary.

What happens if qubesd gets restarted in the mean time?

marmarek:
> What happens if qubesd gets restarted in the mean time?

Then the backup gets aborted (which is rather unpleasant experience - qubesos/qubes-issues#5004). But if any other user would be there, it would still be obligated to call `export_end()` when it's done with the file.
Theoretically, since currently there are no users of this API outside of qubesd, we may consider calling `export_end()` at qubesd startup, but in practice, I don't want to exclude possibility of using this outside of qubesd (for example, to split making backup into separate process).

And BTW, I don't like the file handle idea, I'd prefer an explicit cleanup. Not everything can be cleaned up on FD close (LVM being one example, but even if that can be worked around, I don't want to introduce this API limitation for any future storage pool).

rustybird:
> And BTW, I don't like the file handle idea, I'd prefer an explicit cleanup. Not everything can be cleaned up on FD close (LVM being one example, but even if that can be worked around, I don't want to introduce this API limitation for any future storage pool).

For sure. I was proposing to pass the file handle back to `export_end()` (replacing the current `path` argument), and to have its implementation in the `Volume` base class *default to* closing it. Storage drivers should be free to override this.

DemiMarie:
> > What happens if qubesd gets restarted in the mean time?
> 
> Then the backup gets aborted (which is rather unpleasant experience - [QubesOS/qubes-issues#5004](https://github.com/QubesOS/qubes-issues/issues/5004)). But if any other user would be there, it would still be obligated to call `export_end()` when it's done with the file. Theoretically, since currently there are no users of this API outside of qubesd, we may consider calling `export_end()` at qubesd startup, but in practice, I don't want to exclude possibility of using this outside of qubesd (for example, to split making backup into separate process).

Is there a way to prevent a long-term resource leak without any race conditions?  In theory at least, it should be safe to call `export_end()` at _system_ startup, and have qubesd indicate (via a flag file in `/run` or similar) that all of these calls are finished.  Calling `export()` before then would fail.

> And BTW, I don't like the file handle idea, I'd prefer an explicit cleanup. Not everything can be cleaned up on FD close (LVM being one example, but even if that can be worked around, I don't want to introduce this API limitation for any future storage pool).

Explicit cleanup is fine.  The advantage of a file handle is that it avoids race conditions when calling `open()` on device-mapper devices.  And from what I understand, it is the _only_ way to do so reliably.

marmarek:
PipelineRetry

qubes/tests/integ/storage.py
```diff
@@ -257,6 +257,30 @@ async def _test_003_snapshot(self):
                     size),
                 user='root')
 
+        await self.vm2.shutdown(wait=True)
```
marmarek, line 260: This shutdown consistently times out. VM log ends with `A stop job is running for Session c3 of User user` (which I think is the one that causes the timeout). Maybe some weird interaction with shutting down when not everything started yet? Try adding `wait_for_session(self.vm2)` before.

rustybird: > This shutdown consistently times out. VM log ends with `A stop job is running for Session c3 of User user` (which I think is the one that causes the timeout).

After resetting my `default_shutdown_timeout` (I had it cranked up to 300 seconds) I saw the same kind of error but for vm1.

> Maybe some weird interaction with shutting down when not everything started yet? Try adding `wait_for_session(self.vm2)` before.

Waiting for sessions did the trick. There even was a convenient `start_vm()` helper method.

PR #450, opened 2022-02-03
Use long options for LVM2

DemiMarie:
Long options are usually self-explanatory and are much more readable
than short options, which are quite cryptic.  Also place all non-option
arguments after `--` to mark end of options.


marmarek:
pylint and unit tests found some issue (I think the same one).

marmarek (approved):

PR #451, opened 2022-02-06
Send stderr of qubes.ShowInTerminal to /dev/null

DemiMarie:
It isn't trusted.


marmarek:
Please don't. `qvm-run` does sanitize output printed to the console (unless you explicitly request otherwise). On the other hand, hiding error messages makes it very frustrating to debug things (and also results in a _much_ lower quality bug reports/support requests).

PR #452, opened 2022-02-09
Minor test fixes

marmarek:


DemiMarie (approved):
Looks good, though I do wonder why this test is in `qubes-core-admin` 🙂

marmarek:
> I do wonder why this test is in `qubes-core-admin` slightly_smiling_face

Because this is where we (currently) keep generic integration tests.

PR #454, opened 2022-03-01
Fix interrupting VM startup using qvm-kill

marmarek:
Fixes QubesOS/qubes-issues#6322


DemiMarie:
Thanks for the fix!

jevank:
Will this fix the https://github.com/QubesOS/qubes-issues/issues/7116 too?

marmarek:
yes

PR #457, opened 2022-03-26
tests: increase libvirt cleanup timeout

marmarek:


DemiMarie (approved):
Looks good to me!  I remember this being a problem in the past.

PR #458, opened 2022-03-31
Remove stale pylint suppressions

DemiMarie:
Those were from before `yield from` was replaced by `await`.


marmarek:
PipelineRetry

PR #459, opened 2022-03-31
Have qubesd open the backup file

DemiMarie:
This is preparation for taking a snapshot during backup.

### DIFF
--- a/qubes/backup.py
+++ b/qubes/backup.py
@@ -630,13 +630,16 @@ class Backup:
                 path = file_info.path
                 if callable(path):
                     path = await qubes.utils.coro_maybe(path())
+                subdir = file_info.subdir
+                assert re.fullmatch('(?:dom0-home|vm[1-9][0-9]*)/', subdir), \
+                        f'bad subdir {subdir!r}'
                 tar_cmdline = (["tar", "-Pc", '--sparse',
                                 '-C', os.path.dirname(path)] +
                                (['--dereference'] if
-                               file_info.subdir != "dom0-home/" else []) +
+                               subdir != "dom0-home/" else []) +
                                ['--xform=s:^%s:%s\\0:' % (
                                    os.path.basename(path),
-                                   file_info.subdir),
+                                   subdir),
                                    os.path.basename(path)
                                ])
                 file_stat = os.stat(path)
@@ -649,14 +652,12 @@ class Backup:
                         "Renaming directories not supported"
                     tar_cmdline = ['python3', '-m', 'qubes.tarwriter',
                         '--override-name=%s' % (
-                            os.path.join(file_info.subdir, os.path.basename(
+                            os.path.join(subdir, os.path.basename(
                                 file_info.name))),
                         path]
                 if self.compressed:
                     tar_cmdline.insert(-2,
                         "--use-compress-program=%s" % self.compression_filter)
-
-                self.log.debug(" ".join(tar_cmdline))
 
                 # Pipe: tar-sparse | scrypt | tar | backup_target
                 # TODO: log handle stderr
--- a/qubes/backup.py
+++ b/qubes/backup.py
@@ -630,40 +630,44 @@ class Backup:
                 path = file_info.path
                 if callable(path):
                     path = await qubes.utils.coro_maybe(path())
-                subdir = file_info.subdir
-                assert re.fullmatch('(?:dom0-home|vm[1-9][0-9]*)/', subdir), \
-                        f'bad subdir {subdir!r}'
-                tar_cmdline = (["tar", "-Pc", '--sparse',
-                                '-C', os.path.dirname(path)] +
-                               (['--dereference'] if
-                               subdir != "dom0-home/" else []) +
-                               ['--xform=s:^%s:%s\\0:' % (
-                                   os.path.basename(path),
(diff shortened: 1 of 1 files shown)


qubes/backup.py
```diff
@@ -630,39 +630,44 @@ async def _wrap_and_send_files(self, files_to_backup, output_queue):
...
-
-                self.log.debug(" ".join(tar_cmdline))
-
-                # Pipe: tar-sparse | scrypt | tar | backup_target
-                # TODO: log handle stderr
-                # pylint: disable=not-an-iterable
-                tar_sparse = await asyncio.create_subprocess_exec(
-                    *tar_cmdline, stdout=subprocess.PIPE)
+
+                fd = os.open(path, os.O_RDONLY|o.O_NOCTTY|os.O_CLOEXEC)
```
marmarek, line 634: This won't work, `path` may point at a directory (for example for backing up dom0 home).

qubes/backup.py
```diff
@@ -630,39 +630,44 @@ async def _wrap_and_send_files(self, files_to_backup, output_queue):
...
+                try:
+                    fd_path = '/proc/self/fd/' + str(fd)
+                    tar_cmdline = (["tar", "-Pc", '--sparse',
+                                    '-C', os.path.dirname(path)] +
+                                   (['--dereference'] if
+                                   file_info.subdir != "dom0-home/" else []) +
+                                   ['--xform=s:^%s:%s\\0:' % (
```
marmarek, line 641: If you change how the argument looks like, you need to adjust this option too.

qubes/backup.py
```diff
@@ -630,39 +630,44 @@ async def _wrap_and_send_files(self, files_to_backup, output_queue):
...
+                    if self.compressed:
+                        tar_cmdline.insert(-2,
+                            "--use-compress-program=" + self.compression_filter)
+
+                    self.log.debug(" ".join(tar_cmdline))
+
+                    # Pipe: tar-sparse | scrypt | tar | backup_target
+                    # TODO: log handle stderr
+                    tar_sparse = await asyncio.create_subprocess_exec(
+                        *tar_cmdline, stdout=subprocess.PIPE, keep_fds=(fd,))
```
marmarek, line 668: Can you tell _exactly_ what's the point in opening FD in qubesd, when it's taken over by a process started few lines below? If qubesd is terminated before this process starts - nothing would try to open the exported path (so it can be cleaned up). If its terminated after (but before backup finishes), then the backup process will be interrupted anyway (see the stdout of this process is handled in qubesd - so terminating qubesd _will_ interrupt the backup. It doesn't matter if tar will manage to open that path or not.

marmarek (changes requested):

qubes/backup.py
DemiMarie: `os.open()` works just fine on directories (as with C’s `open(2)`).

qubes/backup.py
DemiMarie: Indeed I do.  That reminds me that the current code has a bug: `\`, `&`, and `:` in the second argument must be escaped.

qubes/backup.py
DemiMarie: Good point.  I guess this isn’t strictly necessary in that case.  That said, it does make the code easier to reason about.  With this approach, it is _obvious_ that there are no races, whereas this is not the case with the current approach.  It also avoids having to figure out proper escaping for the match part of `--xform`.

DemiMarie:
Draft because I have not tested this yet.  I will before marking it ready for review.

qubes/backup.py
marmarek: >  That reminds me that the current code has a bug: \, &, and : in the second argument must be escaped.

None of them can possibly appear in `file_info.subdir` (it's either `dom0-home/` or `vmX/`). If you _really_ need to do anything about it (which I think you don't), better add an assert, instead of possibly non-obvious escaping.

marmarek:
PipelineRetry

qubes/backup.py
DemiMarie: > > That reminds me that the current code has a bug: , &, and : in the second argument must be escaped.
> 
> None of them can possibly appear in `file_info.subdir` (it's either `dom0-home/` or `vmX/`). If you _really_ need to do anything about it (which I think you don't), better add an assert, instead of possibly non-obvious escaping.

I will add an assert, mostly because the current code is _really_ non-obvious to someone (me) who has not looked at it before.

marmarek:
PipelineRetry

marmarek:
PipelineRetry

marmarek:
PipelineRetry

marmarek:
Looks like CI has some issue (@fepitre will look in the morning). If pylint would run, it would point out typo in `o.O_NOCTTY`.

DemiMarie:
> Looks like CI has some issue (@fepitre will look in the morning). If pylint would run, it would point out typo in `o.O_NOCTTY`.

yeah, good catch, sorry about that.  I fixed the typo but am closing this PR because I am not sure if it is needed.

fepitre:
PipelineRetry

DemiMarie:
@marmarek would you mind reopening this?  I can’t for some reason.

PR #460, opened 2022-04-01
Draft: lvm2 backend: Disable sync with udev

DemiMarie:
All of the logical volumes managed by qubesd are ignored by udev, so it
is safe for lvm2 to manage the device nodes in /dev directly.  This
should improve performance.

Marking as draft because udev needs to be made to actually ignore these
volumes, to the extent of not even creating `/dev/mapper/*` or `/dev/VG/LV`
symlinks for them.

### DIFF
--- a/qubes/storage/lvm.py
+++ b/qubes/storage/lvm.py
@@ -235,7 +235,7 @@ def _parse_lvm_cache(lvm_output):
 
 def init_cache(log=logging.getLogger('qubes.storage.lvm')):
     cmd = _init_cache_cmd
-    environ={'LC_ALL': 'C.UTF-8', **os.environ}
+    environ={'LC_ALL': 'C.UTF-8', 'DM_DISABLE_UDEV': '1', **os.environ}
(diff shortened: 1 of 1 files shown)


qubes/storage/lvm.py
```diff
@@ -776,11 +776,11 @@ def _get_lvm_cmdline(cmd):
...
         raise NotImplementedError('unsupported action: ' + action)
     if os.getuid() != 0:
-        cmd = [_sudo, _lvm] + lvm_cmd
+        prefix = [_sudo, 'LC_ALL=C.UTF-8', 'DM_DISABLE_UDEV=1', '--', _lvm]
     else:
-        cmd = [_lvm] + lvm_cmd
+        prefix = [_lvm]
 
-    return cmd
+    return cmd + lvm_cmd
```
marmarek, line 783:
```suggestion
    return prefix + lvm_cmd
```

marmarek:
I like it, if it really works.

DemiMarie:
> I like it, if it really works.

I will need to add some more lvm2 patches to make it work, mostly suppressing unhelpful junk spewed to stderr.

PR #461, opened 2022-04-01
lvm2 pool: Export immutable snapshot of volume

### LINKED ISSUE #7198: Restore failure - Unexpected EOF - retrieved data backup size exceed expected size
### Qubes OS release
4.1.0-rc3

### Brief summary
Sporadic but majority failed restores (or corrupted backups) on 4.1x installations. Any size backup, small or large. Using Btrfs. Restoring on same version of Qubes, same and different systems.
Will restore AppVM but contain no data. Will display an error but conclude with "Finished successfully!", when it has not.

Only one reference to the same issue from 2017 found here https://groups.google.com/g/qubes-users/c/0i7RfWhoCxU

Backups and data do appear to restore successfully if running **qvm-backup-restore --ignore-size-limit** via dom0.

Forum issue discussion here - https://forum.qubes-os.org/t/backups-are-corrupted-on-restore/8446

### Steps to reproduce
Create a backup and restore it. 

### Expected behavior
Successful restore containing all data with no errors.

### Actual behavior
Restores empty AppVM containing no data, and errors contained in the output. 

Unable to extract files, unexpected EOF in archive tar, retrieved data backup size exceed expected size

7d496c01e2ff654edaf628a888af6336f19ed590

DemiMarie:
Volumes returned by `export()` must be immutable, since otherwise the
backup will be inconsistent.  Ensure this by exporting a snapshot of the
volume, no the volume itself.

Fixes QubesOS/qubes-issues#7198.

FIXME: tests

### DIFF
--- a/qubes/storage/lvm.py
+++ b/qubes/storage/lvm.py
@@ -156,7 +156,8 @@ class ThinPool(qubes.storage.Pool):
                 continue
             if vol_info['pool_lv'] != self.thin_pool:
                 continue
-            if vid.endswith('-snap') or vid.endswith('-import'):
+            if vid.endswith('-snap') or vid.endswith('-import') or \
+                    vid.endswith('+export'):
                 # implementation detail volume
                 continue
             if vid.endswith('-back'):
@@ -292,6 +293,7 @@ class ThinVolume(qubes.storage.Volume):
             self._vid_snap = self.vid + '-snap'
         if self.save_on_stop:
             self._vid_import = self.vid + '-import'
+        self._is_exporting = False
 
     @property
     def path(self):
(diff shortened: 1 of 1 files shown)


DemiMarie:
@marmarek how do you want me to deal with cleaning up the exported volumes?

marmarek:
> Volumes returned by `export()` must be immutable, since otherwise the
> backup will be inconsistent.

The volume currently used for backup _is_ immutable (in a sense that its content is not modified). VMs are writing to `-snap` snapshot (not the original volume), which is committed on VM shutdown.

While your change will make it more likely that VM size and data are taken from similar time, the behavior change here is actually harmful: it will backup _older_ data than it could. In fact, I consider changing the backup export to take a snapshot of a "dirty" volume to backup _newer_ version, not older. Theoretically, it should be sufficiently safe with journaled filesystem (as long as the snapshot operation itself is atomic).

Please note https://github.com/QubesOS/qubes-issues/issues/7198 in practice not about the time disk usage is recorder, but about recording size of a different thing. And also, the consensus there is to fix it differently - changing how limit is enforced.

DemiMarie:
> While your change will make it more likely that VM size and data are taken from similar time, the behavior change here is actually harmful: it will backup _older_ data than it could. In fact, I consider changing the backup export to take a snapshot of a "dirty" volume to backup _newer_ version, not older. Theoretically, it should be sufficiently safe with journaled filesystem (as long as the snapshot operation itself is atomic).

That is true, and even for non-journaled filesystems it should at most require a call to fsck.

> Please note [QubesOS/qubes-issues#7198](https://github.com/QubesOS/qubes-issues/issues/7198) in practice not about the time disk usage is recorder, but about recording size of a different thing. And also, the consensus there is to fix it differently - changing how limit is enforced.

Does this mean that I should close this PR as “won’t do”?

marmarek:
> Does this mean that I should close this PR as “won’t do”?

It may be a good base for the change I proposed above (which version to backup). But that needs a proper task description (extend https://github.com/QubesOS/qubes-issues/issues/1239?), and also consider other storage drivers.

DemiMarie:
Marking as draft for now.

It is worth noting that snapshotting in-use volumes breaks an invariant that currently holds, and which a direct dm-thin driver would need to consider.  But that is doable.

DemiMarie:
Closing as I won’t finish this.

PR #462, opened 2022-04-01
Have qubesd open the block device

DemiMarie:
This makes the absence of race conditions more obvious.  Also escape paths for the tar command.


DemiMarie:
Passes a simple manual test.

PR #466, opened 2022-04-13
Enable invtsc to let the guest use TSC clocksource, if it wishes so

marmarek:
The alternative ('xen') has rather poor performance. Since Qubes does
not support migration, it is safe to use TSC.

Fixes QubesOS/qubes-issues#7404


marmarek:
PipelineRetry

PR #470, opened 2022-04-23
fix TPM_s to populate into YAML

hardestkit:
Allowing TPM detection to populate YAML instead of only in terminal output. Please see discussion [https://forum.qubes-os.org/t/tpm-column-in-a-hcl-list/8565/3](https://forum.qubes-os.org/t/tpm-column-in-a-hcl-list/8565/3)


hardestkit:
Sorry, new account. GPG incoming soon.

hardestkit:
new PR with signed req incoming

PR #471, opened 2022-05-01
Avoid some “Got empty response from qubesd” errors

DemiMarie:


qubes/vm/__init__.py
```diff
@@ -475,7 +475,7 @@ def __set__(self, instance, value):
             if self.allow_none:
                 super().__set__(instance, value)
                 return
-            raise ValueError(
+            raise qubes.exc.QubesValueError(
```
marmarek, line 478: Should be `QubesPropertyValueError` instead, with proper constructor arguments.

qubes/vm/__init__.py
```diff
@@ -487,8 +487,8 @@ def __set__(self, instance, value):
             raise qubes.exc.QubesVMNotFoundError(value)
 
         if not isinstance(vm, self.vmclass):
-            raise TypeError('wrong VM class: domains[{!r}] is of type {!s} '
-                'and not {!s}'.format(value,
+            raise qubes.exc.QubesValueError('wrong VM class: '
```
marmarek, line 490: Same here.

marmarek (changes requested):

marmarek (approved):

PR #475, opened 2022-05-14
Enable volume downsize for lvm

a-barinov:


qubes/storage/lvm.py
```diff
@@ -764,9 +757,9 @@ def _get_lvm_cmdline(cmd):
         lvm_cmd = ['lvcreate', '--thin', '--setactivationskip=n',
                    '--activate=y', '--name=' + cmd[2],
                    '--virtualsize=' + cmd[3] + 'B', '--', cmd[1]]
-    elif action == 'extend':
+    elif action == 'resize':
         assert len(cmd) == 3, 'wrong number of arguments for extend'
```
DemiMarie, line 761:
```suggestion
        assert len(cmd) == 3, 'wrong number of arguments for resize'
```

DemiMarie (commented):
Minor nit, otherwise looks good code-wise.  That said, I think extending and arbitrary resizing should be separate from the Admin API perspective.

PR #477, opened 2022-05-21
Allow volume downsize requests

a-barinov:
What it does:
- Changes API to introduce volume.extend function that disallows volume downsize
- Removes downsize restrictions from volume.resize
- Adjusts lvm and file backends

What it does not do:
1. Adjusts reflink backend
2. Checks existing callers of volume.resize to use resize or extend call as needed

Once a change in API is conceptually agreed, (1) is hopefully easy to do. Finding 'resize' users across all repositories might be a bit more difficult but doable.

### DIFF
--- a/qubes/storage/__init__.py
+++ b/qubes/storage/__init__.py
@@ -355,10 +355,20 @@ class Volume:
         '''
         raise self._not_implemented("is_outdated")
 
-    async def resize(self, size):
-        ''' Expands volume, throws
+    async def extend(self, size):
+        ''' Extends volume, throws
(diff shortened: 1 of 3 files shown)


marmarek:
I think a better option is to simply drop the restriction at the storage classes level (keep a single `resize` method there), but add appropriate check at the frontend side (qubes-core-admin-client repo). There you could add additional `allow_shrink` or `force` argument that overrides the check; and then expose this argument in `qvm-volume` tool too.

a-barinov:
This sounds like a way better approach as less existing code needs to be touched. Will take this one down and create a new commit.

PR #478, opened 2022-05-28
Allow volume shrinking

a-barinov:
Change Volume.resize call to allow volume shrinking.
Bypass size check for file and lvm if shrinking is allowed.
Implement size check for reflink as it wasn't there.


DemiMarie (approved):
Looks good to me, assuming there are not any marshalling stubs that need changing.

a-barinov:
> Looks good to me, assuming there are not any marshalling stubs that need changing.

Well, almost. Seems like callbacks.py needs to be updated as well.

PR #479, opened 2022-05-28
Allow volume shrinking

a-barinov:
Change Volume.resize call to allow volume shrinking.
Bypass size check for file and lvm if shrinking is allowed.
Implement size check for reflink as it wasn't there.

### DIFF
--- a/qubes/storage/__init__.py
+++ b/qubes/storage/__init__.py
@@ -355,14 +355,17 @@ class Volume:
         '''
         raise self._not_implemented("is_outdated")
 
-    async def resize(self, size):
-        ''' Expands volume, throws
+    async def resize(self, size, allow_shrink=False):
+        ''' Resizes volume, throws
             :py:class:`qubes.storage.StoragePoolException` if
-            given size is less than current_size
+            given size is less than current_size and allow_shrink
+            is not True
 
             This can be implemented as a coroutine.
 
             :param int size: new size in bytes
+            :param bool allow_shrink: if shrinking volume is allowed. \
+                False by default.
         '''
         # pylint: disable=unused-argument
         raise self._not_implemented("resize")
--- a/qubes/storage/callback.py
+++ b/qubes/storage/callback.py
@@ -454,10 +454,10 @@ class CallbackVolume(qubes.storage.Volume):
         await self._callback('post_volume_remove')
         return ret
 
-    async def resize(self, size):
+    async def resize(self, size, allow_shrink=False):
         await self._assert_initialized()
         await self._callback('pre_volume_resize', cb_args=[size])
-        return await coro_maybe(self._cb_impl.resize(size))
+        return await coro_maybe(self._cb_impl.resize(size, allow_shrink))
 
     async def start(self):
         await self._assert_initialized()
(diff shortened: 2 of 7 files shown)


a-barinov:
This PR should not change anything in how volumes are currently resized. An the next step, patch to allow volume when restoring from backup will be submitted.

I see Codecov complaining about reduced coverage, let me know if this needs to be dealt with before PR is accepted.

marmarek:
> I see Codecov complaining about reduced coverage, let me know if this needs to be dealt with before PR is accepted.

A test for new resize parameter would be useful. See qubes/tests/storage_lvm.py for example - you can add something based on test_006_resize that attempt to shrink the volume (either with allow_shrink=True and False - and check for expected outcome).

rustybird:
Is the storage driver API really the right place to distinguish between extending and resizing? Duplicating this safety check in each driver looks strange to me. Why not something like:

1. Add a new admin API method `admin.vm.volume.Extend`, which would centralize the safety check and then call the volume object's existing `resize()` method
2. Wrap this new method as `qubesadmin.storage.Volume.extend()`, and switch existing callers of `qubesadmin.storage.Volume.resize()` over to it (except for cases like `qvm-volume resize --force`)
3. Make `ThinVolume.resize()` and `FileVolume.resize()` unrestricted like `ReflinkVolume.resize()`

marmarek:
> 2\. Wrap this new method as `qubesadmin.storage.Volume.extend()`, and switch existing callers of `qubesadmin.storage.Volume.resize()` over to it (except for cases like `qvm-volume resize --force`)

I'm a bit worried about this step - if you update core-admin but not core-admin-client, then you'll end up without the safety check at all. In dom0, a package dependency could be used to avoid this, but not in sys-gui. Maybe sys-gui is still rare enough to not care about it right now... ?

rustybird:
> In dom0, a package dependency could be used to avoid this, but not in sys-gui.

I hadn't thought of that... @a-barinov's approach of having the safety check in the storage driver might be the way to go then. That also avoids introducing a TOCTOU problem if there are concurrent API calls. (Is there something ReflinkVolume and FileVolume could use to enforce `allow_shrink=False` race free, similar to `lvextend` in ThinVolume? `os.posix_fallocate()` fits, but it probably has some side effects?)

a-barinov:
> A test for new resize parameter would be useful. See qubes/tests/storage_lvm.py for example - you can add something based on test_006_resize that attempt to shrink the volume (either with allow_shrink=True and False - and check for expected outcome).

Tried doing this, but can't find a way to re-run CI test to check if the tests I added do make sense / do not fail. Deleting this PR and creating a new one does not sound like a good solution.

marmarek:
> Tried doing this, but can't find a way to re-run CI test to check if the tests I added do make sense / do not fail.

We're doing some maintenance on CI, should be back online in few hours.

marmarek:
PipelineRetry

marmarek:
This looks good as is, but there needs to be further change to actually use the new parameter (expose it in api/admin.py, and then add support in qvm-volume tool). The backup case is handled by `ImportWithSize` method, so no need to do anything there.

PR #486, opened 2022-08-09
Tests for qubes-vm-update

piotrbartman:
https://github.com/QubesOS/qubes-core-admin-linux/pull/103

### DIFF
--- a/qubes/tests/integ/vm_update.py
+++ b/qubes/tests/integ/vm_update.py
@@ -201,6 +201,7 @@ class VmUpdatesMixin(object):
             self.upgrade_cmd = "apt-get -V dist-upgrade -y"
             self.install_cmd = "apt-get install -y {}"
             self.install_test_cmd = "dpkg -l {}"
+            self.update_test_cmd = "dpkg -l {} | grep 1.1"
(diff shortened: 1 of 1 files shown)


qubes/tests/integ/vm_update.py
```diff
@@ -412,49 +420,200 @@ def test_020_updates_available_notification(self):
...
 
         self.assertFalse(self.testvm1.features.get('updates-available', False))
 
+    def detach_netvm(self):
+        """
+        Detach net vm to enforce connection error.
+
+        :type self: qubes.tests.SystemTestCase | VmUpdatesMixin
+        """
+        self.loop.run_until_complete(self.testvm1.shutdown())
```
marmarek, line 589: Changing netvm dynamically is supported, so you don't necessarily need to shut it down. But if you still want to change netvm while VM is not running, then you probably want to add `wait=True`.

marmarek (changes requested):
Besides some comment below, you add and then remove vm_update_2.py - please adjust initial commit to not add it at all.

Note, I haven't tried to run it yet. But generally looks fine :)

marmarek (approved):

PR #487, opened 2022-08-30
comment out `/etc/qubes-rpc/policy/qubes.UpdatesProxy`

adrelanos:
Because legacy. Already replaced by `/etc/qubes/policy.d/90-default.policy`.

part of https://github.com/QubesOS/qubes-issues/issues/7724

### DIFF
--- a/qubes-rpc-policy/qubes.UpdatesProxy.policy
+++ b/qubes-rpc-policy/qubes.UpdatesProxy.policy
@@ -1,18 +1,2 @@
-## Note that policy parsing stops at the first match,
-## so adding anything below "$anyvm $anyvm action" line will have no effect
-
-## Please use a single # to start your custom comments
-
-# Upgrade all Templates through sys-whonix.
-#$type:TemplateVM $default allow,target=sys-whonix
-
-# Upgrade Whonix Templates through sys-whonix.
-$tag:whonix-updatevm $default allow,target=sys-whonix
-
-# Deny Whonix Templates using UpdatesProxy of any other VM.
-$tag:whonix-updatevm $anyvm deny
-
-# Default rule for all Templates - direct the connection to sys-net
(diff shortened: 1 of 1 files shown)


Minimalist73:
I was curious about this because I still use `/etc/qubes-rpc/policy/qubes.UpdatesProxy` to manage my updates trough Whonix. I discovered that the default rule in `/etc/qubes/policy.d/90-default.policy` is set to update all my Templates using sys-net even if I checked the "update trough Tor" in the Qubes installation (I did it few weeks ago). It means that if this PR is merged and the file is commented out, all my updates would go trough sys-net which will leak all the traffic to my ISP. I don't know if that's just me, but it needs to be checked before anything, because a lot of people can be in this situation.

Here's my `/etc/qubes/policy.d/90-default.policy` file:

```
# HTTP proxy for downloading updates
# Upgrade all TemplateVMs through sys-whonix.
#qubes.UpdatesProxy     *    @type:TemplateVM        @default    allow target=sys-whonix
# Upgrade Whonix TemplateVMs through sys-whonix.
qubes.UpdatesProxy      *   @tag:whonix-updatevm    @default    allow target=sys-whonix
# Deny Whonix TemplateVMs using UpdatesProxy of any other VM.
qubes.UpdatesProxy      *   @tag:whonix-updatevm    @anyvm      deny
# Default rule for all TemplateVMs - direct the connection to sys-net
qubes.UpdatesProxy      *   @type:TemplateVM        @default    allow target=sys-net
qubes.UpdatesProxy      *   @anyvm                  @anyvm      deny
```

adrelanos:
Not sure this should be discussed in the pull request or in the ticket? Ticket seems to have better visibility.

Minimalist73:
> Ticket seems to have better visibility.

You are right, I re-posted my message in the main ticket.

marmarek (approved):

PR #488, opened 2022-09-14
tests: salt function call and grains in jinja template

marmarek:
Regression test for saltstack/salt#62636 and saltstack/salt#61083


marmarek:
PipelineRetry

PR #489, opened 2022-09-20
Disable Hyper-V enlightenments in HVMs

DemiMarie:
This reduces hypervisor attack surface.


marmarek:
This is intentional for Windows performance.

marmarek:
In practice, this turned out to be no-op. Libvirt ignores this option (it's always disabled).

PR #491, opened 2022-10-01
Raise VM limit from 254 to 10000

marmarek:
Make max QID equal to max DispID. Technically QID could be up to 64770
(because of how IP addresses are assigned), but set the limit lower to
allow future adjustments to the IP calculation without being forced into
weird corner cases.

Fixes QubesOS/qubes-issues#2519

### DIFF
--- a/qubes/config.py
+++ b/qubes/config.py
@@ -75,7 +75,7 @@ defaults = {
     },
 }
 
-max_qid = 254
+max_qid = 10000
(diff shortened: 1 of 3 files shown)


marmarek:
PipelineRetry

rustybird:
https://github.com/QubesOS/qubes-issues/issues/2519#issuecomment-1264617756

PR #492, opened 2022-10-29
Validate service names

DemiMarie:
Service names are converted to paths in VMs, so make sure they start with a letter.  This (implicitly) catches service names that are not valid path names.  Furthermore, qubesdb keys are limited to 63 bytes, so limit service names to 48 bytes so that "/qubes-service/{service}" fits.


marmarek:
pylint complains

DemiMarie:
PipelineRetryFailed

marmarek (approved):

PR #495, opened 2022-11-06
Rely on size cache, not the /dev node

DemiMarie:
This might be the cause of QubesOS/qubes-issues#3964.

### DIFF
--- a/qubes/storage/lvm.py
+++ b/qubes/storage/lvm.py
@@ -408,7 +408,7 @@ class ThinVolume(qubes.storage.Volume):
             vid_to_commit = self._vid_snap
 
         assert self._lock.locked()
-        if not os.path.exists('/dev/' + vid_to_commit):
+        if vid_to_commit not in size_cache:
             # nothing to commit
             return False
 


marmarek:
The above issue is about `-root` and `-volatile` volumes, so the `commit()` function is unlikely related. But there is similar check in `qubes_lvm_coro()` when removing volume (which is called from both `_reset()` and `_snapshot()` - which are relevant for that issue).
That said, switching this to use cache may require refreshing cache in more places, and generally I'm not comfortable with relying on this case in start/stop logic, as that may require refreshing it several times during domain start/stop (at least one per volume, if not more), which would be bad for performance.

DemiMarie:
> The above issue is about `-root` and `-volatile` volumes, so the `commit()` function is unlikely related. But there is similar check in `qubes_lvm_coro()` when removing volume (which is called from both `_reset()` and `_snapshot()` - which are relevant for that issue). That said, switching this to use cache may require refreshing cache in more places, and generally I'm not comfortable with relying on this case in start/stop logic, as that may require refreshing it several times during domain start/stop (at least one per volume, if not more), which would be bad for performance.

I _really_ do not want to refresh cache any more than I have to, considering that I have seen `lvm` use a lot of CPU under `top`.  This might be a reason to switch to a different pool in R4.3.

marmarek:
Closing for reasons explained in https://github.com/QubesOS/qubes-core-admin/pull/495#issuecomment-1327518119

PR #500, opened 2022-11-23
Improve `admin.vm.volume.Import` service

DemiMarie:
The main improvement is that a qube cannot cause dom0 to use a huge amount of memory, but there are other improvements as well.  See individual commit messages for details.

### DIFF
--- a/qubes/api/__init__.py
+++ b/qubes/api/__init__.py
@@ -210,6 +210,15 @@ class AbstractQubesAPI:
         if not predicate:
             raise PermissionDenied()
 
+    def validate_size(self, untrusted_size: bytes) -> int:
+        self.enforce(isinstance(untrusted_size, bytes))
+        if not untrusted_size.isdigit():
+            raise qubes.api.ProtocolError('Size must be ASCII digits (only)')
+        if len(untrusted_size) >= 20:
+            raise qubes.api.ProtocolError('Sizes limited to 19 decimal digits')
+        if untrusted_size[0] == 48 and untrusted_size != b'0':
+            raise qubes.api.ProtocolError('Spurious leading zeros not allowed')
+        return int(untrusted_size)
 
 class QubesDaemonProtocol(asyncio.Protocol):
     buffer_size = 65536
--- a/qubes/tools/qubesd_query.py
+++ b/qubes/tools/qubesd_query.py
@@ -45,6 +45,11 @@ parser.add_argument('--single-line',
     action='store_true',
     default=False,
     help='Only read a single line of input')
+parser.add_argument('--null',
+    dest='null',
+    action='store_true',
+    default=False,
+    help="Print an extra NUL byte; useful for shell scripts")
 
 def sighandler(loop, signame, coro):
     print('caught {}, exiting'.format(signame))
@@ -157,6 +162,8 @@ def main(args=None):
 
     if args.fail:
         return returncode
+    if args.null:
+        sys.stdout.buffer.write(b'\0')
     return 0
 
 if __name__ == '__main__':
--- a/qubes/tools/qubesd_query.py
+++ b/qubes/tools/qubesd_query.py
@@ -73,6 +73,8 @@ async def qubesd_client(socket, payload, *args):
 
     try:
         header_data = await reader.read(1)
+        if not header_data.isdigit():
+            return 1
         returncode = int(header_data)
         sys.stdout.buffer.write(header_data)  # pylint: disable=no-member
         while not reader.at_eof():
--- a/qubes/tools/qubesd_query.py
+++ b/qubes/tools/qubesd_query.py
@@ -2,6 +2,7 @@
 
 import argparse
 import asyncio
+import os
 import signal
 import sys
 
@@ -35,6 +36,15 @@ parser.add_argument('dest', metavar='DEST',
 parser.add_argument('arg', metavar='ARGUMENT',
     nargs='?', default='',
     help='argument to method')
+parser.add_argument('--max-bytes',
+    dest='max_bytes',
+    type=int,
+    help='Maximum number of bytes to read from stdin')
+parser.add_argument('--single-line',
+    dest='single_line',
+    action='store_true',
+    default=False,
+    help='Only read a single line of input')
 
 def sighandler(loop, signame, coro):
     print('caught {}, exiting'.format(signame))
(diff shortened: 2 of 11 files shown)


marmarek:
```
DEBUG: Sphinx error:
DEBUG: Undocumented arguments for command 'qubesd-query': '--max-bytes, --single-line'
```

marmarek:
Any specific reason for adding `isdigit()` call before every `int()` on a string? The latter will reject non-numeric strings (also if there is just non-numeric prefix-suffix) with `ValueError` exception, which I think is handled in all the places you modified.

qubes/tools/qubesd_query.py
```diff
@@ -75,20 +87,56 @@ async def qubesd_client(socket, payload, *args):
...
 
+# pylint: disable=too-many-return-statements
 def main(args=None):
     args = parser.parse_args(args)
     loop = asyncio.get_event_loop()
+    max_payload_size = 1024 if args.single_line else MAX_PAYLOAD_SIZE
+
+    if args.max_bytes is not None:
+        if args.max_bytes > max_payload_size:
+            parser.error('Maximum payload too large (max {})'
```
marmarek, line 98: 64k is fine for the default limit, but may be a bit low for upper bound what is accepted at all. Anyway, preventing the user to raise the limit sounds pointless.

DemiMarie:
> Any specific reason for adding `isdigit()` call before every `int()` on a string?

The call to `isdigit()` should be on `bytes`, not on `str`, as the latter is Unicode-aware (which is not wanted).

> The latter will reject non-numeric strings (also if there is just non-numeric prefix-suffix) with `ValueError` exception, which I think is handled in all the places you modified.

`int()` accepts arbitrary Python base-10 integer literals, including unwanted ones such as literals with leading or trailing whitespace.

qubes/tools/qubesd_query.py
marmarek: Ah, right.

marmarek:
> `int()` accepts arbitrary Python base-10 integer literals, including unwanted ones such as literals with leading or trailing whitespace.

I don't think that's any practical problem here. Anyway, if you really insist, it sounds like this wants `validate_and_parse_int()` utility function.

qubes/api/__init__.py
```diff
@@ -210,6 +210,17 @@ def enforce(predicate):
         if not predicate:
             raise PermissionDenied()
 
+    @classmethod
```
marmarek, line 213: You call it on an object instance, not class type, so it should be normal method.

qubes/api/__init__.py
```diff
@@ -210,6 +210,17 @@ def enforce(predicate):
...
+    def validate_size(cls, untrusted_size: bytes) -> int:
+        cls.enforce(isinstance(untrusted_size, bytes))
+        if not untrusted_size.isdigit():
+            raise qubes.api.ProtocolError('Size must be ASCII digits (only)')
+        if len(untrusted_size) >= 20:
+            raise qubes.api.ProtocolError('Sizes limited to 19 decimal digits')
```
marmarek, line 219: this case is not covered by tests

qubes/api/__init__.py
```diff
@@ -210,6 +210,17 @@ def enforce(predicate):
...
+        try:
+            return int(untrusted_size)
+        except ValueError as e:
+            raise qubes.api.ProtocolError(e.args[0])
```
marmarek, line 223: and this too

qubes/tools/qubesd_query.py
```diff
@@ -75,20 +87,56 @@ async def qubesd_client(socket, payload, *args):
...
+                if 0x20 <= untrusted_c <= 0x7E:
+                    untrusted_v.append(untrusted_c)
+                else:
+                    parser.error('Invalid byte {} in input'.format(untrusted_c))
+                    # make sure to terminate, even if parser.error() would
+                    # return for some reason
+                    return 1
+            else:
+                parser.error('No newline found in first {} bytes'
+                             .format(args.max_bytes))
```
marmarek, line 125:
```suggestion
                             .format(max_payload_size))
```
`self.max_bytes` is `None` usually...

marmarek (changes requested):
Typo in the last commit message

qubes/tools/qubesd_query.py
```diff
@@ -35,6 +36,20 @@
...
+parser.add_argument('--single-line',
+    dest='single_line',
+    action='store_true',
+    default=False,
+    help='Only read a single line of input')
+parser.add_argument('--null',
+    dest='null',
+    action='store_true',
+    default=False,
+    help="Print an extra NUL byte; useful for shell scripts")
```
marmarek, line 52: Can you clarify what's the point of this? Not only the option, but also printing this byte at all. Preferably in man page's description of this option (which is missing, for other options added here too).

marmarek (changes requested):
Besides the comment below, unit tests fail.

qubes/tools/qubesd_query.py
DemiMarie: I added documentation for this option.  The other new options have documentation already IIUC.

qubes/tests/__init__.py
```diff
@@ -383,6 +383,39 @@ def __exit__(self, exc_type, exc_value, tb):
...
+        if exc_info is None:
+            continue
+        ex = exc_info[1]
+        while ex is not None:
+            if isinstance(ex, qubes.exc.QubesVMError):
+                ex.vm = None
+            traceback.clear_frames(ex.__traceback__)
+            ex = ex.__context__
+
+class QubesTestResult(unittest.TestResult):
```
marmarek, line 397: I appreciate your attempt at cleaning this up, but a) it does not belong to this PR, and b) this most likely will fail with alternative runners as they may provide own result class (we do use `nose2` and sometimes `pytest` example).

marmarek (changes requested):

marmarek:
```
************* Module qubes.tools.qubesd_query
qubes/tools/qubesd_query.py:164:10: E0001: Parsing failed: 'expected ':' (<unknown>, line 164)' (syntax-error)
```

qubes/tools/qubesd_query.py
```diff
@@ -75,20 +92,56 @@ async def qubesd_client(socket, payload, *args):
...
+                # make sure to terminate, even if parser.error() would return
+                # for some reason
+                return 1
+            payload = bytes(untrusted_v)
+            del untrusted_c, untrusted_v
+        else:
+            # read one byte more to check for too long payload,
+            # instead of silently truncating
+            payload = sys.stdin.buffer.read(MAX_PAYLOAD_SIZE + 1)
+            if len(payload) > args.max_bytes:
```
marmarek, line 140: Tests say:
```
  File "/usr/bin/qubesd-query", line 5, in <module>
    sys.exit(main())
  File "/usr/lib/python3.8/site-packages/qubes/tools/qubesd_query.py", line 140, in main
    if len(payload) > args.max_bytes:
TypeError: '>' not supported between instances of 'int' and 'NoneType'
```

marmarek (changes requested):

marmarek:
> @DemiMarie
[Avoid relying on undocumented unittest internals](https://github.com/QubesOS/qubes-core-admin/pull/500/commits/e725af8687017d2b6ff5df3658096a6851d4d287)
[e725af8](https://github.com/QubesOS/qubes-core-admin/pull/500/commits/e725af8687017d2b6ff5df3658096a6851d4d287)
@DemiMarie
[Wrap coroutine in Task before passing to asyncio.wait](https://github.com/QubesOS/qubes-core-admin/pull/500/commits/aafbf6c11f10ae2cea7b30ec7ee13accc7eb35c6)
[aafbf6c](https://github.com/QubesOS/qubes-core-admin/pull/500/commits/aafbf6c11f10ae2cea7b30ec7ee13accc7eb35c6)

I requested you to not mix those with other functional changes...

DemiMarie:
> > @DemiMarie
> > [Avoid relying on undocumented unittest internals](https://github.com/QubesOS/qubes-core-admin/pull/500/commits/e725af8687017d2b6ff5df3658096a6851d4d287)
> > [e725af8](https://github.com/QubesOS/qubes-core-admin/pull/500/commits/e725af8687017d2b6ff5df3658096a6851d4d287)
> > @DemiMarie
> > [Wrap coroutine in Task before passing to asyncio.wait](https://github.com/QubesOS/qubes-core-admin/pull/500/commits/aafbf6c11f10ae2cea7b30ec7ee13accc7eb35c6)
> > [aafbf6c](https://github.com/QubesOS/qubes-core-admin/pull/500/commits/aafbf6c11f10ae2cea7b30ec7ee13accc7eb35c6)
> 
> I requested you to not mix those with other functional changes...

oops, fixed

(My local branch includes these commits so that the tests work, but I should not have pushed them.)

DemiMarie:
PipelineRetryFailed

marmarek (approved):

PR #501, opened 2022-11-23
Add a --yaml-only option to hcl report

marmarta:
Cause: need to improve machine-readability
Also minor refactoring of the script to make it
less hellish to adjust.

### DIFF
--- a/qvm-tools/qubes-hcl-report
+++ b/qvm-tools/qubes-hcl-report
@@ -268,8 +268,7 @@ $SCSI
 usb: |
   $USB
 versions:
-
-- works:
+  works:
     'FIXME:yes|no|partial'
   qubes: |
     $QUBES
--- a/qvm-tools/qubes-hcl-report
+++ b/qvm-tools/qubes-hcl-report
@@ -100,7 +100,10 @@ if cat $TEMP_DIR/xl-dmesg | grep "$XL_DMESG_PREFIX_REGEX"'Xen version ' > /dev/n
     XL_DMESG_INCOMPLETE=no
 else
     XL_DMESG_INCOMPLETE=yes
-    echo -e 'WARNING: "xl dmesg" is incomplete. Some information are missing. Please reboot and try again.\n'
+    if ! [[ "$YAML_ONLY" ]]
+      then
+        echo -e 'WARNING: "xl dmesg" is incomplete. Some information are missing. Please reboot and try again.\n'
+    fi
 fi
 
 
@@ -221,7 +224,7 @@ TPM:\t\t$TPM
 Remapping:\t$REMAP
 "
 
-YAML_OUTPUT="---
+YAML_OUTPUT="
 layout:
   'hcl'
 qubes:
@@ -280,8 +283,6 @@ versions:
     FIXAUTHOR
   link: |
     FIXLINK
-
----
 "
 if [[ "$YAML_ONLY" == 1 ]]
   then
--- a/qvm-tools/qubes-hcl-report
+++ b/qvm-tools/qubes-hcl-report
@@ -224,6 +224,8 @@ Remapping:\t$REMAP
 YAML_OUTPUT="---
 layout:
   'hcl'
+qubes:
+  '$QUBES_VER'
 type:
   '$TYPE'
 hvm:
--- a/qvm-tools/qubes-hcl-report
+++ b/qvm-tools/qubes-hcl-report
@@ -44,7 +44,7 @@ while [ $# -gt 0 ]; do
       echo -e "Options are:"
       echo -e "\t-h, --help\tDisplay this help text and exit."
       echo -e "\t-s, --support\tGenerate more detailed HCL Support Files"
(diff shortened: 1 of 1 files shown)


marmarta:
@SvenSemmler , I'm making here some minor changes to qubes-hcl-report to make it less temperamental, I was told this might be of interest to you with regard to qubes-hcl repo.

SvenSemmler:
@marmarta I just downloaded and tested your new version and see some minor issues:

1. document start and end ('---') are missing
2. the line before gpu-short contains to spaces (yamllint complains about that)
3. the '-' in front of 'works' was removed, which means there can no longer be multiple versions in one YML file*

*makes no sense when generating them but when merging them into existing reports (same user/machine but different Qubes OS version).

marmarta:
> @marmarta I just downloaded and tested your new version and see some minor issues:
> 
>     1. document start and end ('---') are missing

Those are supposed to be used (according to yaml specs) to separate directives from document contents or multiple documents from each other, neither of which is applicable here. I mean, I can add the initial `---` back, it's not incorrect, but the trailing `---` is not correct according to specs.
> 
>     2. the line before gpu-short contains two spaces (yamllint complains about that)
Thanks, will fix that.
> 
>     3. the '-' in front of 'works' was removed, which means there can no longer be multiple 'versions' in one YML file*
> 
> 
> *makes no sense when generating them but when merging them into existing reports (same user/machine but different Qubes OS version).
Yeah, I did remove it because it did make no sense for a single report... What would multiple versions in one report even mean? Merging into multiple documents or a list of reports will still work, but multiple versions were quite illogical.

SvenSemmler:
@marmarta I have no strong feelings about any of this. It would be nice if the resulting YML makes it through yamllint without error or warning. Having multiple versions is a very frequent use case, but adding that '-' manually when merging is no big deal.

marmarta:
The only problem I see remaining in yamllint is too-long-lines sometimes, but I think this can't be helped with certain processors or GPUs? At least in my own hcl-report I get a very long chipset and gpu lines. I'd rather keep versions as a single dict, if possible.

SvenSemmler:
Agreed.

https://github.com/QubesOS/qubes-hcl/blob/master/.yamllint

;-)

SvenSemmler:
On 11/23/22 15:00, Marta Marczykowska-Górecka wrote:
> Yeah, I did remove it because it did make no sense for a single report... What would multiple versions in one report even mean? Merging into multiple documents or a list of reports will still work, but multiple versions were quite illogical.

I'm not sure about removing the '-' from 'works'. Leave 'versions:' but remove the '-'? What would that improve?

It is true that at the moment of generation there will just be one 'versions' entry, but then an active Qubes OS user will submit multiple versions for e.g. R3.2, R4.0 and R4.1 etc.

Everything before 'versions' will stay the same, except the user made hardware upgrades or updated the BIOS. BTW, if anything I'd put the BIOS version into the 'versions' record. But then again, that might break the code that generates the table on the website.

So I guess what I am trying to say is that by definition a HCL report has multiple versions, even though at the point of generation that list will always be just one entry (the one that was just generated). So if that is intended to be a list, why don't we leave it the way it is with the entry properly marked by '-' and maybe even indented correctly, since this is a thing I *always* have to fix with each and every report.

Like so:


versions:

   - works:
       'FIXME:yes|no|partial'
     qubes: |
       R4.1
     xen: |
       4.14.5
     kernel: |
       5.15.76-1
     remark: |
       FIXME
     credit: |
       FIXAUTHOR
     link: |
       FIXLINK

marmarta:
I think the key difference between our perceptions is the use for the HCL report :D I started touching it up because I'm doing a GUI tool to show general system state, including things like "so how much RAM do I actually have" and other info that might be useful and is not that obvious how to get for a slightly-less-technical user. A given instance of a report file has only one version, because, well, it's been generated by a single instance of a running Qubes. 

I can add the list there for the aggregation purposes, though - also, I can fix the indenting, sure.

marmarta:
Alright, I think the yaml indents are better now, and I also did the list in versions in the way you proposed; should be working better now, but if you have any more problems with that yaml let me know

SvenSemmler:
I re-ran your latest script and noticed that the `qubes` field now exists twice ...

1.  between `layout` and `type` with more details ... e.g. `4.1.1 (R4.1)`
2.  between `works` and `xen` in the `versions` record without the additional detail

I don't see the rationale for including it twice nor for having it in the machine/hardware details part of the report. Is this a mistake or intentional? 

I love the added detail. Can we do `R4.1.1` instead of `4.1.1 (R4.1)`? Here are all the Qubes OS revisions in the existing reports:

R1.0
R2.0
R2.0-Beta1
R2.0-Beta2
R2.0-Beta3
R2.0-RC1
R2.0-RC2
R3.0
R3.0-RC1
R3.1
R3.2
R4.0
R4.0.4
R4.0.1
R4.0.1-RC2
R4.0.2-RC1
R4.1
R4.1-Beta
R4.1-RC1
R4.1-RC2
R4.1-RC3
R4.1-RC4

marmarek (changes requested):
See last comment from @SvenSemmler

PR #502, opened 2022-11-30
Replace ImageMagick with GraphicsMagick

mati7337:
https://github.com/QubesOS/qubes-issues/issues/5009

### DIFF
--- a/qubes/tests/integ/mime.py
+++ b/qubes/tests/integ/mime.py
@@ -185,7 +185,7 @@ class TC_50_MimeHandlers:
 
     def prepare_pdf(self, filename):
         self.prepare_txt("/tmp/source.txt")
-        cmd = "convert text:/tmp/source.txt {}".format(filename)
+        cmd = "gm convert text:/tmp/source.txt {}".format(filename)
         try:
             self.loop.run_until_complete(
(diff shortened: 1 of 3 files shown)


DemiMarie:
GraphicsMagick still has a very large attack surface.  Could something that is _not_ ImageMagick or GraphicsMagick be used, such as libvips?

marmarek:
First, in this repo we use it only for tests, so its attack surface is rather irrelevant. But then, in places where it can process untrusted data (like in https://github.com/QubesOS/qubes-app-linux-pdf-converter/pull/24), we explicitly request `rgb` or `rgba` format. I think it's reasonable to assume it's too simple to be screwed up in something that is a mature software.

DemiMarie:
> First, in this repo we use it only for tests, so its attack surface is rather irrelevant. But then, in places where it can process untrusted data (like in [QubesOS/qubes-app-linux-pdf-converter#24](https://github.com/QubesOS/qubes-app-linux-pdf-converter/pull/24)), we explicitly request `rgb` or `rgba` format. I think it's reasonable to assume it's too simple to be screwed up in something that is a mature software.

True.  Perhaps the more annoying thing is that both ImageMagick and GraphicsMagick are overkill for this task and pull in dependencies we don’t need.

PR #506, opened 2022-12-26
qmemman use memory hotplug

marmarek:
If VM supports memory hotplug, start it with maxmem=memory and then use 
memory hotplug to (potentially) increase its memory. Qmemman almost had all
the needed parts - especially mem_set function already adjusts maxmem too.
The missing part was letting it know actual maxmem - do that via xenstore
(as qmemman uses xenstore heavily already).

The use of memory hotplug can be enabled (or disabled) explicitly via
'memory-hotplug' feature, or detected based on
'memory-hotplug-supported' file in dom0-provided kernel dir, or
'supported-feature.memory-hotplug' for VM-provided kernel.

Fixes QubesOS/qubes-issues#7956


qubes/vm/qubesvm.py
```diff
@@ -1627,6 +1627,27 @@ def is_memory_balancing_possible(self):
...
+        The support needs to be enabled in the VM's kernel.
+        """
+        feature = self.features.check_with_template('memory-hotplug', None)
+        if feature is not None:
+            return bool(feature)
+        # if not explicitly set, check if support is advertised
+        # for dom0-provided kernel - check there
+        if self.kernel is not None:
+            return os.path.exists(os.path.join(
+                self.storage.kernels_dir, 'memory-hotplug-supported'))
```
DemiMarie, line 1643:
```suggestion
            try:
                os.stat(os.path.join(
                    self.storage.kernels_dir, 'memory-hotplug-supported'))
                return True
            except FileNotFoundError:
                pass
```

Other errors should not be swallowed.

qubes/vm/qubesvm.py
```diff
@@ -1627,6 +1630,28 @@ def is_memory_balancing_possible(self):
                 return False
         return True
 
+    @property
+    def use_memory_hotplug(self):
+        """Use memory hotplug for memory balancing.
+        This is preferred if supported, because it has less initial overhead
+        and reduces Xen's attack surface.
+        The support needs to be enabled in the VM's kernel.
```
DemiMarie, line 1638:
```suggestion
        This must be supported by the VM's kernel.        
```

DemiMarie (commented):
Grammer nit

marmarek:
This may require some more work on the kernel side. VM started with initial 400MB (the default) cannot reach 4GB, it stops around 2924MB.

marmarek:
> This may require some more work on the kernel side. VM started with initial 400MB (the default) cannot reach 4GB, it stops around 2924MB.

Should be fixed with https://github.com/QubesOS/qubes-core-libvirt/pull/47

PR #507, opened 2023-01-01
Avoid relying on undocumented unittest internals

DemiMarie:
The previous code relied on undocumented implementation details of the unittest module to clear local variables from tracebacks.  These implementation details changed in Python 3.11, causing the tests to fail.  Instead, subclass unittest.TestResult and override the methods that take stack traces as arguments.  This avoids having to rely on undocumented Python internals.


marmarek:
Duplicate of #505 ?

DemiMarie:
I closed #505 in favor of this.

marmarek:
I tested it with various runners we use, and it doesn't work with any of them. They provide explicit result instance to the `run()` method. I have an alternative idea.

PR #508, opened 2023-01-01
Wrap coroutine in Task before passing to asyncio.wait

DemiMarie:
Needed in Python 3.11+.

### DIFF
--- a/qubes/utils.py
+++ b/qubes/utils.py
@@ -269,7 +269,7 @@ async def void_coros_maybe(values):
         with each other. If there were exceptions, raise the leftmost
         one (not necessarily chronologically first). Return nothing.
     '''
-    coros = [val for val in values if asyncio.iscoroutine(val)]
+    coros = [asyncio.create_task(val) for val in values if asyncio.iscoroutine(val)]
     if coros:
         done, _ = await asyncio.wait(coros)
         for task in done:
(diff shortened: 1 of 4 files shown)


rustybird:
Just in case you haven't already seen this and decided it's okay, the [documentation for `asyncio.Task()`](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task) says

> Use the high-level [`asyncio.create_task()`](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task) function to create Tasks, or the low-level [`loop.create_task()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_task) or [`ensure_future()`](https://docs.python.org/3/library/asyncio-future.html#asyncio.ensure_future) functions. Manual instantiation of Tasks is discouraged.

qubes/utils.py
```diff
@@ -269,7 +269,7 @@ async def void_coros_maybe(values):
         with each other. If there were exceptions, raise the leftmost
         one (not necessarily chronologically first). Return nothing.
     '''
-    coros = [val for val in values if asyncio.iscoroutine(val)]
+    coros = [asyncio.Task(val) for val in values if asyncio.iscoroutine(val)]
```
marmarek, line 272:
```suggestion
    coros = [asyncio.create_task(val) for val in values if asyncio.iscoroutine(val)]
```

marmarek (changes requested):

marmarek (approved):

PR #510, opened 2023-01-02
Block not script

DemiMarie:


not-script/not-script.c
```diff
@@ -0,0 +1,414 @@
...
+                errx(1, "Bad path %s", path);
+
+            extra_path = path + 5;
+
+            qdb_handle_t h = qdb_open(NULL);
+
+            if (!h)
+                err(1, "qdb_open(NULL)");
+
+            if (asprintf(&qubesdb_path, "/qubes-block-devices/%s/desc", extra_path) < 0)
```
marmarek, line 151: I doubt checking against qubesdb is helpful at all. For accidental race to happen, user action is very likely to be involved (and user wouldn't do it between clicking attach and the actual attach). And even if it would, then udev trigger is very likely to update the qubesdb entry before this check happens (the window when `/dev/loop*` points at a different thing than qubesdb says is _very_ small).
As for malicious race (backend trying to make user attach a different thing), this won't help at all, as backend process (especially capable of changing loop devices) can simply replace file content.

What could be useful, is to generate names that are somehow bound to the content (like `loop-XXXX` where `XXXX` is some short checksum of the target path, even crc32 would do), but that's out of scope of this PR.

DemiMarie: I will submit a PR to use `X-old-name`, where `X` is the device diskseq.  The not-script can check for that.  At some point I want to rewrite the udev trigger script in something other than shell, perhaps C.

DemiMarie:
PipelineRetry

marmarek:
FYI I cherry-picked the last two commits to the main branch.

marmarek:
This should be marked as ready, right?

marmarek (changes requested):
(this should really be a comment to the already merged PR in linux-utils)

The log is full of `/etc/xen/scripts/qubes-block remove [12191] exited with error status 1`, and xen-hotplug.log has:
```
qubes-block: Bad physical device value 7:135
qubes-block: Bad physical device value 7:138
qubes-block: Bad physical device value fd:2d
qubes-block: Bad physical device value fd:2f
qubes-block: Bad physical device value 7:13e
qubes-block: Bad physical device value 7:13f
qubes-block: Bad physical device value fd:2a
qubes-block: Bad physical device value fd:2c
qubes-block: Bad physical device value 7:139
qubes-block: Bad physical device value 7:13c
qubes-block: Bad physical device value fd:2d
qubes-block: Bad physical device value fd:2f
qubes-block: Bad physical device value 7:142
qubes-block: Bad physical device value 7:143
qubes-block: Bad physical device value fd:2a
qubes-block: Bad physical device value fd:2c
qubes-block: Bad physical device value 7:13d
qubes-block: Bad physical device value 7:140
qubes-block: Bad physical device value fd:2a
qubes-block: Bad physical device value fd:2c
qubes-block: Bad physical device value 7:141
qubes-block: Bad physical device value 7:144
```

DemiMarie:
Ugh, that is clearly a bug.  Will fix.

templates/libvirt/devices/block.xml
```diff
@@ -2,20 +2,21 @@
...
 
     {%- if options.get('read-only', 'no') == 'yes' %}
-        <readonly />
+    <readonly />
     {%- endif %}
 
     {%- if device.backend_domain.name != 'dom0' %}
     <backenddomain name="{{ device.backend_domain.name }}" />
     {%- endif %}
+    <script path="/etc/xen/scripts/qubes-block" />
```
marmarek, line 21: There are a few issues with using this not-script for qvm-block:
1. It will remove the loop device on detach, even if it didn't created it in the first place (pre-existing loop dev).
2. It requires the backend to use template from R4.2 (in other words: breaks compatibility with templates migrated from older system but not upgraded to R4.2 yet).

The second one is not a big issue in practice - while we try to not break existing features with new releases within the same major version, I suspect block backend outside of dom0 is mostly used in sys-usb, and there is little motivation to migrate system qubes from older version (especially when it's disposable). That said, R4.2 finally [advertises qubes tools version inside template](https://github.com/QubesOS/qubes-core-admin-client/blob/main/doc/manpages/qvm-features.rst#qubes-agent-version), so you could put the above under `{% if device.backend_domain.check_with_template('qubes-agent-version', '4.1') >= '4.2' %}` (`>=` on strings wont work for `4.10`, but I don't think we'll reach such high versions). The dom0 backend case don't need this check (which means either duplicated script line, or longer condition).

But the first issue needs fixing if this line is to stay. You need some way to distinguish if loop device was created by the qubes-block. Maybe compare `physical-device-path` xenstore entry against `params`?

(this issue is the cause of all the failures in `system_tests_devices` job - it creates loop device, and then does qvm-block attach, qvm-block detach, qvm-block attach - the last attach fails)

marmarek (changes requested):

marmarek:
> That said, R4.2 finally [advertises qubes tools version inside template](https://github.com/QubesOS/qubes-core-admin-client/blob/main/doc/manpages/qvm-features.rst#qubes-agent-version), so you could put the above under {% if device.backend_domain.check_with_template('qubes-agent-version', '4.1') >= '4.2' %} (>= on strings wont work for 4.10, but I don't think we'll reach such high versions). The dom0 backend case don't need this check (which means either duplicated script line, or longer condition).

This is still an issue, and since it's (finally) easy to fix, IMO we should.

templates/libvirt/devices/block.xml
```diff
@@ -2,20 +2,25 @@
...
     {%- endif %}
 
     {%- if options.get('read-only', 'no') == 'yes' %}
-        <readonly />
+    <readonly />
     {%- endif %}
 
     {%- if device.backend_domain.name != 'dom0' %}
     <backenddomain name="{{ device.backend_domain.name }}" />
+    {%- if device.backend_domain.check_with_template('qubes-agent-version', '4.1') >= '4.2' %}
```
marmarek, line 20: Sorry, that should be `device.backend_domain.features.check_with_template`

marmarek (changes requested):

marmarek (approved):

PR #515, opened 2023-01-19
tests: move traceback cleanup to test method wrappers

marmarek:
Python 3.11 changed internal details that cleanup_traceback() relied on.
Move the functionality to wrappers around actual test functions. This
way, it's contained within test class, and doesn't depend on any
specific implementation of test runner.
Wrap those methods dynamically in QubesTestCase.__init__(), to handle
any overriden methods in subclasses too. This avoids needing to use
explicit decorators for every single test method.

This replaces #507

### DIFF
--- a/qubes/tests/__init__.py
+++ b/qubes/tests/__init__.py
@@ -32,6 +32,7 @@
 
 import asyncio
 import collections
+import contextlib
 import functools
 import logging
 import os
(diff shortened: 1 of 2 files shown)


marmarek:
Tested with standard unittest runner, `qubes.tests.run`, nose2 and pytest. In all cases tracebacks seems to be correctly cleanup up now.

PR #519, opened 2023-02-12
PipeWire integration test support

DemiMarie:
This adds integration tests for PipeWire.

### DIFF
--- a/qubes/tests/integ/vm_qrexec_gui.py
+++ b/qubes/tests/integ/vm_qrexec_gui.py
@@ -909,15 +909,15 @@ class TC_10_Generic(qubes.tests.SystemTestCase):
             ' qrexec-client-vm output: {} {}'.format(stdout, stderr))
 
 def create_testcases_for_templates():
-    yield from qubes.tests.create_testcases_for_templates('TC_00_AppVM',
-        TC_20_AudioVM_Pulse, qubes.tests.SystemTestCase,
-        module=sys.modules[__name__])
-    yield from qubes.tests.create_testcases_for_templates('TC_00_AppVM',
-        TC_20_AudioVM_PipeWire, qubes.tests.SystemTestCase,
-        module=sys.modules[__name__])
-    yield from qubes.tests.create_testcases_for_templates('TC_00_AppVM',
-        TC_20_NonAudio, qubes.tests.SystemTestCase,
-        module=sys.modules[__name__])
+    yield from qubes.tests.create_testcases_for_templates(
+        'TC_20_AudioVM_Pulse', TC_20_AudioVM_Pulse,
+        qubes.tests.SystemTestCase, module=sys.modules[__name__])
+    yield from qubes.tests.create_testcases_for_templates(
+        'TC_20_AudioVM_PipeWire', TC_20_AudioVM_PipeWire,
+        qubes.tests.SystemTestCase, module=sys.modules[__name__])
+    yield from qubes.tests.create_testcases_for_templates(
+        'TC_20_NonAudio', TC_20_NonAudio,
+        qubes.tests.SystemTestCase, module=sys.modules[__name__])
 
 def load_tests(loader, tests, pattern):
     tests.addTests(loader.loadTestsFromNames(
--- a/qubes/tests/integ/vm_qrexec_gui.py
+++ b/qubes/tests/integ/vm_qrexec_gui.py
@@ -138,13 +138,15 @@ systemctl --user start {start_units}""")
             self.fail('frequency {} not in specified range'
                     .format(rec_freq))
 
-    def common_audio_playback(self):
+    def common_audio_playback(self, pulse: bool):
         # sine frequency
         sfreq = 4400
         # generate signal
         audio_in = np.sin(2*np.pi*np.arange(44100)*sfreq/44100)
+        # Need to use .snd extension so that pw-play (really libsndfile)
+        # recognizes the file as raw audio.
         self.loop.run_until_complete(
-            self.testvm1.run_for_stdio('cat > audio_in.raw',
+            self.testvm1.run_for_stdio('cat > audio_in.snd',
             input=audio_in.astype(np.float32).tobytes()))
         local_user = grp.getgrnam('qubes').gr_mem[0]
         with tempfile.NamedTemporaryFile() as recorded_audio:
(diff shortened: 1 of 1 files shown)


qubes/tests/integ/vm_qrexec_gui.py
```diff
@@ -436,15 +436,32 @@ def wait_for_pulseaudio_startup(self, vm):
...
+        elif backend == 'pipewire':
+            start_units = pipewire_units
+            stop_units = pulseaudio_units
+            manage_cmd = 'touch'
+        else:
+            self.fail('bad audio backend')
         try:
-            self.loop.run_until_complete(
-                self.testvm1.run_for_stdio('which parecord'))
+            self.loop.run_until_complete("which parecord")
```
marmarek, line 456: `parecord` is executed in VM, not only in dom0, so the check should remain as it was. The check for dom0 is done via test decorator.

marmarek (changes requested):
You need to update `create_testcases_for_template()` for new classes split.

qubes/tests/integ/vm_qrexec_gui.py
```diff
@@ -65,6 +65,262 @@ def setUp(self):
...
+            manage_cmd = 'touch'
+            packages = 'wireplumber pipewire-pulseaudio'
+        else:
+            self.fail('bad audio backend')
+        try:
+            self.loop.run_until_complete("which parecord")
+        except subprocess.CalledProcessError:
+            self.skipTest('pulseaudio-utils not installed in VM')
+        try:
+            self.testvm1.run_for_stdio(f"""sudo {manage_cmd} /run/qubes-service/pipewire &&
```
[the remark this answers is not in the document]
DemiMarie: Will that leave the legacy agent untested?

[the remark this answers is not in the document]
marmarek: No, legacy agent will get tested by running this test (pulseaudio variant) on a template with either just pulseaudio installed, or both but without pipewire-pulseaudio (since it defaults to pulseaudio then, unless "pipewire" qvm-service is enabled).

qubes/tests/integ/vm_qrexec_gui.py
```diff
@@ -65,6 +65,279 @@ def setUp(self):
...
+                         "pulseaudio-utils not installed in dom0")
+    def test_221_audio_rec_muted_pulseaudio(self):
+        self.prepare_audio_vm('pulseaudio')
+        self.common_audio_record_muted()
+
+    @unittest.skipUnless(spawn.find_executable('parecord'),
+                         "pulseaudio-utils not installed in dom0")
+    def test_222_audio_rec_unmuted_pulseaudio(self):
+        self.prepare_audio_vm('pulseaudio')
+        self.common_audio_record_unmuted(False)
```
marmarek, line 287: and here pulse=False in pulseaudio test

qubes/ext/core_features.py
```diff
@@ -36,7 +36,8 @@ async def qubes_features_request(self, vm, event, untrusted_features):
 
         requested_features = {}
         for feature in (
-                'qrexec', 'gui', 'gui-emulated', 'qubes-firewall', 'vmexec'):
+                'qrexec', 'gui', 'gui-emulated', 'qubes-firewall', 'vmexec',
+                'pipewire'):
```
marmarek, line 40: You don't need to add it here. `supported-service.*` features are handled by "services" extension already. Here you'd handle plain `pipewire` but it isn't what you need.

marmarek (changes requested):

marmarek:
> You need to update create_testcases_for_template() for new classes split.

You forgot about this part, so the test didn't run at all :/

qubes/tests/integ/vm_qrexec_gui.py
```diff
@@ -909,8 +909,14 @@ def test_000_anyvm_deny_dom0(self):
             ' qrexec-client-vm output: {} {}'.format(stdout, stderr))
 
 def create_testcases_for_templates():
-    return qubes.tests.create_testcases_for_templates('TC_00_AppVM',
-        TC_00_AppVMMixin, qubes.tests.SystemTestCase,
+    yield from qubes.tests.create_testcases_for_templates('TC_00_AppVM',
```
marmarek, line 912: The new class prefix must be different for each class too - usually I match them with the actual base class name

marmarek (changes requested):
Full test run:
 - with pipewire: https://openqa.qubes-os.org/tests/68923 (debian-11 lacks pipewire-pulse there, so look at fedora mostly)
 - without pipewire: https://openqa.qubes-os.org/tests/68829

qubes/tests/integ/vm_qrexec_gui.py
```diff
@@ -65,6 +65,262 @@ def setUp(self):
...
+        self.common_audio_record_unmuted()
+
+    @unittest.skipUnless(spawn.find_executable('parecord'),
+                         "pulseaudio-utils not installed in dom0")
+    def test_223_audio_play_hvm(self):
+        self.testvm1.virt_mode = 'hvm'
+        self.testvm1.features['audio-model'] = 'ich6'
+        self.prepare_audio_vm('pulseaudio')
+        self.loop.run_until_complete(
+            self.testvm1.run_for_stdio('pacmd unload-module module-vchan-sink'))
```
marmarek, line 279: This fails with `pipewire-pulseaudio`, which isn't really surprising. I'm okay (at least for now) with skipping this test (and the two below) if no real pulseaudio is installed. The purpose of this test is checking audio passthrough through stubdomain, regardless of the audio stack in the VM. But if there is a simple method to ensure pipewire will use emulated sound card instead of pipewire-qubes module, you can try that too.

marmarek (changes requested):
Tests for pulseaudio (both real and pipewire-pulseaudio) seems to work now. Native pipewire is still running, so not sure yet. But I already see the HVM audio tests need some fix or a skip - see below.

marmarek:
FYI, the latest test run have saved the bad sample (the one with ~2000 Hz): https://openqa.qubes-os.org/tests/71361/file/system_tests-audio-sample-qubes.tests.integ.vm_qrexec_gui.tar.gz
It is heavily distorted. In fact, it sounds like it's in a wrong format. I'll experiment with it a bit.

qubes/tests/integ/vm_qrexec_gui.py
```diff
@@ -65,6 +65,276 @@ def setUp(self):
...
+        # generate signal
+        audio_in = np.sin(2*np.pi*np.arange(44100)*sfreq/44100)
+        # Need to use .snd extension so that pw-play (really libsndfile)
+        # recognizes the file as raw audio.
+        self.loop.run_until_complete(
+            self.testvm1.run_for_stdio('cat > audio_in.snd',
+            input=audio_in.astype(np.float32).tobytes()))
+        local_user = grp.getgrnam('qubes').gr_mem[0]
+        if self.testvm1.features['service.pipewire']:
+            cmd = 'pw-play --format=f32 --rate=44100 --channels=1 audio_in.snd'
```
marmarek, line 142: This looks to not do the desired thing, With `-v` it prints `PCM: fmt:ulaw rate:8000 channels:1 width: 1`. If I understand `--help` correctly, those explicit options are used only for recording, but otherwise `pw-play` tries to find and parse a header (and here it happened to interpret it this way).
After converting the raw samples to proper wav (`sox -t f32 -r 44100 -c 1 input.snd output.wav`), it works. I don't really like this extra dependency, so I'll see if maybe prepending hand-crafted header is enough. But if that would turn too complex, I'll install `sox` for the tests.

marmarek: Oh, it works if the data is passed via stdin, instead of file argument!

marmarek, line 142: Here is the fix:
```suggestion
            cmd = 'pw-play --format=f32 --rate=44100 --channels=1 - < audio_in.snd'
```
 :)

marmarek (approved):

PR #520, opened 2023-02-13
More fixes for Python 3.11

marmarek:


qubes/tests/__init__.py
```diff
@@ -389,10 +389,17 @@ class _clear_ex_info(contextlib.ContextDecorator):
...
+        self._result_callback = result_callback
+
     def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
+        if self._result_callback:
+            self._result_callback(bool(
+                not exc_type or isinstance(exc_type, unittest.SkipTest)
+            ))
```
DemiMarie, line 402: Why the call to `bool` here?  It looks redundant.

marmarek: Earlier version needed it, now can be removed.

PR #524, opened 2023-03-10
Require features to start with an ASCII letter

DemiMarie:
This prevents any option injection naughtiness.


marmarek:
This PR breaks many (if not most?) usages of this API. First of all, the most common values are `0` and `1` - both rejected by your change. Then, empty value is perfectly fine too (in qvm-features convention it means "false"). And finally, value can very well starts with `-` or whatever else. Features requested by this API by a VM are _not_ automatically saved as actual features, there need to be an extension that handles specific values and saves only them after validation. If `--something` (or such) value is inappropriate for a specific feature, extension saving it should validate it (I don't think any currently implemented extension would allow such value anyway).

PR #530, opened 2023-04-20
HCL report generator: Do not hard-code /home/user

DemiMarie:
The default user might not be named "user".  Also massively cleanup the script (tested locally on my laptop).

### DIFF
--- a/qvm-tools/qubes-hcl-report
+++ b/qvm-tools/qubes-hcl-report
@@ -1,6 +1,6 @@
-#!/bin/bash
+#!/bin/bash --
 
-# The Qubes OS Project, http://www.qubes-os.org
+# The Qubes OS Project, https://www.qubes-os.org
 #
 # Copyright (C) 2013  Laszlo Zrubecz <mail@zrubi.hu>
 #
@@ -17,10 +17,11 @@
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library; if not, see <https://www.gnu.org/licenses/>.
 
+set -uo pipefail
 VERSION=2.5
 COPY2VM="dom0"
 SUPPORT_FILES=0
-
+YAML_ONLY=0
 XL_DMESG_PREFIX_REGEX='^(XEN) \(\[[^]]*\] \)\?'
 
 while [ $# -gt 0 ]; do
@@ -34,7 +35,27 @@ while [ $# -gt 0 ]; do
       YAML_ONLY=1
       ;;
 
-    -* | -h | --help)
+    ''|--|[!-]*)
+      if [ "$1" = '--' ]; then shift; fi
+      case $# in
+      (0) :;;
+      (1)
+        /usr/bin/qvm-check -q -- "$1"
+        if [[ $? -eq 0 ]]
+        then
+          COPY2VM="$1"
+        else
+          echo -e "ERROR:\tAppVM with the name '$1' does not exist in the system!"
+          exit 1
+        fi
+        ;;
+      (*)
+        echo -e "ERROR:\\tToo many non-option arguments (expected 0 or 1, got $#)"
+        ;;
+      esac
+      ;;
+
+    *)
       echo -e "qubes-hcl-report v$VERSION"
       echo ""
       echo "This tool is used to gather basic hardware information for the Qubes HCL (Hardware Compatibility List)"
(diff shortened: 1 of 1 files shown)


qvm-tools/qubes-hcl-report
```diff
@@ -69,34 +70,104 @@ while [ $# -gt 0 ]; do
...
         exit 1
     fi
 fi
 
 
-DATE=`date +%Y%m%d-%H%M%S`
-
-
-TEMP_DIR=`mktemp --tmpdir -d HCL.XXXXXXXXXX`
-cat /etc/qubes-release > $TEMP_DIR/qubes-release
```
marmarek, line 91: All used files should still be copied into temp dir, otherwise the support archive will be incomplete.

qvm-tools/qubes-hcl-report
```diff
@@ -69,34 +70,104 @@ while [ $# -gt 0 ]; do
...
-cat /proc/cpuinfo > $TEMP_DIR/cpuinfo
-sudo lspci -nnvk > $TEMP_DIR/lspci
-cat /proc/scsi/scsi > $TEMP_DIR/scsi
-sudo dmidecode > $TEMP_DIR/dmidecode
```
marmarek, line 95: Same comment as above

qvm-tools/qubes-hcl-report
```diff
@@ -107,33 +178,22 @@ else
...
-USB=`lspci|grep USB|wc -l`
-BIOS=`cat $TEMP_DIR/dmidecode |grep -A9 "BIOS Information" |grep "Version" |cut -d ' ' -f2-`
-XEN_MAJOR=`cat $TEMP_DIR/xl-info |grep xen_major |cut -d: -f2 |tr -d ' '`
-XEN_MINOR=`cat $TEMP_DIR/xl-info |grep xen_minor |cut -d: -f2 |tr -d ' '`
-XEN_EXTRA=`cat $TEMP_DIR/xl-info |grep xen_extra |cut -d: -f2 |tr -d ' '`
-XL_VTX=`cat $TEMP_DIR/xl-info |grep xen_caps | grep hvm`
-XL_VTD=`cat $TEMP_DIR/xl-info |grep virt_caps |grep hvm_directio`
-XL_HAP=`cat $TEMP_DIR/xl-dmesg |grep "$XL_DMESG_PREFIX_REGEX"'HVM: Hardware Assisted Paging (HAP) detected\( but disabled\)\?$'`
-XL_REMAP=`cat $TEMP_DIR/xl-dmesg |grep "$XL_DMESG_PREFIX_REGEX"'\(Intel VT-d Interrupt Remapping enabled\|Interrupt remapping enabled\)'`
+QUBES_VER=$(cut -d ' ' -f3 /etc/qubes-release)
```
marmarek, line 181: For consistency, use files from the temp dir (it makes it easier to test the script on files retrieved from somewhere else). Same comment to few instances below.

marmarek (changes requested):

qvm-tools/qubes-hcl-report
```diff
@@ -17,10 +17,11 @@
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library; if not, see <https://www.gnu.org/licenses/>.
 
+set -euo pipefail
```
marmarek, line 20: Adding `set -e` here means you should review every command in the script and consider if it failing should really be fatal. In a script like this, I'd prefer to have some field missing if it wasn't detected than having no output at all (which is the case right now). Because of that, IMO better to _not_ use `set -e`.

The cases I found already:
 - network devices (you may have none on PCI)
 - same for USB (although unlikely)
 - SCSI (this I'm actually surprised worked for you, do you really have some SCSI disk?)
 - HAP and Interrupt Remapping - when those are missing, it should report "no", instead of crashing the script
 - same for VT-x, VT-d

There are probably more.

marmarek (changes requested):

qvm-tools/qubes-hcl-report
```diff
@@ -17,10 +17,11 @@
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library; if not, see <https://www.gnu.org/licenses/>.
 
+set -uo pipefail
```
marmarek, line 20: Ok, now lets talk about `-u`, as the next hardening option that actually fired back:
```
$ qubes-hcl-report --yaml
/usr/bin/qubes-hcl-report: line 80: $1: unbound variable
```
I think that's the only erroneous case, but you may want to re-check all variables just in case.

marmarek (changes requested):

marmarek (approved):

PR #532, opened 2023-05-02
Bring up network interface before using it

DemiMarie:
'udevadm settle' is not sufficient to guarantee that a network interface is up, as /usr/lib/qubes/setup-ip is run asynchronously by a systemd unit that runs after udev rule propagation.  Instead, start the systemd unit explicitly and wait for it to finish.

### DIFF
--- a/qubes/tests/integ/network.py
+++ b/qubes/tests/integ/network.py
@@ -430,7 +430,7 @@ class VmNetworkingMixin(object):
         self.testvm1.netvm = self.testnetvm
         # wait for it to settle down
         self.loop.run_until_complete(self.testvm1.run_for_stdio(
-            'udevadm settle'))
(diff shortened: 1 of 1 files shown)


marmarek:
NAK, the test is supposed to check also if something actually called the script, starting it manually would make the test pointless.

Besides, the observed failure collects logs and interfaces info with some delay, I don't think it's just timing issue.

DemiMarie:
> NAK, the test is supposed to check also if something actually called the script, starting it manually would make the test pointless.
> 
> Besides, the observed failure collects logs and interfaces info with some delay, I don't think it's just timing issue.

It _is_ a timing issue.  If I sleep 30 seconds it works.  If I sleep 2 seconds it does not.

PR #533, opened 2023-05-03
Work around race condition in network tests

DemiMarie:
`udevadm settle` is not sufficient to guarantee that the network scripts have run.  Instead, retry the ping command with a 10-second timeout.

### DIFF
--- a/qubes/tests/integ/network.py
+++ b/qubes/tests/integ/network.py
@@ -45,7 +45,7 @@ class VmNetworkingMixin(object):
     # filled by load_tests
     template = None
 
-    def run_cmd(self, vm, cmd, user="root"):
+    def run_cmd(self, vm, cmd, user="root", timeout=1 << 62):
         '''Run a command *cmd* in a *vm* as *user*. Return its exit code.
         :type self: qubes.tests.SystemTestCase | VmNetworkingMixin
         :param qubes.vm.qubesvm.QubesVM vm: VM object to run command in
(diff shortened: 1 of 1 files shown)


marmarek:
Detaching may need some better synchronization too, now I found this in the log:
```
[2023-05-05 03:28:18] [   41.447516] qrexec-agent[884]: 2023-05-05 03:28:18.259 qrexec-agent[884]: qrexec-agent-data.c:243:handle_new_process_common: executed: root:until ping -W 1 -n -c 1 192.168.123.45; do sleep 1; done (pid 903)
[2023-05-05 03:28:18] [   41.563949] systemd[1]: Stopping qubes-network-uplink@eth0.service - Qubes network uplink (eth0) setup...
[2023-05-05 03:28:18] [   41.659936] systemd[1]: Requested transaction contradicts existing jobs: Resource deadlock avoided
[2023-05-05 03:28:18] [   41.698929] systemd[1]: qubes-network-uplink@eth0.service: Deactivated successfully.
[2023-05-05 03:28:18] [   41.730458] systemd[1]: Stopped qubes-network-uplink@eth0.service - Qubes network uplink (eth0) setup.
```

So, it seems systemd processes network detach only after it was already re-attached.

DemiMarie:
I don’t think this can be properly handled without relying on `qubes-firewall.service` (or some other persistent daemon) unless we are willing to block the udev event loop (IMO a bad idea) or use `systemd-run` from the udev rule or something else equally hacky.

marmarek (approved):

PR #535, opened 2023-05-08
reflink: implement `usage_details()`, handle absent pool

rustybird:
Not sure if this would completely f*x QubesOS/qubes-issues#8187 and QubesOS/qubes-issues#8188 because I don't have an R4.2 installation at hand to test, and maybe the other storage drivers need to be tweaked as well


rustybird:
Pylint noise will be fixed by #534

marmarek:
PipelineRetry

marmarek (approved):

PR #543, opened 2023-06-20
Work around Fedora 38 LVM command line parsing bug

DemiMarie:
Fedora 38's LVM does not accept --snapshot as an argument to vgcreate, but it is possible to have the same effect without --snapshot.

I’m marking this as draft because Qubes OS bundles its own patched version of LVM (which does not have this problem) and I’m not sure I want to make the code harder to understand.

### DIFF
--- a/qubes/storage/lvm.py
+++ b/qubes/storage/lvm.py
@@ -769,7 +769,7 @@ def _get_lvm_cmdline(cmd):
     elif action == 'clone':
         assert len(cmd) == 3, 'wrong number of arguments for clone'
         lvm_cmd = ['lvcreate', '--setactivationskip=n', '--activate=y',
-                   '--snapshot', '--type=thin', '--name=' + cmd[2],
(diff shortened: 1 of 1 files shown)


marmarek:
> Work around Fedora 38 LVM command line parsing bug

Link to the bug report?

DemiMarie:
> > Work around Fedora 38 LVM command line parsing bug
> 
> Link to the bug report?

https://bugzilla.redhat.com/show_bug.cgi?id=2216286

rustybird:
>  I’m not sure I want to make the code harder to understand.

It could be reverted as soon as the [bugfix](https://sourceware.org/git/?p=lvm2.git;a=commit;h=ea6b5b694bf0886052b048b2e04b941ec65d90f9) has landed in the fc38 repo (presumably with `lvm2-2.03.22`). But in the meantime it would reduce friction for people setting up their development environment.

DemiMarie:
Fedora 39 is out and isn’t affected by this bug, so I’m closing this PR.

PR #554, opened 2023-08-18
Host support for high precision time

DemiMarie:
This adds support for nanosecond precision in the host's qubes.GetDate. Since Python only supports microseconds the date command is used instead.

Fixes: QubesOS/qubes-issues#7081

### DIFF
--- a/qubes-rpc/qubes.GetDate
+++ b/qubes-rpc/qubes.GetDate
@@ -20,10 +20,19 @@
 #
 
 import qubesadmin
-import datetime
+import os
 import subprocess
+import sys
 
-def main():
+def main(argv):
+    args = len(sys.argv)
+    if not (1 <= args <= 2):
+        print("wrong number of arguments, must have at most 1", file=sys.stderr)
+        sys.exit(1)
+    if args >= 2:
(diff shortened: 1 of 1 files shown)


marmarek:
> Since Python only supports microseconds the date command is used instead.

Honestly, with transfer latency (even on a local socket) nanoseconds precision is illusion anyway, microseconds precision would be more than enough.

qubes-rpc/qubes.GetDate
```diff
@@ -20,23 +20,36 @@
...
+    else:
+        arg = ""
     app = qubesadmin.Qubes()
 
     clockvm = app.clockvm
     if clockvm is None:
         return
 
     if clockvm.klass == 'AdminVM' or not clockvm.is_running():
+        env = {"LC_ALL": "C", **os.environ}
```
marmarek, line 43:
```suggestion
        env = {**os.environ, "LC_ALL": "C"}
```
Otherwise, LC_ALL in os.environ would override your value.

marmarek (changes requested):

marmarek (approved):

PR #558, opened 2023-10-05
Add Virtual DNS for IPv6

### LINKED ISSUE #10038: IPv6 DNS support

While Qubes does support IPv6 forwarding, DNS is still IPv4-only. When connecting to IPv6-only network, DNS doesn't work.

### The solution you'd like

Add IPv6 DNS address(es) and forward it to real IPv6 DNS (if present). Ensure that IPv4-only DNS still work without extra delay.

Better support for IPv6+IPv4 networks, and support at all of IPv6-only networks.

_No response_

1cho1ce:
Add Virtual IPv6 DNS for ipv6-enabled qubes.
The IPv6 address for DNS is using the same logic as IPv4:
10.137.x.x - normal qubes
10.138.x.x - disposable qubes
10.139.x.x - Virtual DNS
fd09:24ef:4179::a89:x - normal qubes
fd09:24ef:4179::a8a:x - disposable qubes
fd09:24ef:4179::a8b:x - Virtual DNS

Fixes https://github.com/QubesOS/qubes-issues/issues/10038

### DIFF
--- a/qubes/vm/mix/net.py
+++ b/qubes/vm/mix/net.py
@@ -57,6 +57,8 @@ def _default_ip6(self):
         return None
     if not self.features.check_with_netvm('ipv6', False):
         return None
+    if not self.features.check_with_template("supported-feature.ipv6", False):
+        return None
     if self.netvm is not None:
         return self.netvm.get_ip6_for_vm(self)  # pylint: disable=no-member
 
@@ -172,7 +174,9 @@ class NetVMMixin(qubes.events.Emitter):
     @qubes.stateless_property
     def visible_gateway6(self):
         '''Default (IPv6) gateway of this domain as seen by the domain.'''
-        if self.features.check_with_netvm('ipv6', False):
+        if not self.features.check_with_netvm('ipv6', False):
+            return None
+        if self.features.check_with_template("supported-feature.ipv6", False):
             return self.netvm.gateway6 if self.netvm else None
         return None
 
@@ -220,7 +224,9 @@ class NetVMMixin(qubes.events.Emitter):
     @qubes.stateless_property
     def gateway6(self):
         '''Gateway (IPv6) for other domains that use this domain as netvm.'''
-        if self.features.check_with_netvm('ipv6', False):
+        if not self.features.check_with_netvm('ipv6', False):
+            return None
+        if self.features.check_with_template("supported-feature.ipv6", False):
             return self.visible_ip6 if self.provides_network else \
                 None
         return None
@@ -246,17 +252,24 @@ class NetVMMixin(qubes.events.Emitter):
     @qubes.stateless_property
     def dns(self):
         '''DNS servers set up for this domain.'''
-        if self.netvm is not None:
-            if self.netvm.features.check_with_netvm('ipv6', False):
-                return StrSerializableTuple((
-                    'fd09:24ef:4179::a8b:1',
-                    '10.139.1.1',
-                ))
-            else:
-                return StrSerializableTuple((
-                    '10.139.1.1',
-                    '10.139.1.2',
-                ))
+        if self.netvm is not None or self.provides_network:
+            return StrSerializableTuple((
+                '10.139.1.1',
+                '10.139.1.2',
+            ))
+
+        return None
+
+    @qubes.stateless_property
+    def dns6(self):
+        '''IPv6 DNS servers set up for this domain.'''
+        if self.features.check_with_netvm('ipv6', False) and \
+           self.features.check_with_template("supported-feature.ipv6", \
+           False) and (self.netvm is not None or self.provides_network):
+            return StrSerializableTuple((
+                'fd09:24ef:4179::a8b:1',
+                'fd09:24ef:4179::a8b:2',
+            ))
 
         return None
 
@@ -512,7 +525,7 @@ class NetVMMixin(qubes.events.Emitter):
             vm.fire_event(
                 'property-reset:visible_gateway', name='visible_gateway')
 
-    @qubes.events.handler('property-set:ip6', 'property-reset:ipv6')
+    @qubes.events.handler('property-set:ip6', 'property-reset:ip6')
     def on_property_set_ip6(self, _event, name, newvalue=None, oldvalue=None):
         # pylint: disable=unused-argument
         if newvalue == oldvalue:
--- a/qubes/vm/mix/net.py
+++ b/qubes/vm/mix/net.py
@@ -247,10 +247,16 @@ class NetVMMixin(qubes.events.Emitter):
     def dns(self):
         '''DNS servers set up for this domain.'''
         if self.netvm is not None or self.provides_network:
-            return StrSerializableTuple((
-                '10.139.1.1',
-                '10.139.1.2',
-            ))
+            if self.features.check_with_netvm('ipv6', False):
+                return StrSerializableTuple((
+                    'fd09:24ef:4179::a8b:1',
+                    '10.139.1.1',
+                ))
+            else:
+                return StrSerializableTuple((
+                    '10.139.1.1',
+                    '10.139.1.2',
+                ))
 
         return None
 
--- a/qubes/vm/mix/net.py
+++ b/qubes/vm/mix/net.py
@@ -246,8 +246,8 @@ class NetVMMixin(qubes.events.Emitter):
     @qubes.stateless_property
     def dns(self):
         '''DNS servers set up for this domain.'''
-        if self.netvm is not None or self.provides_network:
-            if self.features.check_with_netvm('ipv6', False):
+        if self.netvm is not None:
+            if self.netvm.features.check_with_netvm('ipv6', False):
                 return StrSerializableTuple((
                     'fd09:24ef:4179::a8b:1',
                     '10.139.1.1',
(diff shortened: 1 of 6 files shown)


qubes/vm/qubesvm.py
```diff
@@ -2251,8 +2251,9 @@ def create_qdb_entries(self):
                                          str(self.gateway6))
             self.untrusted_qdb.write('/qubes-netvm-netmask', str(self.netmask))
 
-            for i, addr in zip(('primary', 'secondary'), self.dns):
-                self.untrusted_qdb.write('/qubes-netvm-{}-dns'.format(i), addr)
+            if self.netvm is not None:
```
marmarek, line 2254: This change made you hardcode virtual DNS addresses in the other PR. If you drop this condition, you will always have them available.

qubes/vm/mix/net.py
```diff
@@ -246,11 +246,17 @@ def connected_vms(self):
     @qubes.stateless_property
     def dns(self):
         '''DNS servers set up for this domain.'''
-        if self.netvm is not None or self.provides_network:
-            return StrSerializableTuple((
-                '10.139.1.1',
-                '10.139.1.2',
-            ))
+        if self.netvm is not None:
```
marmarek, line 249: It was `self.netvm is not None or self.provides_network` specifically to have virtual DNS addresses in sys-net too. Since you use `self.netvm`.features in a line below, you should check for `self.features` in case of no netvm. This asymmetric handling of sys-net is a bit weird, I agree, but since the addresses are constant anyway, it works just fine.

marmarek (changes requested):

marmarek:
See also CI failures - both pylint and unit tests (the latter failed due to changed condition for setting dns property).
You may also want to add a test for this new case, should be quite easy based on an existing test + setting `ipv6` feature.

marmarek:
And finally, this need to be guarded with some condition "if qube knows how to handle IPv6 DNS". Otherwise setting IPv6 DNS address will confuse network-related startup scripts.
I think you can use [qvm-features-request](https://dev.qubes-os.org/projects/core-admin/en/latest/qubes-features.html#qvm-features-request-qubes-postinstall-service) framework. Basically announce `supported-feature.ipv6dns=1` in https://github.com/QubesOS/qubes-core-agent-linux/blob/main/qubes-rpc/post-install.d/10-qubes-core-agent-features.sh, and check for it (`features.check_with_template("supported-feature.ipv6dns", False)`) here, in addition to checking for ipv6 itself.

qubes/vm/mix/net.py
1cho1ce: Is there a way to check if the qube has netvm or not from inside it?
This change was mainly so I could add reject rules to sys-net based on whatever this qube's netvm DNS servers are set. Without it I couldn't find a way to check if qube is connected to a netvm to determine if it should pass or reject the "virtual" DNS queries.

1cho1ce: Also the reason as to why did I check `self.netvm.features.check_with_netvm` is because I wanted to pass the qube usable primary and secondary virtual DNS servers.
For example, with ipv6 feature enabled only in sys-vpn and all qubes connected to it, if I just check if sys-vpn has ipv6 feature and give it IPv6 primary and IPv4 secondary virtual DNS servers then it won't be able to use the primary IPv6 server because the sys-vpn's netvm doesn't support IPv6 so it's better to give sys-vpn two IPv4 virtual DNS from netvm that are supported by the qube's netvm.

1cho1ce: > Is there a way to check if the qube has netvm or not from inside it?

I guess I should just check if `/qubes-ip` is set.

1cho1ce:
> And finally, this need to be guarded with some condition "if qube knows how to handle IPv6 DNS". Otherwise setting IPv6 DNS address will confuse network-related startup scripts.
> I think you can use [qvm-features-request](https://dev.qubes-os.org/projects/core-admin/en/latest/qubes-features.html#qvm-features-request-qubes-postinstall-service) framework. Basically announce `supported-feature.ipv6dns=1` in https://github.com/QubesOS/qubes-core-agent-linux/blob/main/qubes-rpc/post-install.d/10-qubes-core-agent-features.sh, and check for it (`features.check_with_template("supported-feature.ipv6dns", False)`) here, in addition to checking for ipv6 itself.

I can't think of a proper universal way to just check if IPv6 DNS handling is supported. I think that just checking if the system supports IPv6 is enough. But in that case this IPv6 support check should be done for all IPv6 networking and not just for DNS.
Should I add `supported-feature.ipv6` instead?

qubes/vm/mix/net.py
marmarek: > For example, with ipv6 feature enabled only in sys-vpn and all qubes connected to it, if I just check if sys-vpn has ipv6 feature and give it IPv6 primary and IPv4 secondary virtual DNS servers then it won't be able to use the primary IPv6 server because the sys-vpn's netvm doesn't support IPv6 so it's better to give sys-vpn two IPv4 virtual DNS from netvm that are supported by the qube's netvm.

That very much depends what your VPN does. If it tunnels all the traffic (IPv4 + IPv6) inside a IPv4 tunnel, then IPv6 DNS should also work. And in that case, the VPN software will set DNS in sys-vpn (in /etc/resolv.conf, or systemd-resolved) to something else than virtual DNS, so all should be fine.

We have two sets of DNS addresses in the qubesdb:
1. `/qubes-primary-dns`, `/qubes-secondary-dns` - this is what qube itself should use (and when providing network to others - where should redirect DNS traffic _to_)
2. `/qubes-netvm-primary-dns`, `/qubes-netvm-secondary-dns` - this is what addresses should be intercepted and redirected to the actual DNS (which may be the virtual one in connected netvm)

The first set in a qube should match the second set in the qube's netvm. I wouldn't be surprised if it's mixed up right now (because we had just one possibility there).

So, lets suppose setup like this: sys-net (no ipv6) -> sys-vpn (ipv6) -> some-vm (ipv6 too, because connected to sys-vpn)
- sys-net should have both sets IPv4-only
- sys-vpn should have first set IPv4-only, but second IPv6+IPv4
- some-vm should have the first set IPv6+IPv4 (and no second set because it doesn't provide network to others)

In sys-vpn, if it doesn't tunnel IPv6 DNS into VPN (to some real DNS server), it should have a reject rule on IPv6 virtual DNS address, because sys-net doesn't support it. This can be done by having virtual IPv6 DNS reject rule in postrouting instead of prerouting - then if VPN redirects the IPv6 DNS to somewhere else, it will bypass the reject rule and will work; otherwise it will get rejected (and hopefully the ICMP response will be forwarded back to the client - requires testing).

> > Is there a way to check if the qube has netvm or not from inside it?
> 
> I guess I should just check if `/qubes-ip` is set.

Yes, that works.

marmarek:
> I can't think of a proper universal way to check if just IPv6 DNS handling is supported. I think that just checking if the system supports IPv6 is enough. But in that case this IPv6 support check should be done for all IPv6 networking and not just for DNS.

The point is that adding virtual IPv6 DNS without regression in case of only IPv4 real DNS requires the reject rules you add in the other repo. If appropriate reject rules are added, it shouldn't matter if real IPv6 DNS is present or not. And to know that, you just need to know if core-agent-linux is new enough in relevant qube - which can be done with the mechanism I proposed above (core-agent-linux announcing support for ipv6dns feature).

1cho1ce:
Oh, I got it now.
What is the proper way to check the  qubes-core-agent-linux version?
I can see that it gets the qubes-agents version from the `/usr/share/qubes/marker-vm` file here:
https://github.com/QubesOS/qubes-core-agent-linux/blob/main/qubes-rpc/post-install.d/10-qubes-core-agent-features.sh#L5
But it has only major version (4.2) and not minor one (like 4.2.24).
Or is it enough to just check for next major version 4.3 since this feature won't be included before Qubes OS 4.3?

marmarek:
> What is the proper way to check the qubes-core-agent-linux version?

The idea is not to check explicit version, but to check if the version installed there announced appropriate feature.
1. core-agent-linux that has ipv6dns support should announce `supported-feature.ipv6dns=1` (add it to https://github.com/QubesOS/qubes-core-agent-linux/blob/main/qubes-rpc/post-install.d/10-qubes-core-agent-features.sh)
2. code here checks if `supported-feature.ipv6dns` was announced by the qube using `features.check_with_template("supported-feature.ipv6dns", False)`

1cho1ce:
Got it, that was obvious and was I just dumbed out.

qubes/vm/mix/net.py
1cho1ce: > In sys-vpn, if it doesn't tunnel IPv6 DNS into VPN (to some real DNS server), it should have a reject rule on IPv6 virtual DNS address, because sys-net doesn't support it. This can be done by having virtual IPv6 DNS reject rule in postrouting instead of prerouting - then if VPN redirects the IPv6 DNS to somewhere else, it will bypass the reject rule and will work; otherwise it will get rejected (and hopefully the ICMP response will be forwarded back to the client - requires testing).

I've decided to add jump from `dnat-dns` to separate chain `custom-dnat-dns` for VPN rules instead of adding reject rules in `postrouting` or `forward` chains for the sake of uniformity in the same way as chain `input` has chain `custom-input` and chain `forward` has chain `custom-forward`.
VPN should add its DNS redirects to `custom-dnat-dns` chain.

I think the PR is ready for a review now.

marmarek:
I'll run this (and the other PR) through openQA. For now just to test if IPv4 DNS still works properly. IPv6 DNS support will need additional test, but I'll add it later.

PR #559, opened 2023-10-09
Partial fix for https://github.com/QubesOS/qubes-issues/issues/6587

marmarek:
This is forward ported version of https://github.com/QubesOS/qubes-core-admin/pull/407

Original description by @v6ak :

Partial fix for https://github.com/QubesOS/qubes-issues/issues/6587

- Persistently assigned block devices no longer reorder. (This does not matter much on its own, as their order in qubes.xml is mostly irelevant, but it is much easier to test, as I can increase the reproduction rate to nearly 100 % by assigning plenty of block devices.)
- PCI devices seem to reorder less frequently. I'd like to investigate this deeper, but this is all I know ATM.
- They are rather conservative – I haven't touched the original method assignments, but rather copied and created a new one. It doesn't seem to be hard to change other occurrences, but I am not sure about testing.
- Tests: works on my machine. I see, it deserves to be better, but I am not sure how to do it properly.
- I believe I have simplified the for cycle of DeviceCollection.assignments_list a bit. If the evolution is not clear, there is some intermediate step: https://gist.github.com/v6ak/23c6b5f84238e314a8a751c8ac6de48d

### DIFF
--- a/qubes/devices.py
+++ b/qubes/devices.py
@@ -341,8 +341,8 @@ class DeviceCollection:
         :file:`qubes.xml`) or not. Device can also be in :file:`qubes.xml`,
         but be temporarily detached.
 
-        :param Optional[bool] persistent: only include devices which are or are not
-        attached persistently.
+        :param Optional[bool] persistent: only include devices which are or are
+        not attached persistently.
         '''
 
         try:
--- a/qubes/devices.py
+++ b/qubes/devices.py
@@ -55,6 +55,8 @@ Extension may use QubesDB watch API (QubesVM.watch_qdb_path(path), then handle
 `domain-qdb-change:path`) to detect changes and fire
 `device-list-change:class` event.
 '''
+from typing import Optional
+
 import qubes.utils
 
 class DeviceNotAttached(qubes.exc.QubesException, KeyError):
@@ -339,7 +341,7 @@ class DeviceCollection:
         :file:`qubes.xml`) or not. Device can also be in :file:`qubes.xml`,
         but be temporarily detached.
 
-        :param bool persistent: only include devices which are or are not
+        :param Optional[bool] persistent: only include devices which are or are not
         attached persistently.
         '''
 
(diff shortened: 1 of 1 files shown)


qubes/devices.py
```diff
@@ -433,7 +431,7 @@ class PersistentCollection:
     '''
 
     def __init__(self):
-        self._dict = {}
+        self._dict = OrderedDict()
```
rustybird, line 434: [Regular dicts are already ordered these days:](https://docs.python.org/3/library/stdtypes.html#dict)  

"Changed in version 3.7: Dictionary order is guaranteed to be insertion order. This behavior was an implementation detail of CPython from 3.6."

v6ak: I think it is OK to exclude this change. I've found three reasons to use OrderedDict even in modern Python versions:

1. You need some extra method provided by OrderedDict that is missing in dict. It doesn't apply to this PR.
2. Equality (===). It doesn't apply to this PR.
3. Documentation. By using OrderedDict, you make it explicit that you care about the order. I neither strongly support nor strongly oppose this.

marmarek: I dropped OrderedDict then. And also squashed the two commits, so it's easier to follow the history (no more splitting+merging again the function)

qubes/devices.py
```diff
@@ -351,24 +353,19 @@ def assignments(self, persistent=None):
                 self._bus))
             if persistent is True:
                 # don't break app.save()
-                return self._set
+                return list(self._set)
```
DemiMarie, line 356: Could these be sorted?

qubes/devices.py
```diff
@@ -351,24 +353,19 @@ def assignments(self, persistent=None):
...
+            result.extend(self._set)
+        if not persistent:  # None or False
+            for dev, options in devices:
+                if dev not in self._set:
+                    result.append(
+                        DeviceAssignment(
+                            backend_domain=dev.backend_domain,
+                            ident=dev.ident, options=options,
+                            bus=self._bus))
         return result
```
DemiMarie, line 369: Should this be `sorted(result)`?

marmarek: No. Otherwise adding new device could move another one attached earlier.

qubes/devices.py
v6ak: I know the line looks a bit silly, because it looks like we are converting a set (i.e., randomly sorted collection) to a list. However, `self._set` is a PersistentCollection, which is backed by dict. So, IIUC, PersistentCollection keeps the order of the items.

Yeah, we could rename `self._set` to some more descriptive. But it is a set (it cannot contain any item twice), just not an ordinary Python set.

PR #560, opened 2023-10-10
Ignore PCI devices on non-0000 domain instead of crashing

marmarek:
This is a quick band-aid to unbreak Qubes on systems having such device. 
Later change should implement handling them properly instead of ignoring.

Fixes QubesOS/qubes-issues#6932


docelic:
Related/duplicate bug report: https://github.com/QubesOS/qubes-issues/issues/8143

PR #561, opened 2023-10-17
Add a timeout to qubes-core.service

DemiMarie:
This ensures that if there is a bug and the service hangs, the user is still able to use the system eventually, rather than having to edit the kernel command line to recover.

### DIFF
--- a/linux/systemd/qubes-core.service
+++ b/linux/systemd/qubes-core.service
@@ -9,6 +9,8 @@ StandardOutput=syslog
 RemainAfterExit=yes
 # Needed to avoid rebooting before all VMs have shut down.
 TimeoutStopSec=180
+# To prevent the system from hanging during startup if there is a bug
+TimeoutStartSec=180
 ExecStart=/usr/lib/qubes/startup-misc.sh
 ExecStop=/usr/bin/qvm-shutdown -q --all --wait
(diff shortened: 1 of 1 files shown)


marmarek:
I don't like it, some of the operations may take a long time - especially on a slower system after unclean shutdown, and a static timeout won't be appropriate here. Interrupting it would be wrong. And if I'm right about diagnosis of https://github.com/QubesOS/qubes-issues/issues/8619, it wouldn't help at all.

DemiMarie:
> I don't like it, some of the operations may take a long time - especially on a slower system after unclean shutdown, and a static timeout won't be appropriate here. Interrupting it would be wrong.

The problem is that this blocks system boot — the user cannot log in until the service is finished.  This is really bad for user experience.

Do these operations *need* to be done during bootup, or can they be done in the background once the system is running?  The latter don’t require a timeout, but the former really should have one.

> And if I'm right about diagnosis of QubesOS/qubes-issues#8619, it wouldn't help at all.

You weren’t.

marmarek:
Cleaning up dispVMs (from unclean shutdown) should be done before user has a chance to interact with the system, to prevent them from interacting with to-be-removed dispvm. And as said before, static timeout won't do here. I'd rather figure out what hanged in your case (which seems to be rather unique report) than risking breaking unclean shutdown recovery for many users.

PR #567, opened 2023-12-05
Ensure networking packages are not installed

DemiMarie:
systemd-resolved, systemd-networkd, and NetworkManager are all worse than useless in dom0 and should not be present there.  Ensure that they do not get installed in dom0, and that if a user's system does have them, they are removed.

### DIFF
--- a/rpm_spec/core-dom0.spec.in
+++ b/rpm_spec/core-dom0.spec.in
@@ -115,6 +115,14 @@ Provides:	qubes-core-dom0-doc
 Obsoletes:	preupgrade < 2.0
 Provides:	preupgrade = 2.0
 
+# Ensure that systemd-resolved, systemd-networkd, and NetworkManager are not installed
+Conflicts:      systemd-networkd
+Obsoletes:      systemd-networkd
+Conflicts:      systemd-resolved
+Obsoletes:      systemd-resolved
+Conflicts:      NetworkManager
+Obsoletes:      NetworkManager
+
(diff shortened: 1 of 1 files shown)


marmarek:
While I agree they are not useful, this is definitely _not_ the way to remove them, it almost surely will break something. If nothing else, it will make packages depending on specific version (or `>=` or such) uninstallable (or if they are already installed, then update of qubes-core-dom0 will be uninstallable). But also, something may depend on their dbus API, or even just files shipped by those packages, and will break without them.

The proper way to avoid them is to check what makes them installed in the first place and adjust it there - making sure thing will work with that dependency (or maybe adjust some build flags if relevant?).

DemiMarie:
NetworkManager gets pulled in by Anaconda, but systemd-networkd and systemd-resolved are not IIUC.  systemd-resolved just provides DNS resolution, which doesn’t work anyway and falls back to `/etc/hosts`, so that should not be a problem.

marmarek:
I understand their functionality shouldn't be necessary in dom0. But it doesn't mean you can safely remove them by effectively breaking dependencies. Something has pulled the packages in and expects them to be present. Something may for example expect presence of `/usr/sbin/resolvconf`, or its dbus service. Before removing, every package depending on them needs to be inspected why it depends on it, and modified to not depend anymore (or at least document why is it safe to remove). AFAIR Anaconda actually does query NetworkManager dbus API and will fail to start if it gets timeout, so the change you did is hard no-go.

PR #572, opened 2023-12-27
Correct misleading check_with_* docstrings

marmarta:
fixes QubesOS/qubes-issues#8584


marmarta:
I'd like to note that this is 100% docstring change and codecov has 0 legitimate reasons to complain.

marmarek (approved):

PR #574, opened 2024-01-14
Support using UUIDs instead of VM names

DemiMarie:
This supports using UUIDs (instead of VM names) in the Admin API.  It also provides the UUIDs to qrexec-policy-daemon, which will use them to support the `@uuid:` keyword.

All admin APIs now support UUIDs as source and destination.  Specific APIs changed include:

- `admin.vm.CreateDisposable`: if the "uuid" argument is used, the VM UUID is returned instead of the name.
- `admin.vm.List`: the UUID is included in the returned list.
- `internal.GetSystemInfo`: the UUID is returned in the `uuid` key of each VM's JSON object.

Fixes: QubesOS/qubes-issues#8862

### DIFF
--- a/qubes/api/__init__.py
+++ b/qubes/api/__init__.py
@@ -24,10 +24,13 @@ import errno
 import functools
 import io
 import os
+import re
 import shutil
 import socket
 import struct
 import traceback
+from typing import Union, Any
+import uuid
 
 import qubes.exc
 from qubes.exc import ProtocolError, PermissionDenied
@@ -88,6 +91,32 @@ def apply_filters(iterable, filters):
         iterable = filter(selector, iterable)
     return iterable
 
+# This regex allows incorrect-length UUIDs,
+# but there is an explicit length check to catch that.
+_uuid_regex = re.compile(rb"\Auuid:[0-9a-f]{8}(?:-[0-9a-f]{4}){3}-[0-9a-f]*")
+def decode_vm(
+    untrusted_input: bytes, domains: qubes.app.VMCollection
+) -> qubes.vm.qubesvm.QubesVM:
+    lookup: Union[uuid.UUID, str]
+    vm = untrusted_input.decode("ascii", "strict")
+    if untrusted_input.startswith(b"uuid:"):
+        if (len(untrusted_input) != 41 or
+            not _uuid_regex.match(untrusted_input)):
+            raise qubes.exc.QubesVMInvalidUUIDError(vm[5:])
+        lookup = uuid.UUID(vm[5:])
+    else:
+        # throws if name is invalid
+        qubes.vm.validate_name(None, None, vm)
+        lookup = vm
+    try:
+        return domains[lookup]
+    except KeyError:
+        # normally this should filtered out by qrexec policy, but there are
+        # two cases it might not be:
+        # 1. The call comes from dom0, which bypasses qrexec policy
+        # 2. Domain was removed between checking the policy and here
+        # we inform the client accordingly
+        raise qubes.exc.QubesVMNotFoundError(vm)
 
 class AbstractQubesAPI:
     '''Common code for Qubes Management Protocol handling
@@ -108,25 +137,23 @@ class AbstractQubesAPI:
     #: the preferred socket location (to be overridden in child's class)
     SOCKNAME = None
 
-    def __init__(self, app, src, method_name, dest, arg, send_event=None):
+    app: qubes.Qubes
+    src: qubes.vm.qubesvm.QubesVM
+    def __init__(self,
+                 app: qubes.Qubes,
+                 src: bytes,
+                 method_name: bytes,
+                 dest: bytes,
+                 arg: qubes.Qubes,
+                 send_event: Any = None) -> None:
         #: :py:class:`qubes.Qubes` object
         self.app = app
 
-        try:
-            vm = src.decode('ascii')
-            #: source qube
-            self.src = self.app.domains[vm]
+        #: source qube
+        self.src = decode_vm(src, app.domains)
 
-            vm = dest.decode('ascii')
-            #: destination qube
-            self.dest = self.app.domains[vm]
-        except KeyError:
-            # normally this should be filtered out by qrexec policy, but there
-            # are two cases it might not be:
-            # 1. The call comes from dom0, which bypasses qrexec policy
-            # 2. Domain was removed between checking the policy and here
-            # we inform the client accordingly
-            raise qubes.exc.QubesVMNotFoundError(vm)
+        #: destination qube
+        self.dest = decode_vm(dest, app.domains)
 
         #: argument
         self.arg = arg.decode('ascii')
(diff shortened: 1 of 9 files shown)


qubes/api/__init__.py
```diff
@@ -94,6 +97,33 @@ def apply_filters(iterable, filters):
         iterable = filter(selector, iterable)
     return iterable
 
+# pylint: disable=line-too-long
+_uuid_regex = re.compile(rb"\A@uuid:[0-9a-f]{8}(?:-[0-9a-f]{4}){3}-[0-9a-f]{12}\Z")
+_qube_name_regex = re.compile(rb"\A[A-Za-z][A-Za-z0-9_.-]*\Z")
```
marmarek, line 102: Do not duplicate name validation code, there is `qubes.vm.validate_name` already  (you can pass None for holder and prop args).

qubes/api/__init__.py
```diff
@@ -94,6 +97,33 @@ def apply_filters(iterable, filters):
...
+def decode_vm(
+    untrusted_input: bytes, domains: qubes.app.VMCollection
+) -> qubes.vm.qubesvm.QubesVM:
+    lookup: Union[uuid.UUID, str]
+    vm = untrusted_input.decode("ascii", "strict")
+    if untrusted_input.startswith(b"@uuid:"):
+        if (len(untrusted_input) != 42 or
```
marmarek, line 109: Length check is redundant here, regex already covers that.

qubes/api/admin.py
```diff
@@ -1134,10 +1135,14 @@ async def _vm_create(self, vm_type, allow_pool=False,
             raise
         self.app.save()
 
-    @qubes.api.method('admin.vm.CreateDisposable', no_payload=True,
+    @qubes.api.method('admin.vm.CreateDisposable',
         scope='global', write=True)
-    async def create_disposable(self):
+    async def create_disposable(self, untrusted_payload):
         self.enforce(not self.arg)
+        if untrusted_payload not in (b"", b"uuid"):
```
marmarek, line 1142: In the description you said it should be given in the argument, not the payload. So, which one is it?

qubes/tests/api_internal.py
```diff
@@ -160,6 +162,7 @@ def test_010_get_system_info(self):
         vm.label.icon = 'icon-vm'
         vm.guivm = vm
         vm.get_power_state.return_value = 'Halted'
+        vm.uuid = uuid.uuid4()
```
marmarek, line 165: I'd prefer specific hardcoded value in the test (simply generate new random one) instead of ever changing random here.

qubes/tests/api_internal.py
```diff
@@ -182,6 +186,7 @@ def test_010_get_system_info(self):
                     'icon': 'icon-vm',
                     'guivm': 'vm',
                     'power_state': 'Halted',
+                    "uuid": str(vm.uuid)
```
marmarek, line 189: With the the static value, here put exact expected string.

marmarek (changes requested):

marmarek (approved):
looks okay now, but I'll merge it only together with other related PRs

DemiMarie:
> looks okay now, but I'll merge it only together with other related PRs

That turned out to be a good idea, because the API isn’t sufficient for what qrexec needs.  qrexec needs not only the UUID of the created VM, but also its Xen domain ID.  This is needed to connect to the VM’s qrexec daemon in a race-free way.  Marked as draft.

marmarek:
> This is needed to connect to the VM’s qrexec daemon in a race-free way. Marked as draft.

The domain cannot change its name while running, this is non-issue. The race you consider would require a domain to be stopped, removed, re-created and started, all between qrexec-policy-daemon getting info from qubesd and calling qrexec-client (or otherwise connecting to it) which is very much unrealistic to happen. In fact, I'd strongly prefer to _not_ return Xen ID in the system info structure, exactly because it changes on qube restart, which will make any kind of caching (if happen at some point) less effective.

What can be done at some point (not R4.2, as will likely have compatibility issues) is to create a uuid->xid symlink to qrexec socket (in addition to name->xid) and use that for connections.

DemiMarie:
> > This is needed to connect to the VM’s qrexec daemon in a race-free way. Marked as draft.
> 
> The domain cannot change its name while running, this is non-issue. The race you consider would require a domain to be stopped, removed, re-created and started, all between qrexec-policy-daemon getting info from qubesd and calling qrexec-client (or otherwise connecting to it) which is very much unrealistic to happen. In fact, I'd strongly prefer to _not_ return Xen ID in the system info structure, exactly because it changes on qube restart, which will make any kind of caching (if happen at some point) less effective.
> 
> What can be done at some point (not R4.2, as will likely have compatibility issues) is to create a uuid->xid symlink to qrexec socket (in addition to name->xid) and use that for connections.

There should not be any compatibility issues.  All that is needed is for qubesd to either create the symlink itself or tell `qrexec-daemon` to do so.  That requires new command-line options, but package dependencies will guarantee that the new `qrexec-daemon` is installed before the new qubesd is.  Existing daemons will work fine, and the new daemon can fall back to the name if the UUID symlink is missing.  Yes, I do plan to test all of these scenarios.

marmarek:
Please don't. Qrexec is too critical part to risk it.

DemiMarie:
> Please don't. Qrexec is too critical part to risk it.

Ack.  This will still be useful for the builder, where the race window is much larger.

DemiMarie:
> Please don't. Qrexec is too critical part to risk it.

Simpler solution:

- Tell qrexec-daemon to create a symbolic link using the UUID.  This information can be passed via an environment variable, which the current daemon ignores.
- If connecting via the UUID fails, fall back to a name-based connection.
- If the environment variable is set, qrexec-daemon uses the new `admin.vm.CreateDisposable` API, which gives UUID, name, and XID.  The XID is used to make the connection.

marmarek:
I still don't like changes like this in a stable release. To be clear: backporting feature or an API change to a stable release should be considered exception not a rule. Based on risk-benefit analysis I do not grant exception for this case.
Using UUID-based symlink in R4.3 is fine. And at this point, I'd prefer it to be done explicitly (extra command line option with UUID?), not via a sneaky env variable. I'll add relevant package dependencies to ensure both sides are updated (on dist-upgrade).

marmarek:
This got some conflicts too, and is needed for https://github.com/QubesOS/qubes-core-qrexec/pull/135

marmarek:
Test fails. And also pylint complains about too long lines (others look as false positives)

qubes/vm/qubesvm.py
```diff
@@ -801,6 +802,17 @@ def xid(self):
                 e.get_error_code()))
             raise
 
+    @qubes.stateless_property
+    def stubdom_uuid(self):
+        stubdom_xid = self.stubdom_xid
```
marmarek, line 807: pylint needs some "help": `# pylint: disable=comparison-with-callable`

marmarek: Actually, I have a better idea: https://github.com/QubesOS/qubes-core-admin/pull/602/commits/3362ffc7ed4ea33c479f1df24fbdb25417a3584a
So, don't disable the message (it will be okay once I merge the other PR).

marmarek:
PipelineRetry

marmarek:
one test (different this time) still fails

PR #579, opened 2024-02-06
new device API

piotrbartman:
need QubesOS/qubes-linux-utils/pull/111

draft of implementation of new device API.
split persistent flag into automatically_attach and required, so usb devices can be auto-attached but in case of lack of device or renumbering usb bus we can still start a domain. 
TODO: move Device DeviceInterface, DeviceCategorym DeviceInfo, DeviceAssignment etc. to one place (it is shared between this package and client)
TODO: tests in this repo can be broken

 ## To Do and to Considerate
Currently, we can make one assignment of a selected port to a single machine in two ways:

1. Using `identity='any'`, which means that anything plugged into the designated port will be automatically attached to the VM, as it has been functioning thus far.
2. Using the identity of the currently plugged device, in which case automatic attachment will only occur if the device presents itself in the same way. This introduces a level of risk, but I believe it is acceptable because if a device can effectively impersonate another device, we cannot protect the user from it anyway.

An open issue is what to do with a device that attempts to forcibly present itself in various ways. At the moment, I consider such an attack to be unlikely. 

Another consideration is the user interface. It would be beneficial to provide options such as:

- The ability to assign at least two different devices, so that if either of them is plugged into a specific port, automatic attachment occurs.
- Automatically attach any device that presents itself as for example a mass storage and is plugged into a specific port, but not devices that resemble mass storage but are actually keyboards, for example.

The handling of USB bus number changes is still not implemented. According to @marmarek's suggestion, we could check the parent device (PCI) in this regard. However, in the current configuration, this would require significant refactoring, as the current protocol relies on device identification as `backend-vm:bus-port`. Thus, we need to find a method to ensure that this number is always correct. This will also address issues with block devices because currently, if we plug in USB sticks in different orders or do not properly eject and reinsert one of them, the `ident` of mass storage devices changes, and automatic attachment fails. Theoretically, using `self_identity` resolves this issue, but there are still assumptions in many places that `backend-vm:block-name` identifies the device, which is not straightforward.

https://github.com/QubesOS/qubes-issues/issues/4626


marmarek:
Please document the intended behavior and usage of this API. More or less update module docstring in `qubes/devices.py`. Besides updating what's currently there, I'm interested especially in two more things:
 - differences between "assign" and "attach" and how they interact with starting a qube and connecting a device
 - device identification: which properties are used to identify the device when

As for more specific comments:
 - you use vendor/product as strings from hwdata - are they stored anywhere (or otherwise expected to not change)? in rare cases they may change, for example when somebody fixes a typo in hwdata, or add info about the device that wasn't included there before
 - I see new `device-added` event, but no `device-removed` (or similar), at least in the documentation (but block extension actually sends `device-removed` event)
 - why only PCI devices can be "required"?

marmarek (changes requested):
Please document the intended behavior and usage of this API. More or less update module docstring in `qubes/devices.py`. Besides updating what's currently there, I'm interested especially in two more things:
 - differences between "assign" and "attach" and how they interact with starting a qube and connecting a device
 - device identification: which properties are used to identify the device when

As for more specific comments:
 - you use vendor/product as strings from hwdata - are they stored anywhere (or otherwise expected to not change)? in rare cases they may change, for example when somebody fixes a typo in hwdata, or add info about the device that wasn't included there before
 - I see new `device-added` event, but no `device-removed` (or similar), at least in the documentation (but block extension actually sends `device-removed` event)
 - why only PCI devices can be "required"?

piotrbartman:
Ok, I'm writing the docs

> * you use vendor/product as strings from hwdata - are they stored anywhere (or otherwise expected to not change)? in rare cases they may change, for example when somebody fixes a typo in hwdata, or add info about the device that wasn't included there before

The aim was that the device identification is "human-readable" and even more "non-tech-user-readable", meaning in `qubes.xml` we have a name under which an user identifies the device. But we can use 'raw' data transmitted by the device here, which will bypass the `hwdata` file and ensure that once created configuration will not be broken by update.

> * I see new `device-added` event, but no `device-removed` (or similar), at least in the documentation (but block extension actually sends `device-removed` event)

So I need to add this to documentation (there are `device-removed:block` and `device-removed:block`).

> * why only PCI devices can be "required"?

I think we've decided that there is no benefit in making USB/block `required`. Moreover, there is still the issue that the USB bus number can change between restarting `sys-usb` so the device will not be auto-attached, and the question is what we want to do about that. I considered just ignoring that first number, but I'm not sure if it is safe.

DemiMarie:
> > * why only PCI devices can be "required"?
> 
> I think we've decided that there is no benefit in making USB/block `required`. Moreover, there is still the issue that the USB bus number can change between restarting `sys-usb` so the device will not be auto-attached, and the question is what we want to do about that. I considered just ignoring that first number, but I'm not sure if it is safe.

The USB bus number changing is itself a problem, IMO.

marmarek:
> Moreover, there is still the issue that the USB bus number can change between restarting `sys-usb`

I'm not sure if we can ensure static bus numbers, but PCI devices should not be reordered anymore. So, maybe you can replace bus number in the ID with PCI device providing it (see symlink target of `/sys/bus/usb/devices/usbN`, or related udev property)? There is one caveat - one PCI device can provide multiple buses (specifically: USB 3 controller will have two buses: one for USB 2.0 and older and one for 3.0 and newer). So, some smart renumbering would be needed anyway.

As for "required" - for USB indeed it's a tricky problem, but I'd prefer this limitation to be enforced at the frontend side, or in extensions for specific device classes, not in the generic code in `devices.py`. And also, even if we disallow it for USB, IMO it still makes sense for (at least some) block devices.

> But we can use 'raw' data transmitted by the device here, which will bypass the `hwdata` file and ensure that once created configuration will not be broken by update.

It's good to show human readable name (translated via hwdata) to humans, but internally IMO it should use numeric vendor/product.

piotrbartman:
> > > * why only PCI devices can be "required"?
> > 
> > I think we've decided that there is no benefit in making USB/block `required`.
> 
> What if I want a qube to which my USB Bluetooth controller is permanently attached?

You can have USB Bluetooth controller *auto-attached* to a qube.

qubes/ext/block.py
```diff
@@ -105,22 +113,168 @@ def size(self):
...
+            p_info = p.backend_domain.devices[p.devclass][p.ident]
+            parent_identity = p_info.self_identity
+            if p.devclass == 'usb':
+                parent_identity = f'{p.ident}:{parent_identity}'
+        if self._interface_num:
+            # device interface number
+            self_id = self._interface_num
+        else:
+            # partition number: 'sda1' -> '1'
+            self_id = self.ident[3:]
```
marmarek, line 171: That wont always work, there can be other names (`xvdi1`), or even different schemes (`nvme0n1p2`). Generally, to get partition number, take last number (not just last digit, there can be `sda12`). But there is still one corner case - whole device ending with a number, like `nvme0n1`, or `loop0` - in those cases the partition number is separated with the `p`, but you can't tell it's a partition just because it ends with a number... Does the partition device always has "whole block device" as a parent? If so, you can use that fact and only then take partition number. But if parent device is not a block device, then it isn't a partition, so getting just the final number is not appropriate.

qubes/ext/block.py
```diff
@@ -283,6 +458,23 @@ def on_device_pre_attached_block(self, vm, event, device, options):
             vm.app.env.get_template('libvirt/devices/block.xml').render(
                 device=device, vm=vm, options=options))
 
+    @qubes.ext.handler('domain-start')
+    async def on_domain_start(self, vm, _event, **_kwargs):
+        # pylint: disable=unused-argument
+        for assignment in vm.devices['block'].get_assigned_devices():
+            asyncio.ensure_future(self._attach_and_notify(
+                vm, assignment.device, assignment.options))
```
marmarek, line 466: Block devices attached at domain start can be specified in libvirt xml, in fact you already adjust that part.

marmarek (changes requested):
few comments on the recent push; I'm waiting with the full review for the documentation.

DemiMarie (commented):
Would this be a good opportunity to use diskseq to identify block devices?  diskseq + partition number is guaranteed to be unique until the backend is rebooted, whereas other identifiers are not guaranteed to be unique.  Right now this rule is violated by loop and device-mapper devices, but that’s a kernel bug.

qubes/ext/block.py
piotrbartman: Youre right, I fixed it

piotrbartman:
> Would this be a good opportunity to use diskseq to identify block devices? diskseq + partition number is guaranteed to be unique until the backend is rebooted, whereas other identifiers are not guaranteed to be unique. Right now this rule is violated by loop and device-mapper devices, but that’s a kernel bug.

Not really, here we extract device info basing on qubes-db data, so it's just string processing.

piotrbartman:
> few comments on the recent push; I'm waiting with the full review for the documentation.

Most parts of docstring in `qubes/devices.py` are the same since the main concept remains unchanged. To find more docs you can go to `qubes-devices.rst`. Let me know if anything else need to be documented.

qubes/ext/block.py
piotrbartman: Can you elaborate?

marmarek: Block devices that are assigned when VM is starting are attached already at https://github.com/QubesOS/qubes-core-admin/pull/579/files#diff-bd38d7fc85084f576ddb6d1d4702f72158e0e30d5e1deb15cf3d49c28c0eb82eR152-R156. So, it looks like you attach them again at `domain-start` event here.

marmarek:
> To find more docs you can go to `qubes-devices.rst`

Oh, I missed you already updated it.

qubes/ext/block.py
marmarek: This is still the case.

qubes/ext/pci.py
```diff
@@ -150,16 +168,86 @@ def __init__(self, backend_domain, ident, libvirt_name=None):
                 raise UnsupportedDevice(libvirt_name)
             ident = '{bus}_{device}.{function}'.format(**dev_match.groupdict())
 
-        super().__init__(backend_domain, ident, None)
+        super().__init__(
+            backend_domain=backend_domain, ident=ident, devclass="pci")
+
+        if hasattr(self, 'regex'):
```
marmarek, line 174: This `if` was necessary when it was in a generic class. Here, the `regex` attribute is always set.

qubes/ext/pci.py
```diff
@@ -150,16 +168,86 @@ def __init__(self, backend_domain, ident, libvirt_name=None):
...
+        if self._vendor is None:
+            result = self._load_desc()["vendor"]
+        else:
+            result = self._vendor
+        return result
+
+    @property
+    def product(self) -> str:
+        """
+        Device name from local database `/usr/share/hwdata/usb.ids`
```
marmarek, line 207:
```suggestion
        Device name from local database `/usr/share/hwdata/pci.ids`
```

qubes/ext/pci.py
```diff
@@ -150,16 +168,86 @@ def __init__(self, backend_domain, ident, libvirt_name=None):
...
 
         # lazy loading
         self._description = None
+        self._vendor_id = None
+        self._product_id = None
+
+    @property
+    def vendor(self) -> str:
+        """
+        Device vendor from local database `/usr/share/hwdata/usb.ids`
```
marmarek, line 192:
```suggestion
        Device vendor from local database `/usr/share/hwdata/pci.ids`
```

marmarek (changes requested):
Since the API now considers also `self_identity` when automatically attaching devices (which is good!), I think there need to be support for multiple assignments of the same port but with different device `self_identity`. For example, I may have 3 different devices I'd like to automatically connect to a qube X (when plugged into specific port). Ofc only one of them can be attached at the same time, but I think it should still be possible to have assignments there.
This is problematic for the API, because some places use backend+ident as unique identifier for the asssignment. I made a comment about it in one place, but the problem is in several more places.
Maybe in such cases there can be some suffix to the device ident, like yet another `+something` (has
h of `self_identity` if not `self_identity` itself? Or not only in such cases, but always? And then
for port assignments that would be `+any`.
There can be a corner case of multiple assignments of the same port, all with `required=True`. That of course will always fail to start, I think such configuration should still be allowed (for example when user adds new device just to remove the old one later). Maybe a startup message can include a
message when such configuration will be detected.

BTW, one day maybe we will have a bus that has unspoofable device identifiers (based on some cryptography?). When that happens, maybe it will make sense to create assignment that has only backend and `self_identity` (or some other property then?) but has `any` in place of the current `ident`. Or maybe, for such trusted identities use it instead of the device port (not sure if a good idea...)? In any case, it would allow automatically attaching specific device regardless of which port it's plugged in. Such support may require some more changes (for example (potential) assignment would not have a specific port value, but it will need to be resolved when actually attaching the device), so lets leave implementation for another iteration. But it's good to think about such possibility when designing API interface.

piotrbartman:
> Since the API now considers also `self_identity` when automatically attaching devices (which is good!), I think there need to be support for multiple assignments of the same port but with different device `self_identity`. (...)

Agree, but as I mentioned in the last section 'To Do and to Consider' in the PR description, I believe this is a change for the next iteration, as it requires modifications in several places, and I think currently there are already quite a few changes done.

qubes/ext/pci.py
piotrbartman: done

qubes/ext/block.py
```diff
@@ -189,10 +190,41 @@ def parent_device(self) -> Optional[qubes.devices.Device]:
...
+        """
+        if not self.backend_domain.is_running():
+            return None
+        for vm in self.backend_domain.app.domains:
+            if not vm.is_running():
+                continue
+            if self._is_attached_to(vm):
+                return vm
+
+        return None
```
DemiMarie, line 206: Should there be an index to make looking this up fast?

piotrbartman: Probably, but we need to remember that it should always be up to date (not cached), if you have an idea how to do it, I'm in.

DemiMarie: Updating it whenever a device is attached or detached is the first idea that comes to mind.  Caches are fine so long as they are invalidated synchronously when needed.

piotrbartman: `BlockDeviceExtension` already has a `devices_cache`, but `attachment` is intended to return the actual state (including, in particular, situations where something went wrong and the potential cache is outdated). For API compatibility, this is a `property`, and the warning is there because when we want to get information about `attachment` for a bunch of devices, we should use some form of bulk data loading (just like we do for updating `devices_cache`) rather than in a loop.

DemiMarie: Are there any situations where the cache could be outdated, or is that always a bug?

piotrbartman: Not sure, but if it is predictable it's already update a cache

qubes/ext/pci.py
piotrbartman: ok

qubes/devices.py
```diff
@@ -106,58 +113,811 @@ def __eq__(self, other):
...
+        value = rest[:-1]  # ending '
+        values.append(deserialize_str(value))
+
+        properties = dict()
+        for key, value in zip(keys, values):
+            if key.startswith("_"):
+                # it's handled in cls.__init__
+                properties[key[1:]] = value
+            else:
+                properties[key] = value
```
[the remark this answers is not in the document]
piotrbartman: done

piotrbartman:
@marmarek Can you elaborate? 
https://github.com/QubesOS/qubes-core-admin/pull/579#discussion_r1525443653

qubes/devices.py
```diff
@@ -54,50 +57,54 @@
...
-from typing import Optional
+"""
+import itertools
+import string
+import sys
+from enum import Enum
+from typing import Optional, List, Type, Dict, Any, Iterable, Union
 
 import qubes.utils
+from qubes.api import ProtocolError
```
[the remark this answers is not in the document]
piotrbartman: done

qubes/api/admin.py
```diff
@@ -1316,41 +1358,46 @@ async def vm_device_detach(self, endpoint):
...
             for ep in pkg_resources.iter_entry_points('qubes.devices')),
-        scope='local', write=True, execute=True)
-    async def vm_device_set_persistent(self, endpoint, untrusted_payload):
+        scope='local', write=True)
+    async def vm_device_set_assignment(self, endpoint, untrusted_payload):
+        """
+        Update assignment of already attached device.
+
+        Payload:
+            `None` -> unassign device from qube
```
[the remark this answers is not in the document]
piotrbartman: yes

qubes/api/admin.py
```diff
@@ -1194,118 +1194,160 @@ async def deviceclass_list(self):
...
-                    (('persistent', 'yes' if dev.persistent else 'no'),)
-                ))
-            self.enforce('\n' not in properties_txt)
-            ident = '{!s}+{!s}'.format(dev.backend_domain, dev.ident)
-            dev_info[ident] = properties_txt
+        device_assignments = self.fire_event_for_filter(
+            device_assignments, devclass=devclass)
+
+        dev_info = {
+            f'{assignment.backend_domain}+{assignment.ident}':
```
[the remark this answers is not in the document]
piotrbartman: next iteration?

qubes/api/admin.py
```diff
@@ -1316,41 +1358,46 @@ async def vm_device_detach(self, endpoint):
...
-    async def vm_device_set_persistent(self, endpoint, untrusted_payload):
+        scope='local', write=True)
+    async def vm_device_set_assignment(self, endpoint, untrusted_payload):
+        """
+        Update assignment of already attached device.
+
+        Payload:
+            `None` -> unassign device from qube
+            `False` -> device will be auto-attached to qube
+            `True` -> device is required to start qube
```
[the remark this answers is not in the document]
piotrbartman: Not only, if "required" is `None` it modify `attach_automatically` flag, so whole assignment. In fact the assignment has only 3 valid states, and setting `attach_automatically` and `required` separably can lead to non-valid state. In other words,  DeviceAssignment has one state, but exposed to the user/programmer as 2 flags. I even consider to put `DeviceAssignment.__required` and `DeviceAssignment.__automatically_attach` as one variable with 3 possible values.

qubes-rpc-policy/90-admin-default.policy.header
```diff
@@ -20,14 +20,20 @@
...
 
+!include-service admin.vm.device.mic.Assign * include/admin-local-rwx
+!include-service admin.vm.device.mic.Assigned * include/admin-local-ro
 !include-service admin.vm.device.mic.Attach * include/admin-local-rwx
+!include-service admin.vm.device.mic.Attached * include/admin-local-ro
 !include-service admin.vm.device.mic.Available * include/admin-local-ro
 !include-service admin.vm.device.mic.Detach * include/admin-local-rwx
-!include-service admin.vm.device.mic.List * include/admin-local-ro
-!include-service admin.vm.device.mic.Set.persistent * include/admin-local-rwx
+!include-service admin.vm.device.mic.Set.assignment * include/admin-local-rwx
```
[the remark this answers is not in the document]
piotrbartman: see comment below

qubes/devices.py
```diff
@@ -106,58 +113,811 @@ def __eq__(self, other):
...
+    @property
+    def attachment(self) -> Optional['qubes.vm.BaseVM']:
+        """
+        VM to which device is attached (frontend domain).
+        """
+        return self._attachment
+
+    def serialize(self) -> bytes:
+        """
+        Serialize object to be transmitted via Qubes API.
```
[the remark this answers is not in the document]
piotrbartman: https://github.com/QubesOS/qubes-doc/pull/1387

qubes/ext/block.py
marmarek: Block devices attached at domain startup should be specified in libvirt xml (linked above), instead of attached after domain startup. This is especially important for "required" devices, as one may need them during system startup (for example some standalone qube may have it listed in /etc/fstab and startup will fail without it).

qubes/api/admin.py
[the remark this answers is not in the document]
marmarek: > if "required" is `None` it modify `attach_automatically` flag

But this is "Unassign" method listed above, no?

qubes/api/admin.py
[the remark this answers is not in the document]
marmarek: ok

qubes/api/admin.py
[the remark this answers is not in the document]
piotrbartman: ok, now I get it, I will correct that

marmarek:
Integration tests need an update: https://openqa.qubes-os.org/tests/96815

And [unit tests detected](https://gitlab.com/QubesOS/qubes-core-admin/-/jobs/6588234160) I believe a real issue - meminfo-writer service (as well as enabling/disabling dynamic memory management) is toggled based on PCI devices presence - this part likely misbehave now.

qubes/device_protocol.py
```diff
@@ -0,0 +1,943 @@
...
+                    result[class_id + '****'] = class_name
+                    subclass_id = None
+
+        return result
+
+
+class DeviceInfo(Device):
+    """ Holds all information about a device """
+    ALLOWED_CHARS_KEY = Device.ALLOWED_CHARS_KEY + string.punctuation + ' '
+    ALLOWED_CHARS_PARAM = Device.ALLOWED_CHARS_PARAM + string.punctuation + ' '
```
marmarek, line 402: Here I'm confused, you add `string.punctuation` again. Maybe it wasn't meant to be included in nether sets in the `Device` class?
And you add space several times too.

marmarek (changes requested):

marmarek:
And also, this has conflicts now...

qubes/device_protocol.py
```diff
@@ -0,0 +1,948 @@
...
+        # 'backend_domain', 'attachment', 'interfaces', 'data', 'parent_device'
+        # are not string, so they need special treatment
+        default_attrs = {
+            'ident', 'devclass', 'vendor', 'product', 'manufacturer', 'name',
+            'serial', 'self_identity'}
+        properties = b''
+        for key, value in ((key, getattr(self, key)) for key in default_attrs):
+            properties = self.pack_property(properties, key, value)
+        # remove unnecessary space in the beginning
+        properties = properties[1:]
```
marmarek, line 602: This is a bit weird - `self.pack_property` gets already serialized `properties`, yet still adds the leading space. And in fact, it's a bit weird to have that parameter, since it isn't used for anything besides appending result to it.
Isn't it better to `self.pack_property` really pack a single property (without concatenating to earlier content) and then do `b' '.join()` on results? Which would be also faster (not sure if that really matters here).

marmarek (changes requested):
The serialization/deserialization code really needs some unit tests for tricky cases - values including quotes, equal sign, spaces etc. And since you allow spaces in keys (is that used anywhere?) that wants a test too.

marmarek:
And also, some of the tests fail on `import qubesdb`, which means not patched `vm.untrusted_qdb` (see how `TestQubesDB()` is used).

marmarek:
> The serialization/deserialization code really needs some unit tests for tricky cases - values including quotes, equal sign, spaces etc. And since you allow spaces in keys (is that used anywhere?) that wants a test too.

Oh, you added some already :)
But see the point about spaces in keys.

qubes/device_protocol.py
```diff
@@ -0,0 +1,944 @@
...
+    """
+    Sanitize given untrusted string.
+
+    If `replace_char` is not None, ignore `error_message` and replace invalid
+    characters with the string.
+    """
+    if replace_char is None:
+        not_allowed_chars = set(untrusted_value) - allowed_chars
+        if not_allowed_chars:
+            print(untrusted_value)
```
marmarek, line 745: leftover debug?

marmarek:
There is a conflict now after merging https://github.com/QubesOS/qubes-core-admin/pull/590 :(

qubes/ext/block.py
```diff
@@ -105,29 +152,208 @@ def size(self):
...
+                return None
+            untrusted_parent_info = self.backend_domain.untrusted_qdb.read(
+                f'/qubes-block-devices/{self.ident}/parent')
+            if untrusted_parent_info is None:
+                return None
+            # '4-4.1:1.0' -> parent_ident='4-4.1', interface_num='1.0'
+            # 'sda' -> parent_ident='sda', interface_num=''
+            parent_ident, sep, interface_num = self._sanitize(
+                untrusted_parent_info).partition(":")
+            devclass = 'usb' if sep == ':' else 'block'
```
marmarek, line 186: Looking at it now, since this is new code (so, no need to keep compatibility with some specific API), isn't it better for VM to explicitly say which device class parent is, instead of guessing based on ident format? I'm not sure what else it could be here, and USB vs block idents are quite different, but explicit info might be easier to handle.

marmarek:
From a practical point of view, I don't like how device descriptions are now formatted for PCI devices. See https://openqa.qubes-os.org/tests/99725#step/qubesmanager_vmsettings/60, you can select another screenshot from the drop down and click "eye" icon on the right to compare old/new. Old descriptions had clear info what type of device is there, and then vendor + product. With new layout the device type is gone (sometimes product name suggests what that is, but not always). And also, IMHO "vendor product" is more intuitive than "product (vendor)", but this might be just being used to `lspci` output, so I'm not that attached to it.

piotrbartman:
I chose this description format to align with USB device descriptions; we can go back to the old one if preferred. Another matter is that the new API in general enable device description creation on the 'client side', where IMO decisions about which data to display and in what order should be made. Initially, we even considered removing the 'description' attribute and maybe it's not that bad idea...

marmarek:
Wouldn't that mean more devclass-specific parts on the client side? Or maybe that's unavoidable anyway?

piotrbartman:
Not really, the new API is based on `DeviceInfo`, which exposes (`vendor`, `manufacturer`, `product`, `name`, `serial`, `interfaces`, etc., which is common for all devclasses (sometimes it's just `None`/`"unknown"`).

marmarek:
> Not really, the new API is based on `DeviceInfo`, which exposes (`vendor`, `manufacturer`, `product`, `name`, `serial`, `interfaces`, etc., which is common for all devclasses (sometimes it's just `None`/`"unknown"`).

But the device type I'm talking about is specific to PCI (for example it doesn't make sense for block devices) - if the description would be constructed on the client side, it would need to have a special case for it, no?

piotrbartman:
This can be handled (and partially already is) by `device_info.interfaces[0].category`. It is partial because we choose a subset of all device categories/classes (like `Mouse`, `Keyboard`, etc.), but in the case of PCI, the category can be the exact category of PCI based on pci.ids (the same applies to USB). For block devices, the category is already `Mass_Data`, and for microphones: `Microphone`.

qubes/ext/block.py
piotrbartman: IIUC we prefer to have information like `usb:ident:interface` or `block:ident` in qubesdb? For now, I have proposed something in between, i.e., I use info from qubesdb to find device exposed by the VM

marmarek: IMO the (parent) device class should be explicit. It can be a prefix like you say here, or a separate qubesdb entry like it's in the DeviceInfo object serialization. The former is probably nicer?

marmarek:
pylint complains (and maybe even it's right this time?)

piotrbartman:
Interesting, I haven't changed this part lately... It not really a case since the `(sub)class_name` assignment is related to the `(sub)class_id`. I'll add an exception there

marmarek (approved):

PR #580, opened 2024-02-12
Enable S0ix workarounds with one knob

marmarek:
To enable all S0ix parts, do one single:

    qvm-features dom0 suspend-s0ix 1

It will enable "suspend" instead of "freeze" suspend mode of PCI-having
VMs (sys-net, sys-usb), and also assign thunderbolt controllers to
pciback which will force suspending them.

QubesOS/qubes-issues#6411

### DIFF
--- a/linux/aux-tools/startup-misc.sh
+++ b/linux/aux-tools/startup-misc.sh
@@ -17,3 +17,37 @@ fi
 
 # Hide mounted devices from qubes-block list (at first udev run, only / is mounted)
 udevadm trigger --action=change --subsystem-match=block
+
+bind_to_pciback() {
+    local sbdf="$1"
+
+    if [ -e "/sys/bus/pci/devices/$sbdf/driver" ]; then
+        echo "$sbdf" > "/sys/bus/pci/devices/$sbdf/driver/unbind"
+    fi
+    echo "$sbdf" > /sys/bus/pci/drivers/pciback/new_slot
+    echo "$sbdf" > /sys/bus/pci/drivers/pciback/bind
+}
+
+if [ -n "$(qvm-features dom0 suspend-s0ix)" ]; then
+    # assign thunderbolt root ports to pciback as workaround for suspend
+    # issue without PCI hotplut enabled, see
+    # https://github.com/QubesOS/qubes-linux-kernel/pull/903 for details
+    for dev in /sys/bus/pci/devices/0000:00:*; do
+        [ -h "$dev" ] || continue
+        sbdf=$(basename "$dev")
+        read -r dev_class < "$dev/class"
+        # PCIe bridge
+        if [ "$dev_class" = "0x060400" ]; then
+            # There seems to be no property saying it's thunderbolt other than product id...
+            lspci -s "$sbdf" | grep -q Thunderbolt || continue
+
+            bind_to_pciback "$sbdf"
+            echo "$sbdf" > /sys/bus/pci/drivers/pciback/qubes_exp_pm_suspend
+            echo "$sbdf" > /sys/bus/pci/drivers/pciback/qubes_exp_pm_suspend_force
+        elif [ "$dev_class" = "0x0c0340" ]; then
+            # Thunderbolt 4 NIH device
+            bind_to_pciback "$sbdf"
+            echo "$sbdf" > /sys/bus/pci/drivers/pciback/qubes_exp_pm_suspend
+        fi
+    done
+fi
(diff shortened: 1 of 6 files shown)


marmarek:
Yeah, this logic is wrong. Do you know how to match the TB root ports,
preferably by some property, not by name?

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

linux/aux-tools/startup-misc.sh
```diff
@@ -17,3 +17,32 @@ fi
 
 # Hide mounted devices from qubes-block list (at first udev run, only / is mounted)
 udevadm trigger --action=change --subsystem-match=block
+
+if [ -n "$(qvm-features dom0 suspend-s0ix)" ]; then
+    # assign thunderbolt root ports to pciback as workaround for suspend
+    # issue without PCI hotplut enabled, see
```
DemiMarie, line 23:
```suggestion
    # issue without PCI hotplug enabled, see
```

linux/aux-tools/startup-misc.sh
```diff
@@ -17,3 +17,32 @@ fi
 
 # Hide mounted devices from qubes-block list (at first udev run, only / is mounted)
 udevadm trigger --action=change --subsystem-match=block
+
+if [ -n "$(qvm-features dom0 suspend-s0ix)" ]; then
```
DemiMarie, line 21: Move this out of the `if`?  Right now if this fails the error is ignored.

linux/aux-tools/startup-misc.sh
```diff
@@ -17,3 +17,32 @@ fi
...
+    # assign thunderbolt root ports to pciback as workaround for suspend
+    # issue without PCI hotplut enabled, see
+    # https://github.com/QubesOS/qubes-linux-kernel/pull/903 for details
+    for dev in /sys/bus/pci/devices/0000:00:*; do
+        [ -h "$dev" ] || continue
+        sbdf=$(basename "$dev")
+        # PCIe bridge
+        if [ "$(cat "$dev/class")" = "0x060400" ]; then
```
DemiMarie, line 29: Use `read -r` instead of `cat`?  `read` is a shell builtin.

linux/aux-tools/startup-misc.sh
```diff
@@ -17,3 +17,32 @@ fi
...
+            lspci -s "$sbdf" | grep -q Thunderbolt || continue
+
+            if [ -e "$dev/driver" ]; then
+                echo "$sbdf" > "$dev/driver/unbind"
+            fi
+            echo "$sbdf" > /sys/bus/pci/drivers/pciback/new_slot
+            echo "$sbdf" > /sys/bus/pci/drivers/pciback/bind
+            echo "$sbdf" > /sys/bus/pci/drivers/pciback/qubes_exp_pm_suspend
+            echo "$sbdf" > /sys/bus/pci/drivers/pciback/qubes_exp_pm_suspend_force
+        elif [ "$(cat "$dev/class")" = "0x0c0340" ]; then
```
DemiMarie, line 40: `read` would save a fork and exec here.

linux/aux-tools/startup-misc.sh
marmarek: That's kinda intentional - missing feature should be treated as disabled.

qubes/vm/qubesvm.py
```diff
@@ -229,8 +229,11 @@ def _default_kernelopts(self):
             qubes.config.system_path['qubes_kernels_base_dir'],
             self.kernel)
     pci = bool(list(self.devices['pci'].persistent()))
+    extra_opts = ""
     if pci:
         path = os.path.join(kernels_dir, 'default-kernelopts-pci.txt')
+        if self.app.domains[0].features.get('suspend-s0ix', False):
+            extra_opts = " qubes_exp_pm_use_suspend=1"
```
marmarek, line 236: This ends up overriding default `swiotlb=2048` option, instead of extending.

linux/aux-tools/startup-misc.sh
```diff
@@ -17,3 +17,32 @@ fi
...
+            # Thunderbolt 4 NIH device
+            # it's already assigned to pciback by initramfs, so just set the
+            # flag (unless user deassigned it explicitly)
+            [ "$(basename "$(readlink "$dev/driver")")" = "pciback" ] || continue
```
marmarek, line 44: This was false assumption, the device got assigned to the `thunderbolt` driver.

DemiMarie: That’s not good!

marmarek: Oh, that's actually expected, because I (intentionally) do _not_ have `qubes.hide_all_usb` option set there. So, I need to adjust the script to handle that condition too.

DemiMarie: Phew!!!!  I was scared this was a security vulnerability in Qubes OS.

PR #581, opened 2024-02-23
tests: use pactl to switch source-outputs

marmarek:
pacmd does not work with PipeWire, but pactl does.

QubesOS/qubes-issues#8955

### DIFF
--- a/qubes/tests/integ/vm_qrexec_gui.py
+++ b/qubes/tests/integ/vm_qrexec_gui.py
@@ -224,7 +224,7 @@ class TC_00_AudioMixin(TC_00_AppVMMixin):
         self.loop.run_until_complete(asyncio.sleep(2))
         if record.returncode is not None:
             self.fail("Recording process ended prematurely: exit code {}, stderr: {}".format(
-                      record.returncode, record.stderr.read()))
+                      record.returncode, self.loop.run_until_complete(record.stderr.read())))
         try:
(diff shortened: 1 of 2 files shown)


qubes/tests/integ/vm_qrexec_gui.py
```diff
@@ -116,7 +117,12 @@ def check_audio_sample(self, sample, sfreq):
         rec_size = np.count_nonzero((rec > threshold) | (rec < -threshold))
         if not rec_size:
             self.fail('only silence detected, no useful audio data')
-        if rec_size < 0.95*441000:
+        margin = 0.95
+        if in_qemu and self.testvm1.features.get('audio-model'):
+            # be less strict on HVM tests in nested virt, the test environment
+            # has huge overhead already
+            margin = 0.80
```
DemiMarie, line 124: That make sense.  This is pretty much a worst-case scenario, and I would not be surprised if there are some operations that are particularly expensive here.

marmarek: Yes, but also it worked mostly fine with PA in dom0. That said, I don't want to spend more time on debugging this case.
BTW just switching to PW in dom0 made HVM audio horrible also on bare metal qubes, but then https://github.com/QubesOS/qubes-vmm-xen-stubdom-linux/pull/64 fixes it to be comparable with PA in dom0.

PR #589, opened 2024-05-03
ext/audio: set and reload stubdom pulseaudio module

fepitre:
Fixes QubesOS/qubes-issues#8975

### DIFF
--- a/qubes/ext/audio.py
+++ b/qubes/ext/audio.py
@@ -17,6 +17,8 @@
 # License along with this library; if not, see <https://www.gnu.org/licenses/>.
 #
 
+import asyncio
+import subprocess
 import qubes.config
 import qubes.ext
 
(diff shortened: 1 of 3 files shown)


marmarek:
Tests failed here. My guess is missing dom0 in the app.domains mock collection (and default_audiovm defaults to dom0).

fepitre:
> app.domains

I fixed it and added a specific test with audiovm not being dom0 in xml.

qubes/ext/audio.py
```diff
@@ -89,11 +91,39 @@ def on_property_reset(self, subject, event, name, oldvalue=None):
         newvalue = getattr(subject, "audiovm", None)
         self.on_property_set(subject, event, name, newvalue, oldvalue)
 
+    @staticmethod
+    async def set_stubdom_audiovm_domid(qube, audiovm):
+        try:
+            await qube.run_service_for_stdio(
+                f"qubes.SetAudioVM+{audiovm.xid}",
```
marmarek, line 98: This fails if `audiovm` is None.

templates/libvirt/xen.xml
```diff
@@ -1,3 +1,8 @@
+{% if vm.audiovm is defined %}
+ {% set audiovm_xid = vm.audiovm.xid %}
+{% else %}
+ {% set audiovm_xid = 0 %}
```
marmarek, line 4: Should this really default to 0 (dom0)? I'd say setting audiovm to None should result in some dummy value here (-1? empty?) which then avoids loading the vchan module (so only the null one is loaded). But setting audiovm to something later should still work.

But that's a minor issue. If you prefer to do that later, it's okay, but open an issue to not forget.

marmarek (changes requested):

qubes/ext/audio.py
```diff
@@ -89,11 +91,39 @@ def on_property_reset(self, subject, event, name, oldvalue=None):
...
+            return
+        has_stubdom_qrexec = subject.features.check_with_template(
+            'stubdom-qrexec', None)
+        has_audio_model = subject.features.check_with_template(
+            'audio-model', None)
+        if has_audio_model and not has_stubdom_qrexec:
+            subject.log.warning("Cannot change dynamically audiovm: qrexec"
+                                " is not available (stubdom-qrexec feature)")
+        if has_audio_model and has_stubdom_qrexec:
+            subject.log.warning(f"setting {newvalue.xid}")
```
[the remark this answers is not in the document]
fepitre: it's a leftover sorry.

templates/libvirt/xen.xml
fepitre: I'll set it to -1 so I need to adapt the stubdom part to not load module on start.

templates/libvirt/xen.xml
```diff
@@ -179,13 +184,13 @@
                     {% endif %}
                     {% if vm.netvm %}
```
marmarek, line 185: This all is added only if netvm is set. You need to add a case for no-netvm too.

marmarek (changes requested):

marmarek:
PipelineRetry

marmarek (approved):

marmarek:
While I like it now, CI doesn't. Pylint complains and some tests need an update...

marmarek:
PipelineRetry

qubes/ext/audio.py
```diff
@@ -89,23 +91,50 @@ def on_property_reset(self, subject, event, name, oldvalue=None):
...
         self.on_property_set(subject, event, name, newvalue, oldvalue)
 
+    @staticmethod
+    async def set_stubdom_audiovm_domid(qube, audiovm):
+        if audiovm:
+            audiovm_xid = audiovm.xid
+        else:
+            audiovm_xid = -1
+        try:
+            await qube.run_service_for_stdio(
```
ben-grande, line 101: https://github.com/QubesOS/qubes-issues/issues/10554#issuecomment-3729565433

ben-grande: Rather call it on `domain-start`?

ben-grande: I see, it was going to be dynamic, but only if `qube.is_running()`.

PR #593, opened 2024-05-17
tests: adjust for extended qubes pillar

marmarek:
Adjust for this:
https://github.com/QubesOS/qubes-mgmt-salt-base/pull/17


qubes/tests/integ/salt.py
```diff
@@ -253,6 +253,7 @@ def test_020_qubes_pillar(self):
         vm = self.app.add_new_vm('AppVM', label='red',
             name=vmname)
         self.loop.run_until_complete(vm.create_on_disk())
+        vm.features['vm-config.test'] = 'value'
```
ben-grande, line 256: What about adding a non-allowed feature? The test bellow `self.assertEqual(qubes_pillar, {...})` should not fail and still return only `'features': {'vm-config.test': 'value'}`.

marmarek: Good idea!

PR #595, opened 2024-05-31
Fix service name verification

marmarek:
It was too strict and in a wrong place.

Fixes QubesOS/qubes-issues#9274

### DIFF
--- a/qubes/ext/services.py
+++ b/qubes/ext/services.py
@@ -62,6 +62,12 @@ class ServicesExtension(qubes.ext.Extension):
             if not feature.startswith('service.'):
                 continue
             service = feature[len('service.'):]
+            if not service:
+                vm.log.warning("Empty service name, ignoring: " + service)
+                continue
+            if len(service) > 48:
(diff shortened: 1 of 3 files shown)


qubes/ext/services.py
```diff
@@ -62,6 +62,12 @@ def on_domain_qdb_create(self, vm, event):
             if not feature.startswith('service.'):
                 continue
             service = feature[len('service.'):]
+            if not service:
+                vm.log.warning("Empty service name, ignoring: " + service)
+                continue
```
DemiMarie, line 67: Should there be a check for `.`, `..`, and names containing `/`?  At the very least, creating a file named `/run/qubes-service/.`, `/run/qubes-service/..`, or `/run/qubes-service/a/b` won’t work.

marmarek: It may make sense, although it feels like it's more of an implementation detail of a specific Linux agent (and even if not file, you can still access it via qubesdb). But until changed, probably better to refuse for now instead of more or less silently failing to create the file in `/run/qubes-service` (or removing `..` or such).

PR #603, opened 2024-06-30
Avoid reusing disposable VM names for some predefined time

alimirjamali:
Fixes: https://github.com/QubesOS/qubes-issues/issues/8512

### DIFF
--- a/qubes/app.py
+++ b/qubes/app.py
@@ -427,6 +427,10 @@ class VMCollection:
     def __init__(self, app):
         self.app = app
         self._dict = {}
+        # Recently used disposable IDs: dispid -> destroy seconds since epoch
+        self._recent_dispids = {}
+        # Avoid reuse of disposable IDs for one week
+        self._no_dispid_reuse_period = 7 * 24 * 60 * 60
 
     def close(self):
         del self.app
@@ -538,6 +542,8 @@ class VMCollection:
                 pass
         del self._dict[vm.qid]
         self.app.fire_event('domain-delete', vm=vm)
+        if getattr(vm, 'dispid', None):
+            self._recent_dispids[getattr(vm, 'dispid')] = int(time.monotonic())
 
     def __contains__(self, key):
         return any((key in (vm, vm.qid, vm.name))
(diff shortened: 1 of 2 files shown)


rustybird:
Why not `time.monotonic()`?

alimirjamali:
> Why not `time.monotonic()`?

Oh. Forgot that. We had discussed datetime module. When I switched to time module (since it was already used in the code), I forgot to re-implement monotonic. I will amend the code and force push in a second.

alimirjamali:
Making Pylint happy

qubes/app.py
```diff
@@ -427,6 +427,10 @@ class VMCollection:
     def __init__(self, app):
         self.app = app
         self._dict = {}
+        # Recently used disposable IDs: dispid -> destroy seconds since epoch
+        self._recent_dispIDs = {}
```
ben-grande, line 431: Please don't use uppercase mixed with underscores, I don't see other parts of this file using it.

https://peps.python.org/pep-0008/#descriptive-naming-styles

`_recent_dispIDs` falls into the `Capitalized_Words_With_Underscored` category.

alimirjamali: > `_recent_dispIDs` falls into the `Capitalized_Words_With_Underscored` category.

Done.  Also renamed `self._no_dispID_reuse_period` to `self._no_dispid_reuse_period`

marmarek:
The change itself looks good, but can you add a unit test? Tests for VMCollection are in qubes/tests/app.py. And you can mock the `random.SystemRandom()` with something like
```
with unittest.mock.patch('random.SystemRandom') as random:
    # randrange will return 33, then 44, then 55
    random.return_value.randrange.side_effect = [33, 44, 55]
...
```

alimirjamali:
> The change itself looks good, but can you add a unit test?

Sure. Will work on it.

alimirjamali:
> but can you add a unit test?

I added the Unit Test in the second commit @marmarek .

Unit test number 999999 is inspired by the Dilbert comic strip URL which was added to the original function by fepitre.

qubes/tests/app.py
```diff
@@ -315,6 +317,20 @@ def test_100_get_new_unused_qid(self):
...
 
+    def test_999999_get_new_unused_dispid(self):
+        with mock.patch('random.SystemRandom') as random:
+            random.return_value.randrange.side_effect = [11, 22, 33, 44, 55, 66]
+            # Testing overal functionality
+            self.assertEqual(self.vms.get_new_unused_dispid(), 11)
+            self.assertEqual(self.vms.get_new_unused_dispid(), 22)
+            self.assertEqual(self.vms.get_new_unused_dispid(), 33)
+            # Testing no reuse
+            self.vms._recent_dispids[44] = 0
```
marmarek, line 328: This works only because CI runs tests in a fresh VM, so `time.monitonic()` is quite small. But it will fail if you run it on a system with over a week of uptime... Use `time.monitonic()` here? Or patch it with a constant value?

alimirjamali: Done

alimirjamali: Let me also add time.monotonic to the other test (66)

alimirjamali:
Ok. Test completed successfully.

Here is the link: https://gitlab.com/QubesOS/qubes-core-admin/-/jobs/7250348262#L3221

marmarek (approved):

PR #607, opened 2024-07-06
storage: increase the limit of block devices per qube

3hhh:
Fixes: QubesOS/qubes-issues#9311

### DIFF
--- a/qubes/ext/block.py
+++ b/qubes/ext/block.py
@@ -33,17 +33,13 @@ import qubes.device_protocol
 import qubes.devices
 import qubes.ext
 from qubes.ext.utils import device_list_change
+from qubes.storage import Storage
 
 name_re = re.compile(r"\A[a-z0-9-]{1,12}\Z")
 device_re = re.compile(r"\A[a-z0-9/-]{1,64}\Z")
 # FIXME: any better idea of desc_re?
 desc_re = re.compile(r"\A.{1,255}\Z")
 mode_re = re.compile(r"\A[rw]\Z")
-
-# all frontends, prefer xvdi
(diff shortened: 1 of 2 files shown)


qubes/storage/__init__.py
```diff
@@ -494,7 +494,14 @@ class Storage:
...
 
-    AVAILABLE_FRONTENDS = {'xvd' + c for c in string.ascii_lowercase}
+    # all frontends, prefer xvdi
+    # TODO: get this from libvirt driver?
+    AVAILABLE_FRONTENDS = ['xvd'+c for c in string.ascii_lowercase[8:]]
+    AVAILABLE_FRONTENDS += ['xvd'+c+d for c in string.ascii_lowercase
+                                      for d in string.ascii_lowercase]
+    # xvda - xvdh are reserved by Qubes OS and sometimes hidden from tools,
+    # so we put them to the end of the list
+    AVAILABLE_FRONTENDS += ['xvd'+c for c in string.ascii_lowercase[:8]]
```
alimirjamali, line 504: Changing the `AVAILABLE_FRONTENDS` from a python set to a python list should not be an issue. But line 499 and 504 appear to be the same? Save value is added to the list twice.

qubes/ext/block.py
```diff
@@ -33,18 +33,14 @@
 import qubes.devices
 import qubes.ext
 from qubes.ext.utils import device_list_change
+from qubes.storage import Storage
```
alimirjamali, line 36: Defining the available block devices at one place sounds like a good idea.

alimirjamali (approved):

marmarek (approved):

PR #608, opened 2024-07-08
tests/integ/audio.py: add test for audio-intput in QubesDB

fepitre:


qubes/tests/integ/audio.py
```diff
@@ -273,10 +284,9 @@ def common_audio_record_muted(self):
         if audio_in[:32] in recorded_audio:
             self.fail('VM recorded something, even though mic disabled')
 
-    def common_audio_record_unmuted(self):
-        deva = qubes.device_protocol.DeviceAssignment(self.app.domains[0], 'mic')
-        self.loop.run_until_complete(
-            self.testvm1.devices['mic'].attach(deva))
+    def common_audio_record_unmuted(self, attach_mic=True):
+        if attach_mic:
+            self.attach_mic()
```
marmarek, line 289: For some reason, this fails with DeviceAlreadyAttached error. In the old tests (without audiovm switching), for example `test_222_audio_rec_unmuted_pulseaudio`. I don't see what functional change in this place, so it might be an issue with the other PR.

marmarek (approved):

PR #613, opened 2024-08-16
Validate default_kernel global property on set

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/8992

### DIFF
--- a/qubes/app.py
+++ b/qubes/app.py
@@ -674,6 +674,28 @@ def _setter_default_netvm(app, prop, value):
     return value
 
 
+def validate_kernel(obj, property_name: str, kernel: str) -> None:
+    """Helper function to validate existence of specific kernel"""
+    if not kernel:
+        return
+    dirname = os.path.join(
+        qubes.config.qubes_base_dir,
+        qubes.config.system_path['qubes_kernels_base_dir'],
+        kernel)
+    if not os.path.exists(dirname):
+        raise qubes.exc.QubesPropertyValueError(
+            obj, obj.property_get_def(property_name), kernel,
+            'Kernel {!r} not installed'.format(
+                kernel))
+    for filename in ('vmlinuz',):
+        if not os.path.exists(os.path.join(dirname, filename)):
+            raise qubes.exc.QubesPropertyValueError(
+                obj, obj.property_get_def(property_name), kernel,
+                'Kernel {!r} not properly installed: '
+                'missing {!r} file'.format(
+                    kernel, filename))
+
+
 class Qubes(qubes.PropertyHolder):
     """Main Qubes application
 
@@ -1593,3 +1615,10 @@ class Qubes(qubes.PropertyHolder):
                 # resetting dispvm to its default value
                 vm.fire_event('property-reset:default_dispvm',
                               name='default_dispvm', oldvalue=oldvalue)
+
+    @qubes.events.handler('property-pre-set:default_kernel')
+    # pylint: disable-next=invalid-name
+    def on_property_pre_set_default_kernel(self, event, name, newvalue,
+                                           oldvalue=None):
+        # pylint: disable=unused-argument
+        validate_kernel(self, 'default_kernel', newvalue)
--- a/qubes/tests/__init__.py
+++ b/qubes/tests/__init__.py
@@ -458,6 +458,14 @@ class QubesTestCase(unittest.TestCase):
         self.loop = asyncio.get_event_loop()
         self.addCleanup(self.cleanup_loop)
 
+        self.kernel_validator_original = qubes.app.validate_kernel
+        def kernel_validator_patched(obj, key, value):
+            if value.startswith('unittest'):
+                self.kernel_validator_original(obj, key, value)
+        self.skip_kernel_validation_patch = unittest.mock.patch(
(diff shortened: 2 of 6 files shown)


alimirjamali:
Additional side notes:

I did not go to the extreme of doing it like how it is done in the [QubesVM](https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/vm/qubesvm.py#L1040-L1060) class. A simpler approach of just using the Kernel Pool is implemented. It is similar to how it is done in Qubes Manager.

alimirjamali:
Humm. The pylint complain is very well understandable:
```
qubes/app.py:1598:4: C0103: Method name "on_property_pre_set_default_kernel" doesn't conform to '[a-z_][a-z0-9_]{2,30}$' pattern (invalid-name)
```

On the other hand, the failed test is confusing:
```
    return iter(self._pool.list_volumes())
                ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gitlab-runner/builds/QubesOS/qubes-core-admin/qubes/storage/kernels.py", line 216, in list_volumes
    for kernel_version in os.listdir(self.dir_path)]
                          ^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/qubes/vm-kernels'
```

alimirjamali:
The nice fact about the two last failed tests is that it shows that the code is actually working.

1st. test failed since `/var/lib/qubes/vm-kernels` directory did not exist.
2nd. test failed since `/var/lib/qubes/vm-kernels/1.0` which was used by some unittest did not exist.

I guess there is no need for additional unit tests.

alimirjamali:
Fixing few remaining unit tests.

run-tests
```diff
@@ -19,6 +19,17 @@ if sudo --non-interactive "$name/ci/lvm-manage" setup-lvm vg$$/pool; then
     CLEANUP_LVM=yes
 fi
 
+CLEANUP_KERNEL_POOL=
+if [ ! -d "/var/lib/qubes/vm-kernels" ]; then
```
marmarek, line 23: This won't work when running tests in actual dom0 (vm-kernels dir exists, but "1.0" kernel doesn't). 
Some of the tests patch `qubes_kernels_base_dir`, see for example `test_260_kernelopts` in `qubes/tests/vm/qubesvm.py` which should be safer approach.

alimirjamali:
Nice. The tearDown function is purging `/tmp/qubes-test-dir`. So creating the `/tmp/qubes-test-dir/vm-kernels/1.0/vmlinuz` in the startup bash file is insufficient.

I am not going to abuse CI/CD server anymore. Reverting to local tests and will come back after I patched all unittests

run-tests
alimirjamali: Yes. Just recognized it. I will do the tests locally in order to avoid abusing the CI/CD servers and cluttering the logs. Will come back after I fix it properly

alimirjamali: @marmarek I finally managed to patch one of the major base classes (`AdminAPITestCase`). I believe that I can go forward and properly patch the rest.

I have a question regarding the CI/CD servers. Are they owned by Qubes project or they belong to Gitlab? I am trying to avoid using the online ones as much as possible and test the patches locally. But using the online ones are sometimes more convenient.

marmarek: > I have a question regarding the CI/CD servers. Are they owned by Qubes project or they belong to Gitlab? I am trying to avoid using the online ones as much as possible and test the patches locally. But using the online ones are sometimes more convenient.

We use our own runners, you can use them as much as you like :)

alimirjamali:
The approach of patching the existing untitests to allow `default_kernel` checkup is wrong. Some of them break in different scenarios. The proper approach would be mocking `on_property_pre_set_default_kernel` in relevant unittest classes to ignore it. Then write a single unittest for `on_property_pre_set_default_kernel` scenario. I will do it tomorrow.

alimirjamali:
Ok. We have passed CI checks and Codecov is advising on the new Unittests. I should use the helper function for QubesVM as well and add the necessary unittests. Hopefully tomorrow

alimirjamali:
@marmarek  This is ready for review

qubes/app.py
```diff
@@ -674,6 +674,28 @@ def _setter_default_netvm(app, prop, value):
...
+    """Helper function to validate existence of specific kernel"""
+    if not kernel:
+        return
+    dirname = os.path.join(
+        qubes.config.qubes_base_dir,
+        qubes.config.system_path['qubes_kernels_base_dir'],
+        kernel)
+    if not os.path.exists(dirname):
+        raise qubes.exc.QubesPropertyValueError(
+            obj, property_name, kernel,
```
marmarek, line 687: the second argument is supposed to be QubesProperty object, not str. `obj.property_get_def(property_name)` should do the trick

marmarek (changes requested):

qubes/app.py
alimirjamali: > the second argument is supposed to be QubesProperty object, not str. `obj.property_get_def(property_name)` should do the trick

It is done. Unittests re-run succeeded.

marmarek (approved):

PR #614, opened 2024-08-20
device identity

piotrbartman:
implements: QubesOS/qubes-issues/issues/9325

### DIFF
--- a/doc/qubes-devices.rst
+++ b/doc/qubes-devices.rst
@@ -6,8 +6,8 @@ devices, which can be attached to other domains (frontend). Devices can be of
 different buses (like 'pci', 'usb', etc.). Each device bus is implemented by
 an extension (see :py:mod:`qubes.ext`).
 
-Devices are identified by pair of (backend domain, `ident`), where `ident` is
-:py:class:`str` and can contain only characters from `[a-zA-Z0-9._-]` set.
+Devices are identified by pair of (backend domain, `port_id`), where `port_id`
+is :py:class:`str` and can contain only characters from `[a-zA-Z0-9._-]` set.
 
 
 Device Assignment vs Attachment
@@ -106,11 +106,11 @@ is connected. Therefore, when assigning a device to a VM, such as
 `sys-usb:1-1.1`, the port `1-1.1` is actually assigned, and thus
 *every* devices connected to it will be automatically attached.
 Similarly, when assigning `vm:sda`, every block device with the name `sda`
-will be automatically attached. We can limit this using :py:meth:`qubes.device_protocol.DeviceInfo.self_identity`, which returns a string containing information
+will be automatically attached. We can limit this using :py:meth:`qubes.device_protocol.DeviceInfo.device_id`, which returns a string containing information
 presented by the device, such as, `vendor_id`, `product_id`, `serial_number`,
-and encoded interfaces. In the case of block devices, `self_identity`
+and encoded interfaces. In the case of block devices, `device_id`
 consists of the parent port to which the device is connected (if any),
-the parent's `self_identity`, and the interface/partition number.
+the parent's `device_id`, and the interface/partition number.
 In practice, this means that, a partition on a USB drive will only be
 automatically attached to a frontend domain if the parent presents
 the correct serial number etc., and is connected to a specific port.
--- a/doc/qubes-devices.rst
+++ b/doc/qubes-devices.rst
@@ -1,25 +1,80 @@
 :py:mod:`qubes.devices` -- Devices
-===================================
+==================================
 
-Main concept is that some domain (backend) may expose (potentially multiple)
-devices, which can be attached to other domains (frontend). Devices can be of
-different buses (like 'pci', 'usb', etc.). Each device bus is implemented by
-an extension (see :py:mod:`qubes.ext`).
+The main concept is that a domain (backend) can expose (potentially multiple)
+devices, each through a port, where only one device can be in one port at
+any given time. Such devices can be connected to other domains (frontends).
+Devices can be of different buses (like 'pci', 'usb', etc.). Each device bus
+is implemented by an extension (see :py:mod:`qubes.ext`).
 
-Devices are identified by pair of (backend domain, `port_id`), where `port_id`
-is :py:class:`str` and can contain only characters from `[a-zA-Z0-9._-]` set.
+Devices are identified by a pair (`port`, `device_id`), where `port` is a pair
+(backend domain, `port_id`). Both `port_id` and `device_id` are :py:class:`str`,
+and in addition, port_id is unique per backend. More about the requirements
+for `port_id` and `device_id` can be found in the sections below.
 
+Classes
+-------
+
+:py:class:`qubes.device_protocol.Port`: a pair `<backend_domain>:<port_id>`
+with `devclass` (e.g., `pci`, `usb`). In the previous version (before
+QubesOS 4.3), this was referred to as `Device`, and `port_id` was named `ident`.
+
+:py:class:`qubes.device_protocol.AnyPort`: A class used to handle cases where
+any port is accepted.
+
+:py:class:`qubes.device_protocol.VirtualDevice`: A pair `<port>:<device_id>`.
+This class links a device identified by `device_id` to a specific port.
+If both values are specified, the instance represents a device connected to
+that particular port. If the port is of type `AnyPort`, it represents a device
+identified by `device_id` that can be connected to any port. This is used by
+:py:class:`qubes.device_protocol.DeviceInfo`, which describes what to do with
+a device identified by `device_id` when connected anywhere. Similarly,
+when `device_id` is `*`, the instance represents any potential device
+connected to the given port. As a result, the device is considered "virtual"
+meaning it may or may not represent an actual device in the system.
+A device with `*:*` (any port and any device) is not permitted.
+
+:py:class:`qubes.device_protocol.DeviceInfo`: Derived from `VirtualDevice`.
+Extensions should assume that `Port` is provided, and based on that,
+`device_id` should return the same string for the same device, regardless of
+which port it is connected to. The `device_id` acts as a device hash
+and *should* be "human-readable". It must contain only digits, ASCII letters,
+spaces, and the following characters: `!#$%&()*+,-./:;<>?@[\]^_{|}~`.
+It cannot be empty or equal to `*`.
+
+:py:class:`qubes.device_protocol.DeviceAssignment`: Represents the relationship
+between a `VirtualDevice` and a `frontend_domain`. There are four modes:
+#. `manual` (attachment): The device is manually attached to `frontend_domain`.
+This type of assignment does not persist across domain restarts.
+#. `auto-attach`: Any device that matches a `VirtualDevice` will be
+automatically attached to the `frontend_domain` when discovered
+or during domain startup.
+#. `ask-to-attach`: Functions like `auto-attach`, but prompts the user for
+confirmation before attaching. If no GUI is available, the prompt is ignored.
+#. `required`: The device must be available during `frontend_domain` startup and
+will be attached before the domain is started.
+
+:py:class:`qubes.device_protocol.DeviceInterface`: Represents device interfaces
+as a 7-character code in the format `BCCSSII`, where `B` indicates the devclass
+(e.g., `p` for PCI, `u` for USB, `?` for unknown), `CC` is the class code,
+`SS` is the subclass code, and `II` represents the interface code.
+
+:py:class:`qubes.device_protocol.DeviceCategory`: Provides an easy-to-use,
+arbitrary subset of interfaces with names assigned to categories considered as
+most relevant to users. When needed, the class should be extended with new
+categories. This structure allows for quick identification of the device type
+and can be useful when displaying devices to the end user.
 
 Device Assignment vs Attachment
 -------------------------------
 
-:py:class:`qubes.device_protocol.DeviceAssignment` describes the assignment of a device
-to a frontend VM. For clarity let's us introduce two types of assignments:
+For clarity let's us introduce two types of assignments:
 *potential* and *real* (attachment). Attachment indicates that the device
 has been attached by the Qubes backend to its frontend VM and is visible
 from its perspective. Potential assignment, on the other hand,
-has two additional options: `automatically_attach` and `required`.
-For detailed descriptions, refer to the `DeviceAssignment` documentation.
+has tree modes: `auto-attach`, `ask-to-attach` and `required`.
+For detailed descriptions, take a look at
+:py:class:`qubes.device_protocol.DeviceAssignment` documentation.
 In general we refer to potential assignment as assignment
 and real assignment as attachment. To check whether the device is currently
 attached, we check :py:meth:`qubes.device_protocol.DeviceAssignment.attached`,
@@ -28,17 +83,35 @@ we check :py:meth:`qubes.device_protocol.DeviceAssignment.attach_automatically`.
 Potential and real connections may coexist at the same time,
 in which case both values will be true.
 
+Understanding Device Identity
+-----------------------------
+
+It is important to understand that :py:class:`qubes.device_protocol.Port` does not
+correspond to the device itself, but rather to the *port* to which the device
+is connected. Therefore, when assigning a device to a VM, such as
+`sys-usb:1-1.1`, the port `1-1.1` is actually assigned, and thus
+*every* devices connected to it will be automatically attached.
+Similarly, when assigning `vm:sda`, every block device with the name `sda`
+will be automatically attached. We can limit this using
+:py:meth:`qubes.device_protocol.DeviceInfo.device_id`, which returns a string
+containing information presented by the device, such as for example
+`vendor_id`, `product_id`, `serial_number`, and encoded interfaces.
+In the case of block devices, `device_id` consists of the parent's `device_id`
+to which the device is connected (if any) and the interface/partition number.
+In practice, this means that, a partition on a USB drive will only
+be automatically attached to a frontend domain if the parent presents
+the correct serial number etc.
 
 Actions
 -------
 
-The `assign` action signifies that a device will be assigned to the frontend VM
+The `assign` action means that a device will be assigned to the frontend VM
 in a potential form (this does not change the current system state).
 This will result in an attempt to automatically attach the device
-upon the next VM startup. If `required=True`, and the device cannot be attached,
+upon the next VM startup. If `mode=required`, and the device cannot be attached,
 the VM startup will fail. Additionally, upon device detection (`device-added`),
 an attempt will be made to attach the device. However, at any time
-(unless `required=True`), the user can manually modify this state by performing
+(unless `mode=required`), the user can manually modify this state by performing
 `attach` or `detach` on the device, changing the current system state.
 This will not alter the assignment, and automatic attachment attempts
 will still be made in the future. To remove the assignment the user
@@ -48,23 +121,23 @@ Assignment Management
 ---------------------
 
 Assignments can be edited at any time: regardless of whether the VM is running
-or the device is currently attached. An exception is `required=True`,
-in which case the VM must be shut down. Removing the assignment does not change the real system state, so if the device is currently attached
-and the user remove the assignment, it will not be detached,
-but it will not be automatically attached in the future.
-Similarly, it works the other way around with `assign`.
+or the device is currently attached. Removing the assignment does not change
+the real system state, so if the device is currently attached and the user
+remove the assignment, it will not be detached, but it will not be
+automatically attached in the future. Similarly, it works the other way
+around with `assign`.
 
 Proper Assignment States
 ------------------------
 
 In short, we can think of device assignment in terms of three flags:
 #. `attached` - indicating whether the device is currently assigned,
-#.  `attach_automatically` - indicating whether the device will be
+#. `attach_automatically` - indicating whether the device will be
 automatically attached by the system daemon,
 #. `required` - determining whether the failure of automatic attachment should
 result in the domain startup being interrupted.
 
-Then the proper states of assignment
+Then the possible states of assignment
 (`attached`, `automatically_attached`, `required`) are as follow:
 #. `(True, False, False)` -> domain is running, device is manually attached
 and could be manually detach any time.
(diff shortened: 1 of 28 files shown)


marmarek:
Can you drop commits adding gui, instead of adding+removing gui? the PR is quite big already...

qubes/ext/utils.py
```diff
@@ -100,3 +138,19 @@ def compare_device_cache(vm, devices_cache, current_devices):
...
+            ["attach-confirm", device.backend_domain.name,
+            device.port_id, device.description,
+             *[f.name for f in frontends.keys()]],
+            stdout=subprocess.PIPE, stderr=subprocess.PIPE
+        )
+        (target_name, _) = proc.communicate()
+        return target_name.decode()
+    except Exception as exc:
+        print("attach-confirm", exc, file=sys.stderr)
+        return ""
```
marmarek, line 156: There are a few problems with this function:
1. It must be async, since it may block for some time, and it's not acceptable to block the whole qubesd for this time. In fact, attach-confirm probably won't work this way at all if it tries to talk to qubesd, since it's blocked on waiting for attach-confirm...
2. subprocess.Popen -> asyncio variant
3. The tool name is IMO too generic for a tool in a common /usr/bin/
4. The tool belongs to desktop-linux-manager repo, which looks like a layering violation - dom0 code should also work without any of the GUI frontends installed.
5. Extension of the above: ensure it will work also with GUI domain. See how policy prompts are done for example. And this also will need adjustment to the threat model: verify the response is one of allowed ones (on the `frontends` list?)

marmarek (changes requested):
Just one thing noticed so far (not a full review yet)

qubes/ext/utils.py
marmarek: I see the code of `attach-confirm` already is async. Maybe simply put its code here, instead of calling external program? It means you will need to keep the `params` dict format in sync, but changes there needs to be done in compatible way anyway (due to the GUI domain case, where both ends may be updated independently). Plus, you won't need to make external `get_system_info()` call, as by running inside qubesd you already have all the info here.

qubes/devices.py
```diff
@@ -250,25 +256,24 @@ def load_assignment(self, device_assignment: DeviceAssignment):
...
         assert device_assignment.attach_automatically
-        device_assignment.devclass = self._bus
         self._set.add(device_assignment)
 
-    async def update_required(self, device: Device, required: bool):
+    async def update_assignment(
+            self, device: VirtualDevice, mode: AssignmentMode
+    ):
         """
         Update `required` flag of an already attached device.
```
marmarek, line 265:
```suggestion
        Update assignment mode of an already attached device.
```

qubes/devices.py
```diff
@@ -277,25 +282,27 @@ async def update_required(self, device: Device, required: bool):
 
         # be careful to use already present assignment, not the provided one
         # - to not change options as a side effect
-        if assignment.required == required:
+        if assignment.mode == mode:
             return
```
marmarek, line 287: Should this refuse mode `manual`? Or maybe automatically convert to unassign call?

qubes/devices.py
```diff
@@ -63,10 +63,13 @@
...
 import qubes.exc
 import qubes.utils
-from qubes.device_protocol import (Device, DeviceInfo, UnknownDevice,
-                                   DeviceAssignment)
+from qubes.device_protocol import (Port, DeviceInfo, UnknownDevice,
+                                   DeviceAssignment, VirtualDevice,
+                                   AssignmentMode)
 
 
+DEVICE_DENY_LIST = "/etc/qubes/device-deny.list"
```
marmarek, line 71: As discussed on the summit, this wants to be a directory (`/etc/qubes/device-deny.list.d`?)

qubes/ext/admin.py
```diff
@@ -160,3 +164,37 @@ def on_tag_add(self, vm, event, tag, **kwargs):
...
+                        deny[name] = deny.get(name, set()).union(set(values))
+        except IOError:
+            pass
+
+        # check if any presented interface is on deny list
+        for interface in deny.get(dest.name, set()):
+            pattern = DeviceInterface(interface)
+            for devint in device.interfaces:
+                if pattern.matches(devint):
+                    raise qubes.exc.PermissionDenied()
```
marmarek, line 200: This should be IMO a helper function (or two).

qubes/ext/block.py
```diff
@@ -548,20 +541,40 @@ def pre_attachment_internal(
...
+        for assignment in to_attach.values():
+            asyncio.ensure_future(self.attach_and_notify(vm, assignment))
+
+    async def attach_and_notify(self, vm, assignment):
         # bypass DeviceCollection logic preventing double attach
-        # we expected that these devices are already attached to this vm
-        self.notify_auto_attached(vm, device, options)
+        device = assignment.device
+        if assignment.mode.value == "ask-to-attach":
+            if vm.name != confirm_device_attachment(device, {vm: assignment}):
```
marmarek, line 572: as noted before, this needs to be async

qubes/ext/utils.py
```diff
@@ -54,24 +61,55 @@ def device_list_change(
...
+                    if curr is None or curr < ass:
+                        # chose the most specific assignment
+                        frontends[front_vm] = ass
+                    to_attach[device.port_id] = frontends
+
+    for port_id, frontends in to_attach.items():
+        if len(frontends) > 1:
+            # unique
+            device = tuple(frontends.values())[0].device
+            target_name = confirm_device_attachment(device, frontends)
```
marmarek, line 89: and this needs to be async too

qubes/api/admin.py
marmarek, line 1: There should be also Admin API methods for reading/writing deny list. If nothing more fancy, then at least similar to qrexec policy ones that:
1. Verify syntax when writing
2. Allow race-free edits (read returns a "token" being hash of the file, and write gets the token to verify if nobody changed it in the meantime).

If you prefer to add it in a separate PR, convert this note to an issue.

marmarek (changes requested):
This is a fuller review of this part. I haven't done more testing since last time.
And also, it got few conflicts in the meantime...

There needs to be a documentation somewhere about API changes (compared to R4.2). It should include the following:
1. API for device class extensions (like the one for mic or usb)
2. Extension controlling Admin API access to devices (extension that handles `admin-permission:...` events)
3. Usage of the client part (that's more relevant for the core-admin-client part)
4. Changes to qrexec methods (in case somebody was using Admin API directly, or has alternative client implementation)

It doesn't need to be very detailed, more like a checklist of changes to look into when updating 3rd-party code interacting with devices.

qubes/ext/utils.py
piotrbartman: > * The tool belongs to desktop-linux-manager repo, which looks like a layering violation - dom0 code should also work without any of the GUI frontends installed in dom0.

I agree that this to some level introduces a dependency of the lower layer on the upper one, and that's why I decided to implement it as an independent program. If we can think of another way to confirm device attaching, it should be easy to implement. Currently, however, the GUI is required to implement `ask-to-attach`, and if the package is not available, such an assignment will simply be ignored. So the implementation is now: for `ask-to-attach`, ask the user for confirmation if possible, otherwise ignore the assignment.

piotrbartman:
# New Device API

## Classes

- **Port**: A pair `<backend_domain>:<port_id>` with `devclass` (e.g., `pci`, `usb`). In the previous version, this was referred to as `Device`, and `port_id` was named `ident`.

- **AnyPort**: A class used to handle cases where any port is accepted.

- **VirtualDevice**: A pair `<port>:<device_id>`. This class links a device identified by `device_id` to a specific port. If both values are specified, the instance represents a device connected to that particular port. If the port is of type `AnyPort`, it represents a device identified by `device_id` that can be connected to any port. This is used by `DeviceAssignment`, which describes what to do with a device identified by `device_id` when connected anywhere. Similarly, when `device_id` is `*`, the instance represents any potential device connected to the given port. As a result, the device is considered "virtual" meaning it may or may not represent an actual device in the system. A device with `*:*` (any port and any device) is not permitted.

- **DeviceInfo**: Derived from `VirtualDevice`. Extensions should assume that `Port` is provided, and based on that, `device_id` should return the same string for the same device, regardless of which port it is connected to. The `device_id` acts as a device hash and *should* be "human-readable". It must contain only digits, ASCII letters, spaces, and the following characters: `!#$%&()*+,-./:;<>?@[\]^_{|}~`. It cannot be empty or equal to `*`.

- **DeviceAssignment**: Represents the relationship between a `VirtualDevice` and a `frontend_domain`. There are four modes:
  - **manual**: The device is manually attached to `frontend_domain`. This type of assignment does not persist across domain restarts.
  - **auto-attach**: Any device that matches a `VirtualDevice` will be automatically attached to the `frontend_domain` when discovered or during domain startup.
  - **ask-to-attach**: Functions like `auto-attach`, but prompts the user for confirmation before attaching. If no GUI is available, the prompt is ignored.
  - **required**: The device must be available during `frontend_domain` startup and will be attached before the domain is started.

## Device Assignments

If a connected device has multiple assignments to different `frontend_domain` instances, the user will be asked to choose which domain to connect the device to. If no GUI client is available, the device will not be connected to any domain.

If multiple assignments exist for a connected device with different options but to the same `frontend_domain`, the most specific assignment will take precedence, according to the following order (from highest to lowest priority):
1. Assignment specifies both `port` and `device_id`.
2. Assignment specifies only the `port`.
3. Assignment specifies only the `device_id`.

It is important to note that only one matching assignment can exist within each of the categories listed above.

## API Changelog

- New/Renamed Admin Methods for device assignments:
  - `admin.vm.device.<devclass>.Assign`
  - `admin.vm.device.<devclass>.Unassign`
  - `admin.vm.device.<devclass>.Assigned`
  - `admin.vm.device.<devclass>.List` → `admin.vm.device.<devclass>.Attached`
  - `admin.vm.device.<devclass>.Set.persistent` → `admin.vm.device.<devclass>.Set.assignment`, which accepts one of the following arguments: `auto-attach`, `ask-to-attach`, or `required`.

- Event Firing Changes with New Argument Set:
  - `admin.vm.device.<devclass>.Attach`: (device, devclass, persistent, options) → (device, mode, options)
  - `admin.vm.device.<devclass>.Detach`: (device, devclass) → (device)
  - `admin.vm.device.{endpoint}.Set.persistent`: (device, persistent) → (device, mode)
  - `admin.vm.device.<devclass>.Assign`: (device, mode, options)
  - `admin.vm.device.<devclass>.Unassign`: (device)

- Extensions should provide a class derived from `DeviceInfo`. This class must provide at least `device_id` according to the rules listed in the `DeviceInfo` description.

- Device extensions are responsible for emitting `device-detach:<devclass>` events during domain shutdown if any `<devclass>` device was attached.

- To support auto-assignment, device extensions need to:
  - Implement `async attach_and_notify`, which attaches the device and emits `device-pre-attach:<devclass>` and `device-attach:<devclass>`.
  - Handle the `domain-start` event and check if any device should be attached at startup.

qubes/ext/utils.py
marmarek: For sys-gui case, since this is socket protocol already, it's very easy to pipe it to a qrexec service. As said above, I'd prefer the interface between those two packages be "send this to a socket, read response in this format", instead of "call this executable". It will allow adding qrexec support without needing to still have the GUI-related package installed in dom0.

And yes, for now the approach "deny if confirmation not available" is fine.

marmarek (changes requested):

marmarek:
As for the API changes, looks good. Should it be new file in `doc/` here (linked from index and `doc/qubes-devices.rst`)? Or maybe added to the existing `doc/qubes-devices.rst` directly?

Just minor thing:
> `admin.vm.device.{endpoint}.Set.persistent` - renamed too, and `{endpoint}` remained

qubes/ext/utils.py
```diff
@@ -100,3 +159,56 @@ def compare_device_cache(vm, devices_cache, current_devices):
...
+            for dom in doms.values()
+        },
+    }
+
+    socked_call = asyncio.create_task(
+        call_socket_service(guivm, socket, "dom0", params, SOCKET_PATH)
+    )
+
+    while not socked_call.done():
+        await asyncio.sleep(0.1)
```
marmarek, line 206: Why this loop? the `await socked_call` below should be enough, no?

qubes/ext/utils.py
```diff
@@ -100,3 +159,56 @@ def compare_device_cache(vm, devices_cache, current_devices):
...
+
+    params = {
+        "source": device.backend_domain.name,
+        "device_name": device.description,
+        "argument": device.port_id,
+        "targets": front_names,
+        "default_target": front_names[0] if number_of_targets == 1 else "",
+        "icons": {
+            (
+                dom.name if dom.klass != "DispVM" else f"@dispvm:{dom.name}"
```
marmarek, line 195: shouldn't that be `getattr(dom, "template_for_dispvms", False)` check instead? DispVM class is already running one, not a template to create new one from.

marmarek (changes requested):

doc/qubes-devices.rst
```diff
@@ -79,51 +152,53 @@ because either (i) domain is halted, device (ii) manually detached or
...
 
+Note that if `required=True` then `automatically_attached=True`.
+
+Conflicted Assignments
+----------------------
+
+If a connected device has multiple assignments to different `frontend_domain`
+instances, the user will be asked to choose which domain connect the device to.
+If no GUI client is available, the device will not be connected to any domain.
+beacon how to get mev
```
marmarek, line 163: Cat on keyboard pressed "paste"? ;)

qubes/ext/utils.py
piotrbartman: Yes, it's for running DispVMs to indicate in the vm list which vms are disposible

qubes/ext/utils.py
piotrbartman: If we don't wait for the task to complete, using await in this case results in tasks being destroyed (including resolve_conflicts_and_attach) so attachment is skipped. I'm not entirely sure what causes this.

marmarek: Weird, I'll try to find out what is going on.

marmarek:
First start of sys-usb fails:
```
Oct 26 22:49:26 dom0 kernel: xhci_hcd 0000:05:00.3: USB bus 1 deregistered
Oct 26 22:49:26 dom0 kernel: pciback 0000:05:00.3: xen_pciback: seizing device
Oct 26 22:49:26 dom0 kernel: xen: registering gsi 37 triggering 0 polarity 1
Oct 26 22:49:26 dom0 kernel: Already setup the GSI :37
Oct 26 22:49:26 dom0 kernel: xhci_hcd 0000:05:00.4: remove, state 4
Oct 26 22:49:26 dom0 kernel: usb usb4: USB disconnect, device number 1
Oct 26 22:49:26 dom0 kernel: xhci_hcd 0000:05:00.4: USB bus 4 deregistered
Oct 26 22:49:26 dom0 kernel: xhci_hcd 0000:05:00.4: remove, state 1
Oct 26 22:49:26 dom0 kernel: usb usb3: USB disconnect, device number 1
Oct 26 22:49:26 dom0 kernel: usb 3-3: USB disconnect, device number 2
Oct 26 22:49:26 dom0 kernel: audit: type=1137 audit(1729982966.997:111): pid=2308 uid=0 auid=4294967295 ses=4294967295 msg='op="removed-device" device="/devices/pci0000:00/0000:00:08.1/0000:05:00.4/usb4" device_rule=616C6C6F7720696420316436623A30303033206E616D652022784843492048>
Oct 26 22:49:26 dom0 audit[2308]: USER_DEVICE pid=2308 uid=0 auid=4294967295 ses=4294967295 msg='op="removed-device" device="/devices/pci0000:00/0000:00:08.1/0000:05:00.4/usb4" device_rule=616C6C6F7720696420316436623A30303033206E616D6520227848434920486F737420436F6E74726F6C6C657>
Oct 26 22:49:27 dom0 audit[2308]: USER_DEVICE pid=2308 uid=0 auid=4294967295 ses=4294967295 msg='op="removed-device" device="/devices/pci0000:00/0000:00:08.1/0000:05:00.4/usb3/3-3" device_rule=616C6C6F7720696420316436623A30313034206E616D652022436F6D706F73697465204B564D204465766>
Oct 26 22:49:27 dom0 audit[2308]: USER_DEVICE pid=2308 uid=0 auid=4294967295 ses=4294967295 msg='op="removed-device" device="/devices/pci0000:00/0000:00:08.1/0000:05:00.4/usb3/3-4" device_rule=626C6F636B20696420303430383A35333433206E616D65202248502048442043616D65726122207669612>
Oct 26 22:49:27 dom0 audit[2308]: USER_DEVICE pid=2308 uid=0 auid=4294967295 ses=4294967295 msg='op="removed-device" device="/devices/pci0000:00/0000:00:08.1/0000:05:00.4/usb3" device_rule=616C6C6F7720696420316436623A30303032206E616D6520227848434920486F737420436F6E74726F6C6C657>
Oct 26 22:49:27 dom0 kernel: usb 3-4: USB disconnect, device number 3
Oct 26 22:49:27 dom0 kernel: xhci_hcd 0000:05:00.4: USB bus 3 deregistered
Oct 26 22:49:27 dom0 kernel: pciback 0000:05:00.4: xen_pciback: seizing device
Oct 26 22:49:27 dom0 kernel: xen: registering gsi 38 triggering 0 polarity 1
Oct 26 22:49:27 dom0 kernel: Already setup the GSI :38
Oct 26 22:49:27 dom0 qubesd[2321]: vm.sys-usb: Start failed: Node device not found: no node device with matching name 'usb_usb3'
Oct 26 22:49:27 dom0 qubesd[2321]: vm.sys-usb: start failed
Oct 26 22:49:27 dom0 qubesd[2321]: Traceback (most recent call last):
Oct 26 22:49:27 dom0 qubesd[2321]:   File "/usr/lib/python3.13/site-packages/qubes/api/admin.py", line 868, in vm_start
Oct 26 22:49:27 dom0 qubesd[2321]:     await self.dest.start()
Oct 26 22:49:27 dom0 qubesd[2321]:   File "/usr/lib/python3.13/site-packages/qubes/vm/dispvm.py", line 253, in start
Oct 26 22:49:27 dom0 qubesd[2321]:     await super().start(**kwargs)
Oct 26 22:49:27 dom0 qubesd[2321]:   File "/usr/lib/python3.13/site-packages/qubes/vm/qubesvm.py", line 1184, in start
Oct 26 22:49:27 dom0 qubesd[2321]:     for device in ass.devices:
Oct 26 22:49:27 dom0 qubesd[2321]:                   ^^^^^^^^^^^
Oct 26 22:49:27 dom0 qubesd[2321]:   File "/usr/lib/python3.13/site-packages/qubes/device_protocol.py", line 1285, in devices
Oct 26 22:49:27 dom0 qubesd[2321]:     for dev in self.backend_domain.devices[self.devclass]:
Oct 26 22:49:27 dom0 qubesd[2321]:                ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
Oct 26 22:49:27 dom0 qubesd[2321]:   File "/usr/lib/python3.13/site-packages/qubes/devices.py", line 438, in get_exposed_devices
Oct 26 22:49:27 dom0 qubesd[2321]:     yield from self._vm.fire_event("device-list:" + self._bus)
Oct 26 22:49:27 dom0 qubesd[2321]:                ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Oct 26 22:49:27 dom0 qubesd[2321]:   File "/usr/lib/python3.13/site-packages/qubes/events.py", line 195, in fire_event
Oct 26 22:49:27 dom0 qubesd[2321]:     sync_effects, async_effects = self._fire_event(event, kwargs,
Oct 26 22:49:27 dom0 qubesd[2321]:                                   ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
Oct 26 22:49:27 dom0 qubesd[2321]:         pre_event=pre_event)
Oct 26 22:49:27 dom0 qubesd[2321]:         ^^^^^^^^^^^^^^^^^^^^
Oct 26 22:49:27 dom0 qubesd[2321]:   File "/usr/lib/python3.13/site-packages/qubes/events.py", line 168, in _fire_event
Oct 26 22:49:27 dom0 qubesd[2321]:     effects.extend(effect)
Oct 26 22:49:27 dom0 qubesd[2321]:     ~~~~~~~~~~~~~~^^^^^^^^
Oct 26 22:49:27 dom0 qubesd[2321]:   File "/usr/lib/python3.13/site-packages/qubes/ext/pci.py", line 358, in on_device_list_pci
Oct 26 22:49:27 dom0 qubesd[2321]:     if "pci" not in dev.listCaps():
Oct 26 22:49:27 dom0 qubesd[2321]:                     ~~~~~~~~~~~~^^
Oct 26 22:49:27 dom0 qubesd[2321]:   File "/usr/lib64/python3.13/site-packages/libvirt.py", line 6624, in listCaps
Oct 26 22:49:27 dom0 qubesd[2321]:     raise libvirtError('virNodeDeviceListCaps() failed')
Oct 26 22:49:27 dom0 qubesd[2321]: libvirt.libvirtError: Node device not found: no node device with matching name 'usb_usb3'
Oct 26 22:49:27 dom0 qvm-start[3421]: Error: Start failed: Node device not found: no node device with matching name 'usb_usb3', see /var/log/libvirt/libxl/libxl-driver.log for details
```
Retrying sys-usb start works.

This one is weird, but I think it's related to some caching, maybe on libvirt side? See the USB controllers were disconnected from dom0 just above, yet "usb_usb3" was still listed there. I think the cleanest way is to catch this exception in `dev.listCaps()` call and skip the device (following logic as if caps can't be listed, it isn't "pci"). Alternatively maybe there is some  cache flush to be done, but that still may be racy (if you have two usbvms for example)...

marmarek:
Looking at the traceback a bit closer, here:

> Oct 26 22:49:27 dom0 qubesd[2321]:     for device in ass.devices:

it tries to get device for a specific assignment (so, it knows which device it wants). Yet it results in

> Oct 26 22:49:27 dom0 qubesd[2321]:   File "/usr/lib/python3.13/site-packages/qubes/devices.py", line 438, in get_exposed_devices
> Oct 26 22:49:27 dom0 qubesd[2321]:     yield from self._vm.fire_event("device-list:" + self._bus)

... listing all of them. This sounds quite bad performance-wise (quadratic number of calls to whatever device backend it has)

piotrbartman:
> ... listing all of them. This sounds quite bad performance-wise (quadratic number of calls to whatever device backend it has)

This is not entirely true. The property: `devices`  searches all devices only if the assignment requires it (i.e. it included `*` as port_id). I guess it not a case here, so the last commit should help.

marmarek:
Ok, so remaining issue is about the unit tests, for example:
```
  File "templates/libvirt/xen.xml", line 160, in block 'devices'
    {% for device in assignment.devices %}
    ^^^^^^^^^^^^^^^^^^^^^^^^^
jinja2.exceptions.UndefinedError: 'qubes.device_protocol.DeviceAssignment object' has no attribute 'devices'
```

And there are also 2 minor pylint compains

fepitre:
Please note that from our previous diagnosis, mypy complains with your fix `*` for positional argument:
```
qubes/ext/utils.py:51: error: Too many positional arguments for "DeviceInfo"  [misc]
qubes/ext/utils.py:59: error: Too many positional arguments for "DeviceInfo"  [misc]
qubes/ext/utils.py:62: error: Too many positional arguments for "DeviceInfo"  [misc]
qubes/ext/utils.py:65: error: Too many positional arguments for "DeviceInfo"  [misc]
Found 4 errors in 1 file (checked 60 source files)
```

marmarek:
Note to self: check what happens when starting VM with PCI device that got removed

qubes/ext/utils.py
marmarek: I've tried and it works for me just fine without this part. I've tried by assigning block device with ask, and then connecting and disconnecting it - I got the prompt and when accepted the device got attached.

BTW, attaching device could use some log entry, especially when automatic or with ask (so you have some trail to which qube it got attached).

marmarek: Ok, I got once "Task was destroyed but it is pending". Skipping create task (`ask_response = await call_socket_service(...)`)helped.

piotrbartman: in my case skipping creating task always ended up with `Task was destroyed but it is pending`

marmarek (approved):

PR #615, opened 2024-08-21
tests: assertEquals -> assertEqual

marmarek:
The former is removed in Python 3.13

QubesOS/qubes-issues#9402


DemiMarie (approved):
LGTM!

PR #616, opened 2024-09-03
Kill a paused qube before system shutdown

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/8580

### DIFF
--- a/qubes/tests/__init__.py
+++ b/qubes/tests/__init__.py
@@ -1233,6 +1233,19 @@ class SystemTestCase(QubesTestCase):
             del vm
             self.fail("Timeout while waiting for VM {} shutdown".format(name))
 
+    def shutdown_paused(self, vm, timeout=60):
+        self.loop.run_until_complete(vm.pause())
+        with self.assertRaises(qubes.exc.QubesVMNotRunningError):
+            self.loop.run_until_complete(
+                vm.shutdown(wait=True, timeout=timeout, force=False))
+        try:
+            self.loop.run_until_complete(
+                vm.shutdown(wait=True, timeout=timeout, force=True))
+        except qubes.exc.QubesException:
+            name = vm.name
+            del vm
+            self.fail("Timeout while waiting for VM {} shutdown".format(name))
+
     def prepare_hvm_system_linux(self, vm, init_script, extra_files=None):
         if not os.path.exists('/usr/lib/grub/i386-pc'):
             self.skipTest('grub2 not installed')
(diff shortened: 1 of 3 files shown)


marmarek:
This doesn't look right. If a qube was paused, it was likely done for a reason, it shouldn't execute anything without _explicit_ user action. Maybe kill paused qubes on shutdown instead?

alimirjamali:
> Maybe kill paused qubes on shutdown instead?

This sounds more dangerous. What if the qube had open applications with proper auto-save on shutdown mechanism. Killing qube might lead to serious data-loss. Having said that, killing disposables might be a legitimate option.

marmarek:
Ideally, we'd ask the user, but I don't think we have any way to do that. The autosave is one scenario, but if user wanted to interact with such a qube, why it was paused? Another scenario could be some misbehaving qube paused to limit damage - unpausing it would be very wrong here.

alimirjamali:
So properly killing it would be the way. I will look into `libvirt_domain` to find the clean way to kill a paused domain. `libvirt_domain.shutdown()` fails to do it for suspended domains.

marmarek:
See what qvm-kill does. BTW, I think killing should be only during system shutdown. Normal qvm-shutdown should not automatically change to killing on a paused qube but maybe show an error or some other message?

alimirjamali:
@marmarek It is done. `qvm-pause` documentation is also updated to reflect this behaviour. Considering the fact that the existing behaviour is also indeed killing of paused qubes on shutdown, nothing is really changed here. It is only cleaner and much faster.

alimirjamali:
> Normal qvm-shutdown should not automatically change to killing on a paused qube but maybe show an error or some other message?

Ops. Just saw this. I will work on it.

alimirjamali:
Reason for amendment: Making Pylint happy(er)

alimirjamali:
It appears that CI/CD is broken at the moment.

I do not have access to PipelineRetry

alimirjamali:
OK. Codecov is advising on the necessary unittests. I have a crude idea on the required unittest. Might be wrong

alimirjamali:
@marmarek It appears that none of the basic integration unittests are executed. I am having trouble to find the proper unittest to update for this.

marmarek:
Integration tests are not running in gitlab-ci, they are running only in openqa, see https://www.qubes-os.org/doc/automated-tests/

marmarek (approved):

PR #623, opened 2024-10-13
Extend clipboard tests

marmarek:
QubesOS/qubes-issues#9296

### DIFF
--- a/qubes/tests/integ/basic.py
+++ b/qubes/tests/integ/basic.py
@@ -472,37 +472,35 @@ class TC_30_Gui_daemon(qubes.tests.SystemTestCase):
         super(TC_30_Gui_daemon, self).setUp()
         self.init_default_template()
 
-    @unittest.skipUnless(
-        spawn.find_executable('xdotool'),
-                         "xdotool not installed")
-    def test_000_clipboard(self):
-        testvm1 = self.app.add_new_vm(qubes.vm.appvm.AppVM,
-                                     name=self.make_vm_name('vm1'), label='red')
-        self.loop.run_until_complete(testvm1.create_on_disk())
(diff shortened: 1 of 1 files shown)


marmarek:
FYI @alimirjamali I extended clipboard tests to check few more sizes. I consider adding also one above 256k, but not sure if that's necessary. This should cover testing new version. Interaction of new gui-daemon with old gui-agent needs to be tested manually.
And also, those tests do not check if notification on failure is properly displayed.

alimirjamali:
> FYI @alimirjamali I extended clipboard tests to check few more sizes. I consider adding also one above 256k, but not sure if that's necessary. This should cover testing new version. Interaction of new gui-daemon with old gui-agent needs to be tested manually. And also, those tests do not check if notification on failure is properly displayed.

That is very nice. The test for above 256k will be necessary. Rejecting INCR is easy and could be easily added (after I could finish the existing PR).

alimirjamali:
There should be a unittest where inter-vm shared clipboard content is over 65000 (from a Protocol 1.8 aware VM) and user tries to paste it to a VM with protocol 1.7 with 65000 character limit. xside should handle this gracefully.

marmarek:
> There should be a unittest where inter-vm shared clipboard content is over 65000 (from a Protocol 1.8 aware VM) and user tries to paste it to a VM with protocol 1.7 with 65000 character limit. xside should handle this gracefully.

There seems to be no difference in the paste path between versions, so I expect the current test (200k) to cover this case already.

alimirjamali:
> There seems to be no difference in the paste path between versions, so I expect the current test (200k) to cover this case already.

OK. So no forced limit from the xside to vmside. Makes very sense. Better to keep it this way (now that I reconsider it).

PR #624, opened 2024-10-13
DEBUG: log list of VMs when startup failed

marmarek:
And also basic xen info (including free memory)


DemiMarie:
Should this be marked as draft?  I assume it is not meant for merging.

marmarek:
The actual fix is in #629

PR #626, opened 2024-10-25
UEFI test

DemiMarie:
UEFI VMs didn't boot.  Add a test to ensure they do.

### DIFF
--- a/qubes/tests/__init__.py
+++ b/qubes/tests/__init__.py
@@ -1328,33 +1328,60 @@ class SystemTestCase(QubesTestCase):
 
     def create_bootable_iso(self):
         """Create simple bootable ISO image.
-        Type 'poweroff' to it to terminate that VM.
+        Type 'halt' to it to terminate that VM.
         """
-        isolinux_cfg = (
-            'prompt 1\n'
-            'label poweroff\n'
-            '   kernel poweroff.c32\n'
-        )
         output_fd, output_path = tempfile.mkstemp('.iso')
         with tempfile.TemporaryDirectory() as tmp_dir:
+            f = os.open(tmp_dir, os.O_RDONLY|os.O_DIRECTORY)
             try:
-                shutil.copy('/usr/share/syslinux/isolinux.bin', tmp_dir)
-                shutil.copy('/usr/share/syslinux/ldlinux.c32', tmp_dir)
-                shutil.copy('/usr/share/syslinux/poweroff.c32', tmp_dir)
-                with open(os.path.join(tmp_dir, 'isolinux.cfg'), 'w') as cfg:
-                    cfg.write(isolinux_cfg)
-                subprocess.check_call(['genisoimage', '-o', output_path,
-                                       '-c', 'boot.cat',
-                                       '-b', 'isolinux.bin',
+                os.mkdir('os', mode=0o755, dir_fd=f)
+                os.mkdir('iso', mode=0o755, dir_fd=f)
+                os.mkdir('os/images', mode=0o755, dir_fd=f)
+                os.mkdir('os/EFI', mode=0o755, dir_fd=f)
+                os.mkdir('os/EFI/BOOT', mode=0o755, dir_fd=f)
+                with open(f'/proc/self/fd/{f}/os/images/boot_hybrid.img', 'wb') as v:
+                    subprocess.check_call(['grub2-mkimage',
+                                           '-O', 'i386-pc-eltorito',
+                                           '-d', '/usr/lib/grub/i386-pc',
+                                           '-p', '/boot/grub2',
+                                           'iso9660', 'biosdisk', 'halt'], stdout=v)
+                with open(f'/proc/self/fd/{f}/os/EFI/BOOT/BOOTX64.EFI', 'wb') as v:
+                    subprocess.check_call(['grub2-mkimage',
+                                           '-O', 'x86_64-efi',
+                                           '-d', '/usr/lib/grub/x86_64-efi',
+                                           '-p', '/boot/grub2',
+                                           'iso9660', 'disk', 'halt'], stdout=v)
+                graft_path = '.=' + os.path.join(tmp_dir, "os").replace("\\", "\\\\").replace("=", "\\=")
+                efiboot_img = os.path.join(tmp_dir, 'os/images/efiboot.img')
+                subprocess.check_call(['sudo', 'mkefiboot', '--label=ANACONDA',
+                                       os.path.join(tmp_dir, 'os/EFI/BOOT'),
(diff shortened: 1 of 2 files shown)


qubes/tests/integ/basic.py
```diff
@@ -105,15 +105,42 @@ def test_120_start_standalone_with_cdrom_dom0(self):
...
             self.loop.run_until_complete(asyncio.sleep(1))
         self.assertFalse(self.vm.is_running())
 
+    def test_121_start_uefi(self):
+        vmname = self.make_vm_name('appvm')
+        self.vm = self.app.add_new_vm('StandaloneVM', label='red', name=vmname)
+        self.loop.run_until_complete(self.vm.create_on_disk())
+        self.vm.kernel = None
+        self.vm.virt_mode = 'hvm'
+        subprocess.check_call(['qvm-features', '--', vmname, 'uefi', '1'])
```
marmarek, line 123:
```suggestion
        self.vm.features["uefi"] = "1"
```
?

But also, have you tried if that works? Because the check_call here would deadlock...

qubes/tests/__init__.py
```diff
@@ -1328,33 +1328,54 @@ def prepare_hvm_system_linux(self, vm, init_script, extra_files=None):
...
-                    cfg.write(isolinux_cfg)
-                subprocess.check_call(['genisoimage', '-o', output_path,
-                                       '-c', 'boot.cat',
-                                       '-b', 'isolinux.bin',
+                os.mkdir('os', mode=0o755, dir_fd=f)
+                os.mkdir('iso', mode=0o755, dir_fd=f)
+                os.mkdir('os/images', mode=0o755, dir_fd=f)
+                os.mkdir('os/EFI', mode=0o755, dir_fd=f)
+                os.mkdir('os/EFI/BOOT', mode=0o755, dir_fd=f)
+                subprocess.check_call(['sudo', 'grub2-mkimage',
```
marmarek, line 1342: Does it really need sudo? It shouldn't be needed if you write to the tmp_dir already. Which BTW is also missing here - according to the docs, default is stdout, and it isn't redirected here...

qubes/tests/__init__.py
```diff
@@ -1328,33 +1328,54 @@ def prepare_hvm_system_linux(self, vm, init_script, extra_files=None):
...
+                                       '-O', 'i386-pc-eltorito',
+                                       '-d', '/usr/lib/grub/i386-pc',
+                                       '-p', '/boot/grub2',
+                                       'iso9660', 'biosdisk', 'halt'])
+                graft_path = '.=' + os.path.join(tmp_dir, "os").replace("\\", "\\\\").replace("=", "\\=")
+                efiboot_img = os.path.join(tmp_dir, 'os/images/efiboot.img')
+                with open('os/EFI/BOOT', mode=0o644, dir_fd=f) as g:
+                    subprocess.check_call(['sudo', 'cat', '/boot/efi/EFI/qubes/grubx64.efi'], stdout=g)
```
marmarek, line 1350: This file won't exists on a system installed in a legacy boot mode (like most instances in openQA). But also, do you really need it? A `grub2-mkimage` call similar to the above should give you what you need, no?

marmarek (changes requested):

qubes/tests/__init__.py
DemiMarie: I’ll need to figure it out (later).

qubes/tests/integ/basic.py
DemiMarie: I have not tested this at all (beyond “it has no syntax errors”), because I don’t have a machine to test it on.

marmarek:
The test failed for several reasons:
1. I forgot to tag https://github.com/QubesOS/qubes-vmm-xen/pull/195 for this run.
2. Missing packages (lorax for mkefiboot, grub2-pc-modules (on UEFI system)).
3. https://github.com/QubesOS/qubes-core-admin/pull/631

After fixing all the above, it still fails - the VM starts, but it has only black window and no grub. Doesn't react for typing "halt" - and that's what makes the test fail (VM doesn't shutdown when it should).

DemiMarie:
> The test failed for several reasons:
> 
> 1. I forgot to tag [Depend on edk2-ovmf-xen instead of bundling OVMF qubes-vmm-xen#195](https://github.com/QubesOS/qubes-vmm-xen/pull/195) for this run.
> 2. Missing packages (lorax for mkefiboot, grub2-pc-modules (on UEFI system)).

Is this something that should be handled by this PR, or is it the responsibility of the test runner?

> 3. [Fix OVMF path in libvirt xml config #631](https://github.com/QubesOS/qubes-core-admin/pull/631)
> 
> After fixing all the above, it still fails - the VM starts, but it has only black window and no grub. Doesn't react for typing "halt" - and that's what makes the test fail (VM doesn't shutdown when it should).

That is strange.  I did a test that should be equivalent and that test does pass:

1. Use the Python code to create an ISO image.
2. Boot a VM from the ISO.
3. Type "halt".

This was on R4.2, though.  Can you recheck on R4.2 and see if the test passes there?  If it passes on R4.2 I suspect a regression in R4.3.

marmarek:
> Is this something that should be handled by this PR, or is it the responsibility of the test runner?

Nothing to do here.

> This was on R4.2, though. Can you recheck on R4.2 and see if the test passes there? If it passes on R4.2 I suspect a regression in R4.3.

I can try

marmarek:
Ok, I think I know what the issue is. Or rather, two issues:
1. Grub UI doesn't show up in the VM window, ever.
2. UEFI VM takes longer to start, so "halt" may be entered too early.

The first one is why I haven't noticed it actually worked eventually. The second one is the reason why test fails often (but it does passes from time to time. And when it fails (and with `QUBES_TEST_WAIT_ON_FAIL=1`) entering "halt" later does make the VM to shutdown.

DemiMarie:
> 1. Grub UI doesn't show up in the VM window, ever.

This is actually a regression compared to my (patched) R4.2, where it does show up provided one puts the UEFI firmware file where Xen will find it.

> 2. UEFI VM takes longer to start, so "halt" may be entered too early.

Should I work around this by increasing the timeout?

marmarek:
Interestingly, it does show up on a VM console (click any of the failures in in the bot comment, go to logs tab and see the guest-test-inst-appvm.log). Maybe the test could check for that? Sounds more reliable check if it worked than checking if typing "halt" did something (that could be also VM crashing just after startup...). This should also help with typing "halt" too early.

Technically, it could open the log after starting the VM see to the end and look back for grub messages until last "Logfile Opened" line (and repeat the search with some timeout).

PR #629, opened 2024-10-27
Fix locking in qmemman on Python 3.13

marmarek:
See commit messages for details.

Fixes QubesOS/qubes-issues#9431


DemiMarie:
Is this bug actually specific to Python 3.13+, or just much easier to trigger there?  To me this change ought to be backported.

marmarek:
> To me this change ought to be backported.

Yes, this may be backported (once it really lands in "main").

PR #630, opened 2024-10-29
Adjust qubesd.service ordering once more

marmarek:
Order it after lvm2-activation.service, so that all present groups are
already activated (and its list can be cached).

### DIFF
--- a/qubes/storage/lvm.py
+++ b/qubes/storage/lvm.py
@@ -221,14 +221,17 @@ def _parse_lvm_cache(lvm_output):
             size, usage_percent = row['lv_size'], row['data_percent']
         except KeyError:
             continue
-        if '' in [pool_name, name, size, usage_percent]:
+        if '' in [pool_name, name, size]:
             continue
         pool_lv, attr, origin = row['pool_lv'], row['lv_attr'], row['origin']
         metadata_size = row['lv_metadata_size']
         metadata_percent = row['metadata_percent']
         name = pool_name + "/" + name
         size = int(size[:-1])  # Remove 'B' suffix
-        usage = int(size / 100 * float(usage_percent))
+        if usage_percent:
+            usage = int(size / 100 * float(usage_percent))
+        else:
+            usage = 0
         if metadata_size:
             metadata_size = int(metadata_size[:-1])
             metadata_usage = int(metadata_size / 100 * float(metadata_percent))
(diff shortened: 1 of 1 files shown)


marmarek:
This won't work, as LVM activation is event based (doesn't use lvm2-activation-generator anymore). The event-generated unit looks like this:
```
# /run/systemd/transient/lvm-activate-qubes_dom0.service
# This is a transient unit file, created programmatically via the systemd API. Do not edit.
[Unit]
Description=/usr/sbin/lvm vgchange -aay --autoactivation event qubes_dom0
DefaultDependencies=no
[Service]
ExecStart=
ExecStart="/usr/sbin/lvm" "vgchange" "-aay" "--autoactivation" "event" "qubes_dom0"
# /usr/lib/systemd/system/service.d/10-timeout-abort.conf
# This file is part of the systemd package.
# See https://fedoraproject.org/wiki/Changes/Shorter_Shutdown_Timer.
#
# To facilitate debugging when a service fails to stop cleanly, 
# TimeoutStopFailureMode=abort is set to "crash" services that fail to stop in 
# the time allotted. This will cause the service to be terminated with SIGABRT 
# and a coredump to be generated.
#
# To undo this configuration change, create a mask file:
#   sudo mkdir -p /etc/systemd/system/service.d
#   sudo ln -sv /dev/null /etc/systemd/system/service.d/10-timeout-abort.conf
[Service]
TimeoutStopFailureMode=abort
```
Note there is no common `Before=` point to order. And it even has `DefaultDependencies=no`. It looks like systemd ordering can't solve this problem, we'll need a solution inside qubesd :/

DemiMarie:
I think qubesd needs to not assume that every volume in a pool is active.  That’s a good idea anyway for performance reasons.  The current assumption is technical debt that needs to be paid off.

marmarek:
I don't consider it a technical debt, there are many benefits of keeping volumes activated all the time. In my experiments the major disadvantage of activating on demand is inability to check its size/usage (LVM does not show this info for inactive volumes). But also, in my experiment with activating on demand (#382 ) it turns out there is _a lot more_ corner cases and potential for various race conditions (since you can't use /dev/ file existence anymore, and calling lvm each time is _slow_, you need to rely on the cache much more). And at the same time, it's very unclear if there are any benefits from this approach (some rough tests shows it isn't significantly faster, if at all) - it looks just like extra complexity...

DemiMarie:
> And at the same time, it's very unclear if there are any benefits from this approach (some rough tests shows it isn't significantly faster, if at all) - it looks just like extra complexity...

How many volumes were active in these tests?

marmarek:
AFAIR I did that on a test system with 100+ VMs

PR #632, opened 2024-11-02
Hide local timezone from VMs with specific tag

strategictraveler:
Set fake timezone for VMs with the "anon-timezone" tag. Fixes QubesOS/qubes-issues#8381


marmarek:
Looks good but please add a test for this, see https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/tests/vm/qubesvm.py, especially look at some with "qdb" in name.

strategictraveler:
I'll look at it this weekend.

adrelanos:
Is this still needed?

marmarek:
Superseded by https://github.com/QubesOS/qubes-core-admin/pull/646

PR #633, opened 2024-11-09
Improve `qvm-prefs -D SOMEVM template` error message

alimirjamali:
resolves: https://github.com/QubesOS/qubes-issues/issues/8928

### DIFF
--- a/qubes/vm/appvm.py
+++ b/qubes/vm/appvm.py
@@ -120,6 +120,11 @@ class AppVM(qubes.vm.mix.dvmtemplate.DVMTemplateMixin,
     def on_property_pre_reset_template(self, event, name, oldvalue=None):
         '''Forbid deleting template of running VM
         '''  # pylint: disable=unused-argument
+        default_template = getattr(self.app, 'default_template', None)
+        if default_template is not None:
+            raise qubes.exc.QubesValueError('Cannot unset template. ' \
+                'You can set it to the current default ' \
+                f'({default_template.name}) instead')
         raise qubes.exc.QubesValueError('Cannot unset template')
 
     @qubes.events.handler('property-pre-set:template')
--- a/qubes/tests/vm/appvm.py
+++ b/qubes/tests/vm/appvm.py
@@ -147,6 +147,10 @@ class TC_90_AppVM(qubes.tests.vm.qubesvm.QubesVMTestsMixin,
         vm = self.get_vm()
         with self.assertRaises(qubes.exc.QubesValueError):
             vm.template = qubes.property.DEFAULT
+        self.app.default_template = self.template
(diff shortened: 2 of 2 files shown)


qubes/vm/appvm.py
```diff
@@ -120,7 +120,9 @@ def on_domain_loaded(self, event):
     def on_property_pre_reset_template(self, event, name, oldvalue=None):
         '''Forbid deleting template of running VM
         '''  # pylint: disable=unused-argument
-        raise qubes.exc.QubesValueError('Cannot unset template')
+        default_template = getattr(self.app, 'default_template')
```
marmarta, line 123: You have to have a default value here, otherwise this will raise an exception if attribute is not found (and tests fail on this).

qubes/vm/dispvm.py
```diff
@@ -163,7 +163,9 @@ def on_domain_loaded(self, event):
     def on_property_pre_reset_template(self, event, name, oldvalue=None):
         '''Forbid deleting template of VM
         '''  # pylint: disable=unused-argument
-        raise qubes.exc.QubesValueError('Cannot unset template')
+        default_template = getattr(self.app, 'default_template')
```
marmarta, line 166: as above

marmarta (changes requested):

qubes/vm/appvm.py
alimirjamali: It is possible to have the `default_template` unset? I did not know that.

marmarta (approved):

qubes/vm/appvm.py
```diff
@@ -120,6 +120,11 @@ def on_domain_loaded(self, event):
     def on_property_pre_reset_template(self, event, name, oldvalue=None):
         '''Forbid deleting template of running VM
         '''  # pylint: disable=unused-argument
+        default_template = getattr(self.app, 'default_template', None)
+        if default_template is not None:
+            raise qubes.exc.QubesValueError('Cannot unset template. '\
+                f'Use `qvm-prefs {self.name} template {default_template.name}`'\
```
marmarek, line 126: Is  naming qvm-prefs specifically here a good idea? That error may show up in a bunch of places (not necessarily in CLI), and user may prefer to use GUI. Maybe just say "You can set it to the current default (name) instead" ?

alimirjamali: I thought about the exact scenario. And every existing GUI tool I know uses some sort of ComboBox with predefined choices of non-blank templates. Unless there is one that I do not know. Still your choice.

alimirjamali: > Maybe just say "You can set it to the current default (name) instead" ?

It is done.

marmarek (approved):

PR #636, opened 2024-11-20
Reformat code with python-black and add CI job

fepitre:


marmarek:
> @fepitre Can you provide a script that reproduces your changes?

It's literally in the commit title: `black`

marmarek:
That said, the call seems to be missing `-l 80` ...

qubes/storage/callback.py
```diff
@@ -509,17 +509,17 @@ def __init__(self, pool, impl):
         self._cb_impl = impl  #: Backend volume implementation instance.
 
     async def _assert_initialized(self, **kwargs):
-        await self._cb_pool._assert_initialized(
+        await self._cb_pool._assert_initialized(  # pylint: disable=protected-access
```
marmarek, line 512: shouldn't this be before the call then?

qubes/storage/callback.py
```diff
@@ -509,17 +509,17 @@ def __init__(self, pool, impl):
...
             **kwargs
-        )  # pylint: disable=protected-access
+        )
 
     async def _callback(self, cb, cb_args=None, **kwargs):
         if cb_args is None:
             cb_args = []
         vol_args = [self.name, self.vid, self.source, *cb_args]
-        await self._cb_pool._callback(
+        await self._cb_pool._callback(  # pylint: disable=protected-access
```
marmarek, line 520: and same here

qubes/storage/reflink.py
```diff
@@ -227,9 +227,9 @@ def verify(self):  # pylint: disable=invalid-overridden-method
     @qubes.storage.Volume.locked
     @_coroutinized
     def remove(self):  # pylint: disable=invalid-overridden-method
-        self.pool._volumes.pop(
+        self.pool._volumes.pop(  # pylint: disable=protected-access
```
marmarek, line 230: and here

marmarek (changes requested):

qubes/storage/callback.py
fepitre: I can try before but at this place it works too. Let me try.

DemiMarie:
> > @fepitre Can you provide a script that reproduces your changes?
> 
> It's literally in the commit title: `black`

Which version and with which arguments?

marmarek (approved):
`black -l 80 qubes` gives the same result as 853531e2cf0bcf316c3db074a1d39bd6df346e11 (with except to different wrapping of two docstrings for some reason); other two commits checked manually

PR #639, opened 2024-11-30
Various test improvement, especially for testing minimal templates

marmarek:


qubes/tests/integ/audio.py
```diff
@@ -338,7 +338,7 @@ def _configure_audio_recording(self, vm):
             str(source_index),
         ]
         if audiovm.name != "dom0":
-            self.loop.run_until_complete(audiovm.run(" ".join(cmd)))
+            self.loop.run_until_complete(audiovm.run_for_stdio(" ".join(cmd)))
```
DemiMarie, line 341: Should this escape each element?

marmarek: Not necessary, the command is built literally few lines above and doesn't contain spaces (and if it would, it _should_ explode)

PR #642, opened 2024-12-23
Use the UUID for the machine ID

DemiMarie:
This avoids leaking the real machine ID into guests, and provides an easy way for guests to know their own UUID.

Fixes: QubesOS/qubes-issues#8833

### DIFF
--- a/qubes/vm/qubesvm.py
+++ b/qubes/vm/qubesvm.py
@@ -2533,6 +2533,10 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.BaseVM):
         """
         if not self.kernel:
             return ""
+        if self.features.check_with_template("os", None) == "Linux":
+            base_kernelopts = "systemd.machine_id=" + self.uuid.hex + " "
+        else:
+            base_kernelopts = ""
         kernels_dir = self.storage.kernels_dir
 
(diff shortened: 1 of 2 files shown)


qubes/vm/qubesvm.py
```diff
@@ -2533,16 +2533,17 @@ def kernelopts_common(self):
         """
         if not self.kernel:
             return ""
+        base_kernelopts = "systemd.machine_id=" + self.uuid.hex + " "
```
marmarek, line 2536: Can you at least guard it with `if self.features.check_with_template("os") == "Linux"`? And/or make it possible to disable some other way?

marmarek (changes requested):
And also, tests need updating for this change.

qubes/vm/qubesvm.py
DemiMarie: I’ll do the first for now, and figure out the rest once Qubes OS supports a distro (like Alpine) that is not systemd-based.

marmarek: AFAIK there is a community Alpine template

DemiMarie: Fair, but let’s get this merged first for Whonix.

marmarek (approved):

PR #643, opened 2024-12-29
Solves #9674

piotrbartman:
Delete the file and drop-in folder for the denied device interfaces list and move the list to `qvm-prefs`.

Solves QubesOS/qubes-issues/issues/9674


qubes/api/admin.py
```diff
@@ -1630,6 +1631,74 @@ async def vm_device_set_required(self, endpoint, untrusted_payload):
...
+    @qubes.api.method(
+        "admin.vm.device.denied.List",
+        no_payload=True, scope="local", read=True)
+    async def vm_device_denied_list(self):
+        """
+        List all denied device interfaces for the VM.
+
+        Returns a newline-separated string.
+        """
+        self.enforce(not self.arg)
```
DemiMarie, line 1643: This will result in `Got empty response from qubesd`, which indicates an internal error.  Please throw an exception that will be returned to the user of the Admin API.

qubes/api/admin.py
```diff
@@ -1630,6 +1631,74 @@ async def vm_device_set_required(self, endpoint, untrusted_payload):
...
+
+        self.fire_event_for_permission()
+
+        denied = self.dest.devices_denied
+        return "\n".join(map(repr, DeviceInterface.from_str_bulk(denied)))
```
DemiMarie, line 1648: Maybe `assert` that there are no newlines in the `repr` output?

qubes/device_protocol.py
```diff
@@ -733,6 +733,14 @@ def unknown(cls) -> "DeviceInterface":
         """Value for unknown device interface."""
         return cls("?******")
 
+    @staticmethod
+    def from_str_bulk(interfaces: Optional[str]) -> List["DeviceInterface"]:
+        interfaces = interfaces or []
+        return [
+            DeviceInterface(interfaces[i: i + 7])
+            for i in range(0, len(interfaces), 7)
+        ]
```
DemiMarie, line 742: I suggest raising a `QubesValueError` if the size is not a multiple of 8 or if there are invalid characters in the interface name.

qubes/device_protocol.py
```diff
@@ -733,6 +733,14 @@ def unknown(cls) -> "DeviceInterface":
         """Value for unknown device interface."""
         return cls("?******")
 
+    @staticmethod
+    def from_str_bulk(interfaces: Optional[str]) -> List["DeviceInterface"]:
+        interfaces = interfaces or []
```
DemiMarie, line 738:
```suggestion
        interfaces = interfaces or ""
```

Fix typing error.

DemiMarie (changes requested):

qubes/api/admin.py
marmarek: It's okay the way it is - the client part shouldn't put the argument there.

qubes/api/admin.py
```diff
@@ -1630,6 +1631,74 @@ async def vm_device_set_required(self, endpoint, untrusted_payload):
...
+
+    @qubes.api.method(
+        "admin.vm.device.denied.Remove", scope="local", write=True)
+    async def vm_device_denied_remove(self, untrusted_payload):
+        """
+        Remove device interface(s) from the denied list for the VM.
+
+        Payload:
+            Encoded device interface (can be repeated without any separator).
+            If payload is empty, all interfaces are removed.
```
marmarek, line 1678: This invites mistakes... maybe better some magic value like `all`?

qubes/api/admin.py
```diff
@@ -1630,6 +1631,74 @@ async def vm_device_set_required(self, endpoint, untrusted_payload):
...
+        return "\n".join(map(repr, DeviceInterface.from_str_bulk(denied)))
+
+    @qubes.api.method(
+        "admin.vm.device.denied.Add", scope="local", write=True)
+    async def vm_device_denied_add(self, untrusted_payload):
+        """
+        Add device interface(s) to the denied list for the VM.
+
+        Payload:
+            Encoded device interface (can be repeated without any separator).
```
marmarek, line 1657: What is intended behavior on duplicates? IIUC right now if at least one interface you try to add is already on deny list, the whole operation will fail. Is that intentional? IMHO better would be to ignore duplicates (as in - adding interface that is already denied is no-op, but others non-duplicated added at the same time are still added). Excluding duplicates at the client side is racy (somebody can modify the value between client reads it (to exclude duplicates) and writes it back).

marmarek (changes requested):
Can the "cleaning" commit be squashed into original adding commit?

qubes/api/admin.py
piotrbartman: Yes, this was intentional; I followed the "fail fast" principle. However, now I understand that in the mentioned example, admin methods should be idempotent. 
To keep consistency, in the case of removing an interface that is not currently on the list nothing should be done, am I right?

marmarek: Yes, I think so

qubes/device_protocol.py
```diff
@@ -733,6 +738,19 @@ def unknown(cls) -> "DeviceInterface":
         """Value for unknown device interface."""
         return cls("?******")
 
+    @staticmethod
+    def from_str_bulk(interfaces: Optional[str]) -> List["DeviceInterface"]:
+        interfaces = interfaces or ""
+        if len(interfaces) % 7 != 0:
+            QubesValueError(
```
marmarek, line 745: pylint says: missing `raise`

marmarek (changes requested):
Generally looks fine, besides the issue pylint found. And black is not happy...

marmarek:
As previously - can you fold fixes (especially the exception one) into relevant commits?

marmarek (approved):

PR #645, opened 2025-01-01
Better error messages

DemiMarie:
Use better exceptions for various client errors.

This is DRAFT and not ready yet for many reasons.

### DIFF
--- a/qubes/api/__init__.py
+++ b/qubes/api/__init__.py
@@ -33,7 +33,11 @@ from typing import Union, Any
 import uuid
 
 import qubes.exc
-from qubes.exc import ProtocolError, PermissionDenied
+from qubes.exc import (
+    ProtocolError,
+    PermissionDenied,
+    QubesArgumentNotAllowedError,
+)
 
 
 def method(name, *, no_payload=False, endpoints=None, **classifiers):
@@ -238,6 +242,11 @@ class AbstractQubesAPI:
     def fire_event_for_filter(self, iterable, **kwargs):
         """Fire an event on the source qube to filter for permission"""
         return apply_filters(iterable, self.fire_event_for_permission(**kwargs))
+
+    def enforce_no_arg(self) -> None:
+        """If the argument is not empty, raise an exception."""
+        if self.arg:
+            raise QubesArgumentNotAllowedError(self.method, self.arg)
 
     @staticmethod
     def enforce(predicate):
--- a/qubes/api/__init__.py
+++ b/qubes/api/__init__.py
@@ -37,6 +37,7 @@ from qubes.exc import (
     ProtocolError,
     PermissionDenied,
     QubesArgumentNotAllowedError,
+    DestinationNotDom0Error,
 )
 
 
@@ -247,6 +248,12 @@ class AbstractQubesAPI:
         """If the argument is not empty, raise an exception."""
         if self.arg:
             raise QubesArgumentNotAllowedError(self.method, self.arg)
+
+    def enforce_dest_dom0(self) -> None:
+        """If the destination is not dom0, raise an exception."""
+        name = self.dest.name
+        if name != "dom0":
+            raise DestinationNotDom0Error(name)
 
     @staticmethod
     def enforce(predicate):
--- a/qubes/api/__init__.py
+++ b/qubes/api/__init__.py
@@ -170,10 +170,10 @@ class AbstractQubesAPI:
         self.dest = decode_vm(dest, app.domains)
 
         #: argument
-        self.arg = arg.decode("ascii")
+        self.arg = arg.decode("ascii", "strict")
 
         #: name of the method
-        self.method = method_name.decode("ascii")
+        self.method = method_name.decode("ascii", "strict")
 
         #: callback for sending events if applicable
         self.send_event = send_event
@@ -365,6 +365,11 @@ class QubesDaemonProtocol(asyncio.Protocol):
                 dest,
                 len(untrusted_payload),
             )
+            if self.transport is not None:
+                self.send_exception(err)
+                self.transport.write_eof()
+                self.transport.close()
+            return
 
         except ProtocolError:
             self.app.log.warning(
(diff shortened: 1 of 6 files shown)


qubes/api/__init__.py
```diff
@@ -165,10 +170,10 @@ def __init__(
         self.dest = decode_vm(dest, app.domains)
 
         #: argument
-        self.arg = arg.decode("ascii")
+        self.arg = arg.decode("ascii", "strict")
```
ben-grande, line 173: This is already the default:

- https://docs.python.org/3/library/stdtypes.html#bytes.decode

Was this intended to be explicit, enforce against upstream change of defaults?

qubes/api/__init__.py
```diff
@@ -239,6 +244,17 @@ def fire_event_for_filter(self, iterable, **kwargs):
...
+    def enforce_no_arg(self) -> None:
+        """If the argument is not empty, raise an exception."""
+        if self.arg:
+            raise QubesArgumentNotAllowedError(self.method, self.arg)
+
+    def enforce_dest_dom0(self) -> None:
+        """If the destination is not dom0, raise an exception."""
+        name = self.dest.name
+        if name != "dom0":
+            raise DestinationNotDom0Error(name)
```
ben-grande, line 256: I didn't make special exceptions for target error. It is using `ProtocolError("Destination is not dom0")` for example. I think it is fine. Same thing for the code above regarding prohibiting the use of argument.

qubes/api/__init__.py
```diff
@@ -349,6 +365,11 @@ async def respond(self, src, meth, dest, arg, *, untrusted_payload):
                 dest,
                 len(untrusted_payload),
             )
+            if self.transport is not None:
+                self.send_exception(err)
```
ben-grande, line 369: This was added.

qubes/api/admin.py
```diff
@@ -431,13 +439,15 @@ async def vm_volume_listsnapshots(self):
 
     @qubes.api.method("admin.vm.volume.Revert", scope="local", write=True)
     async def vm_volume_revert(self, untrusted_payload):
-        self.enforce(self.arg in self.dest.volumes.keys())
+        self._check_volume()
+        self.enforce(_snapshot_re.match(untrusted_payload))
```
ben-grande, line 443: Copying the regex here so it makes it easier to read:

```python3
_snapshot_re = re.compile(rb"\A[A-Za-z0-9][0-9A-Za-z_.-]*\n?\Z")
```

The `revision` is in the `payload`. Later on, it is checked if the `untrusted_revision` is in the `snapshots`. I believe the regex you set was to allow a sanitized exception message. Although I am not logging the revision name, the check for correct revision in safe set was added:

```python3
self.enforce_arg(
    wants=self.dest.volumes.keys(),
    short_reason=self.EXC_ARG_NOT_IN_DEST_VOLUMES,
)
untrusted_revision = untrusted_payload.decode("ascii").strip()
del untrusted_payload
allowed_chars = string.ascii_letters + string.digits + "-_."
self.enforce(
    all(c in allowed_chars for c in untrusted_revision),
    reason="Revision name must be in safe set: " + allowed_chars,
)
revision = untrusted_revision
del untrusted_revision
```

qubes/api/admin.py
```diff
@@ -1228,10 +1238,13 @@ async def _vm_create(
...
-                    untrusted_volume
-                    in ["root", "private", "volatile", "kernel"]
-                )
+                if untrusted_volume not in [
+                    "root",
+                    "private",
+                    "volatile",
+                    "kernel",
+                ]:
+                    raise qubes.exc.UnknownVolumeError("<untrusted>")
```
ben-grande, line 1247: This is also done, but specifying the volumes that it is not, rather than what volume it is.

qubes/api/admin.py
```diff
@@ -1621,7 +1634,7 @@ async def vm_device_set_required(self, endpoint, untrusted_payload):
         try:
             mode = allowed_values[untrusted_payload]
         except KeyError:
-            raise qubes.exc.PermissionDenied()
+            raise qubes.exc.QubesAttachmentKindError("invalid attachment kind")
```
ben-grande, line 1637: Changed to `ProtocolError("Invalid assignment mode")`. Wouldn't hurt to add an specific exception. TODO.

Not related to this PR, but the docstring does not match what `allowed_values` expect of `untrusted_payload`.

qubes/api/admin.py
```diff
@@ -1792,19 +1805,19 @@ def _backup_progress_callback(self, profile_name, progress):
...
         self.fire_event_for_permission()
 
         profile_path = os.path.join(
             qubes.config.backup_profile_dir, self.arg + ".conf"
         )
         if not os.path.exists(profile_path):
-            raise qubes.exc.PermissionDenied(
-                "Backup profile {} does not exist".format(self.arg)
-            )
+            raise BackupProfileNotFoundError(self.arg)
```
ben-grande, line 1820: Wouldn't hurt, good idea, TODO.

qubes/api/admin.py
```diff
@@ -1849,19 +1862,19 @@ async def backup_cancel(self):
...
```
ben-grande, line 1877: Same, add, TODO.

qubes/api/admin.py
```diff
@@ -431,13 +439,15 @@ async def vm_volume_listsnapshots(self):
...
+        self._check_volume()
+        self.enforce(_snapshot_re.match(untrusted_payload))
         untrusted_revision = untrusted_payload.decode("ascii").strip()
         del untrusted_payload
 
         volume = self.dest.volumes[self.arg]
         snapshots = volume.revisions
-        self.enforce(untrusted_revision in snapshots)
+        if untrusted_revision not in snapshots:
+            raise qubes.exc.QubesNoSuchSnapshotError(untrusted_revision)
```
ben-grande, line 450: Done.

PR #646, opened 2025-01-02
Do not set timezone for VMs with anon-timezone feature

DemiMarie:
This is implemented as a feature so that the standard
check-with-template mechanism can be used.  VMs can request
the feature using the standard mechanism.
 
Fixes: QubesOS/qubes-issues#8381

### DIFF
--- a/qubes/tests/vm/qubesvm.py
+++ b/qubes/tests/vm/qubesvm.py
@@ -2160,31 +2160,35 @@ class TC_90_QubesVM(QubesVMTestsMixin, qubes.tests.QubesTestCase):
             "-A FORWARD -i vif+ -o vif+ -j DROP\n"
             "COMMIT\n".format(datetime.datetime.now().ctime())
         )
+        data = {
+            "/name": "test-inst-test",
+            "/type": "StandaloneVM",
+            "/default-user": "user",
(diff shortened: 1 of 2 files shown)


marmarek (approved):
this looks okay, but needs a documentation entry in qvm-features man page

DemiMarie:
> this looks okay, but needs a documentation entry in qvm-features man page

It actually had a bug: the feature was there, but VMs could not request it themselves, as Whonix will want.

marmarek:
> It actually had a bug: the feature was there, but VMs could not request it themselves, as Whonix will want.

No, I don't think VM should be able to request it itself. Especially the way you did it will allow VM to request to _disable_ it too...
Setting this feature for Whonix is a task for https://github.com/QubesOS/qubes-core-admin-addon-whonix/

DemiMarie:
> > It actually had a bug: the feature was there, but VMs could not request it themselves, as Whonix will want.
> 
> No, I don't think VM should be able to request it itself. Especially the way you did it will allow VM to request to _disable_ it too... Setting this feature for Whonix is a task for https://github.com/QubesOS/qubes-core-admin-addon-whonix/

I rolled back to the previous commit without that change.

PR #647, opened 2025-01-21
tests: add qrexec performance tests

marmarek:
Add simple connection latency, and throughput tests. Run them with
different type of services (scripts, socket, via fork-server or not).
They print a test run time for comparison - the lower the better.

QubesOS/qubes-issues#5740

This will be especially useful to measure impact of https://github.com/QubesOS/qubes-core-qrexec/pull/141 and similar changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -235,6 +235,8 @@ endif
 	mkdir -p "$(DESTDIR)$(FILESDIR)"
 	cp -r templates "$(DESTDIR)$(FILESDIR)/templates"
 	cp -r tests-data "$(DESTDIR)$(FILESDIR)/tests-data"
+	mkdir -p "$(DESTDIR)/usr/lib/qubes"
+	cp -r tests "$(DESTDIR)/usr/lib/qubes/"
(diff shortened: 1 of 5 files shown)


DemiMarie:
I think it would be useful to provide a standalone script that users can run on their own machines, without having to stop qubesd or do something similarly disruptive.  The integration tests would then call the script with the appropriate arguments.

marmarek:
And what that script would do? Just call call qrexec in a loop? That's oneliner you can simply copy from the test to shell. OTOH, if you'd expect it to create VMs, setup policy etc separately, that would duplicate several parts that our test classes already do (and probably have different bugs while at it).

marmarek:
Thinking more about it, maybe I can make a script that takes already created (and running?) VMs as an argument.

DemiMarie:
> Thinking more about it, maybe I can make a script that takes already created (and running?) VMs as an argument.

I think this is a good idea.  While users could of course make their own benchmarks, having a single benchmark script is much better for reproducibility.

marmarek:
So, the script is there. It needed a bit of boilerplate compared to just running test commands directly, but well...

PR #649, opened 2025-01-26
tests: add simple storage performance tests

marmarek:
Add few simple storage performance tests using fio tool. Tests are checking
dom0's root and varlibqubes pools (by default the same thing, but in case
of XFS or BTRFS setups, they are different). And VM's
root/private/volatile. The last one is tested by creating ext4 filesystem
there first. This isn't very representative
(normally volatile is used for swap and CoW data), but allows comparing 
results with other volumes.

The tests can be also started outside of the full test run by calling
/usr/lib/qubes/tests/storage_perf.py. It requires giving name of a VM to 
test (which may be dom0).

QubesOS/qubes-issues#5740

This PR is on top of #647


tlaurion:
Aims to iterate in goal of fixing https://github.com/QubesOS/qubes-issues/issues/7300

PR #653, opened 2025-02-13
Add boot mode support

ArrayBolt3:
This PR is intended to add support for boot modes as specified here: https://github.com/QubesOS/qubes-issues/issues/9750#issuecomment-2655149125 This naturally only handles the backend part of things, the frontend will need its own PR against qubes-manager.

This is incomplete - I have not yet actually added the part that listens for qvm-request-features calls from TemplateVMs and sets boot mode features appropriately.

This implements (or rather, will implement) the backend component needed to implement https://github.com/QubesOS/qubes-issues/issues/9750.

### DIFF
--- a/qubes/ext/core_features.py
+++ b/qubes/ext/core_features.py
@@ -104,6 +104,86 @@ class CoreFeatures(qubes.ext.Extension):
             untrusted_value = untrusted_features["qubes-agent-version"]
             if _version_re.fullmatch(untrusted_value):
                 vm.features["qubes-agent-version"] = untrusted_value
+
+        # handle boot mode advertisement
+        old_bootmode_info = {}
+        for feature_key, feature_val in vm.features.items():
+            if feature_key.startswith(
+                "boot-mode.kernelopts."
+            ) or feature_key.startswith("boot-mode.name."):
+                old_bootmode_info[feature_key] = feature_val
+        new_bootmode_info = {}
+        new_bootmode_names = []
+        for (
+            untrusted_feature_key,
+            untrusted_feature_value,
+        ) in untrusted_features.items():
+            if untrusted_feature_key.startswith("boot-mode.kernelopts."):
+                bootmode_key_parts = untrusted_feature_key.split(".")
+                if len(bootmode_key_parts) != 3:
+                    # Boot mode key contains unexpected data, reject it
+                    continue
+                bootmode_name = bootmode_key_parts[2]
+                if bootmode_name == "":
+                    continue
+                if bootmode_name == "default":
+                    # "default" is reserved, cannot set kernelopts for it
+                    continue
+                bootmode_feature = untrusted_feature_key
+                bootmode_value = untrusted_feature_value
+                new_bootmode_info[bootmode_feature] = bootmode_value
+        for (
+            untrusted_feature_key,
+            untrusted_feature_value,
+        ) in untrusted_features.items():
+            if untrusted_feature_key.startswith("boot-mode.name."):
+                bootmode_key_parts = untrusted_feature_key.split(".")
+                if len(bootmode_key_parts) != 3:
+                    # Boot mode key contains unexpected data, reject it
+                    continue
+                bootmode_name = bootmode_key_parts[2]
+                if bootmode_name == "":
+                    continue
+                if (
+                    f"boot-mode.kernelopts.{bootmode_name}"
+                    not in new_bootmode_info
+                ) and bootmode_name != "default":
+                    continue
+                bootmode_feature = untrusted_feature_key
+                bootmode_value = untrusted_feature_value
+                new_bootmode_info[bootmode_feature] = bootmode_value
+                new_bootmode_names.append(bootmode_value)
+        if (
+            # Disallow duplicate boot mode names
+            len(new_bootmode_names) == len(set(new_bootmode_names))
+            # Don't allow more than 64 boot modes
+            and len(new_bootmode_info) <= 64
+            # Don't allow wiping all boot modes
+            and len(new_bootmode_info) > 0
+        ):
+            for feature_key in old_bootmode_info:
+                if feature_key not in new_bootmode_info:
+                    del vm.features[feature_key]
+            for feature_key, feature_val in new_bootmode_info.items():
+                vm.features[feature_key] = feature_val
+        if "boot-mode.active" in untrusted_features:
+            untrusted_feature_value = untrusted_features["boot-mode.active"]
+            if (
+                f"boot-mode.kernelopts.{untrusted_feature_value}" in vm.features
+                or untrusted_feature_value == "default"
+            ):
+                bootmode_value = untrusted_feature_value
+                vm.features["boot-mode.active"] = bootmode_value
+        if "boot-mode.appvm-default" in untrusted_features:
+            untrusted_feature_value = untrusted_features[
+                "boot-mode.appvm-default"
+            ]
+            if (
+                f"boot-mode.kernelopts.{untrusted_feature_value}" in vm.features
+                or untrusted_feature_value == "default"
+            ) and hasattr(vm, "appvm_default_bootmode"):
+                bootmode_value = untrusted_feature_value
+                vm.features["boot-mode.appvm-default"] = bootmode_value
         del untrusted_features
 
         # default user for qvm-run etc
--- a/qubes/tests/ext.py
+++ b/qubes/tests/ext.py
@@ -72,6 +72,7 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
         self.assertListEqual(
             self.vm.mock_calls,
             [
+                ("features.items", (), {}),
                 ("features.get", ("qrexec", False), {}),
                 ("features.__contains__", ("qrexec",), {}),
                 ("features.__setitem__", ("qrexec", True), {}),
@@ -101,6 +102,7 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
         self.assertListEqual(
             self.vm.mock_calls,
             [
+                ("features.items", (), {}),
                 ("features.get", ("qrexec", False), {}),
                 ("features.__contains__", ("qrexec",), {}),
                 ("features.__setitem__", ("qrexec", False), {}),
@@ -126,6 +128,7 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
         self.assertListEqual(
             self.vm.mock_calls,
             [
+                ("features.items", (), {}),
                 ("features.get", ("qrexec", False), {}),
                 ("features.get", ("qrexec", False), {}),
             ],
@@ -147,6 +150,7 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
         self.assertListEqual(
             self.vm.mock_calls,
             [
+                ("features.items", (), {}),
                 ("features.get", ("qrexec", False), {}),
                 ("features.__contains__", ("qrexec",), {}),
                 ("features.__setitem__", ("qrexec", True), {}),
@@ -174,6 +178,7 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
         self.assertEqual(
             self.vm.mock_calls,
             [
+                ("features.items", (), {}),
                 ("features.get", ("qrexec", False), {}),
                 ("features.__contains__", ("gui",), {}),
                 ("features.__setitem__", ("gui", True), {}),
@@ -198,6 +203,7 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
         self.assertListEqual(
             self.vm.mock_calls,
             [
+                ("features.items", (), {}),
                 ("features.get", ("qrexec", False), {}),
                 ("features.__contains__", ("qrexec",), {}),
                 ("features.__setitem__", ("qrexec", True), {}),
@@ -250,6 +256,7 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
         self.assertListEqual(
             self.vm.mock_calls,
             [
+                ("features.items", (), {}),
                 ("features.get", ("qrexec", False), {}),
                 ("features.__contains__", ("qrexec",), {}),
                 ("features.__contains__", ("gui",), {}),
@@ -270,6 +277,7 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
             self.vm.mock_calls,
             [
                 ("features.__setitem__", ("qubes-agent-version", "4.1"), {}),
+                ("features.items", (), {}),
                 ("features.get", ("qrexec", False), {}),
             ],
         )
@@ -287,6 +295,7 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
         self.assertListEqual(
             self.vm.mock_calls,
             [
+                ("features.items", (), {}),
                 ("features.get", ("qrexec", False), {}),
             ],
         )
@@ -301,6 +310,7 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
         self.assertListEqual(
             self.vm.mock_calls,
             [
+                ("features.items", (), {}),
                 ("features.get", ("qrexec", False), {}),
             ],
         )
@@ -315,6 +325,7 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
         self.assertListEqual(
             self.vm.mock_calls,
             [
+                ("features.items", (), {}),
                 ("features.get", ("qrexec", False), {}),
             ],
         )
@@ -329,6 +340,7 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
         self.assertListEqual(
             self.vm.mock_calls,
             [
+                ("features.items", (), {}),
                 ("features.get", ("qrexec", False), {}),
             ],
         )
@@ -354,6 +366,7 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
                 ("features.__setitem__", ("os-distribution", "debian"), {}),
                 ("features.__setitem__", ("os-version", "12"), {}),
                 ("features.__setitem__", ("os-eol", "2026-06-10"), {}),
+                ("features.items", (), {}),
                 ("features.get", ("qrexec", False), {}),
             ],
         )
@@ -385,6 +398,7 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
                 ),
                 ("features.__setitem__", ("os-version", "22.04"), {}),
                 ("features.__setitem__", ("os-eol", "2027-06-01"), {}),
+                ("features.items", (), {}),
                 ("features.get", ("qrexec", False), {}),
             ],
         )
@@ -416,6 +430,7 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
                 ),
                 ("log.warning", unittest.mock.ANY, {}),
                 ("log.warning", unittest.mock.ANY, {}),
+                ("features.items", (), {}),
                 ("features.get", ("qrexec", False), {}),
             ],
         )
@@ -447,6 +462,7 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
                 ),
                 ("log.warning", unittest.mock.ANY, {}),
                 ("log.warning", unittest.mock.ANY, {}),
+                ("features.items", (), {}),
                 ("features.get", ("qrexec", False), {}),
             ],
         )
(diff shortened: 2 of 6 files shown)


ArrayBolt3:
Most of the backend is now working, however see https://github.com/QubesOS/qubes-issues/issues/9775, which blocks this from being fully functional.

ArrayBolt3:
This is now ready for review. This should be merged at the same time as the corresponding frontend PR, https://github.com/QubesOS/qubes-manager/pull/407.

marmarek:
The first thing I see - this needs unit tests. See qubes/tests/ dir. You are most likely interested in ext.py and vm/qubes.py there.

ArrayBolt3:
Yep, I figured those were probably necessary. I'm working on those now.

ArrayBolt3:
Unit tests now added (and broken ones fixed). They may need to be enhanced though, not all of the new code is tested because I haven't figured out how to make the defaults mechanism work seamlessly in the mocks. Not sure if that's even possible, or desirable.

qubes/vm/qubesvm.py
```diff
@@ -660,6 +713,24 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.BaseVM):
...
     #
+    active_bootmode = qubes.property(
+        "active_bootmode",
+        type=str,
+        load_stage=4,
+        default=_default_active_bootmode,
+        doc="Active boot mode for this domain"
+    )
+
+    bootmode_kernelopts = qubes.property(
```
marmarek, line 724: This doesn't need to be visible in qvm-prefs, no? You can make it a normal property (`@property` decorator). But if you want this to be visible in qvm-prefs as a read-only property anyway, use `@qubes.stateless_property` decorator. You have examples of both below.

qubes/vm/qubesvm.py
```diff
@@ -216,6 +216,59 @@ def _setter_kbd_layout(self, prop, value):
...
+    with a leading space. Reset the active_bootmode property if the specified
+    boot mode does not exist.
+    """
+    if self.active_bootmode == "":
+        return ""
+    kernelopts_value = self.features.check_with_template(
+        f"boot-mode.{self.active_bootmode}.kernelopts", None
+    )
+    if kernelopts_value is None:
+        self.active_bootmode = qubes.property.DEFAULT
```
marmarek, line 261: Similar comment as above - changing value when reading a property is risky.

qubes/vm/qubesvm.py
```diff
@@ -216,6 +216,59 @@ def _setter_kbd_layout(self, prop, value):
...
+            bootmode_value = subject.template.appvm_default_bootmode
+            kernelopts = subject.features.check_with_template(
+                f"boot-mode.{bootmode_value}.kernelopts", None
+            )
+            if kernelopts is None:
+                subject.template.appvm_default_bootmode = qubes.property.DEFAULT
+            else:
+                return bootmode_value
+        subject = subject.template
+    return ""
```
marmarek, line 237: Your specification says the default boot mode is `base`. But the code in core_features.py assumes it's called `default`. Neither one is empty string... (personally I like `default` more)

qubes/vm/qubesvm.py
```diff
@@ -216,6 +216,59 @@ def _setter_kbd_layout(self, prop, value):
...
+    )
+
+
+def _default_bootmode_kernelopts(self):
+    """
+    Return the kernel options specified by the currently active boot mode,
+    with a leading space. Reset the active_bootmode property if the specified
+    boot mode does not exist.
+    """
+    if self.active_bootmode == "":
```
marmarek, line 255: And same here about default boot mode name.

qubes/vm/qubesvm.py
```diff
@@ -660,6 +713,24 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.BaseVM):
     #
     # properties loaded from XML
     #
+    active_bootmode = qubes.property(
```
marmarek, line 716: shouldn't this be called just "bootmode"?

qubes/tests/ext.py
marmarek, line 1: Announcement mechanism wants testing too. See `TC_20_Services`, tests 010-012, it's a similar mechanism.

marmarek (changes requested):

marmarek:
and see also CI - both pylint and black complain

qubes/ext/core_features.py
```diff
@@ -104,6 +106,54 @@ async def qubes_features_request(self, vm, event, untrusted_features):
...
+        existing_bootmode_name_list = [name for name in vm.features \
+            if _bootmode_name_key_re.match(name)]
+        existing_bootmode_name_count = len(existing_bootmode_name_list)
+        existing_bootmode_kernelopts_list = [mode for mode in vm.features \
+            if _bootmode_kernelopts_key_re.match(mode)]
+        existing_bootmode_kernelopts_count = len(
+            existing_bootmode_kernelopts_list)
+        for _, (
+            untrusted_feature_key, untrusted_feature_value
+        ) in enumerate(untrusted_features.items()):
```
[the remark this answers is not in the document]
ArrayBolt3: You're right. What happened here was I tried to iterate over the dict, discovered that gave me keys only, tried `enumerate` to see if it would fix it because I forgot about `.items()`, then ended up finally using `.items()` to make that work anyway and then left the enumerate around. Will fix.

qubes/ext/core_features.py
```diff
@@ -104,6 +106,54 @@ async def qubes_features_request(self, vm, event, untrusted_features):
...
+                c in string.printable for c in untrusted_feature_value
+            ):
+                continue
+            if _bootmode_name_key_re.match(untrusted_feature_key):
+                # Only 64 boot modes are allowed
+                if existing_bootmode_name_count >= 64:
+                    continue
+                bootmode_feature = untrusted_feature_key
+                bootmode_value = untrusted_feature_value
+                vm.features[bootmode_feature] = bootmode_value
```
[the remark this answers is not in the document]
ArrayBolt3: That's a good idea. I didn't think of that algorithm when I wrote this, I'll implement that.

qubes/ext/core_features.py
```diff
@@ -104,6 +106,54 @@ async def qubes_features_request(self, vm, event, untrusted_features):
...
+                if untrusted_feature_key == \
+                    "boot-mode.default.kernelopts":
+                    continue
+                bootmode_feature = untrusted_feature_key
+                bootmode_value = untrusted_feature_value
+                vm.features[bootmode_feature] = bootmode_value
+                existing_bootmode_kernelopts_count += 1
+            elif untrusted_feature_key == "boot-mode.active":
+                bootmode_feature = untrusted_feature_key
+                bootmode_value = untrusted_feature_value
```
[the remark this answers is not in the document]
ArrayBolt3: Good idea. Technically the worst that could happen is the boot mode that gets set ends up falling back to default later on, but that's needless confusion and your idea prevents that.

qubes/ext/core_features.py
```diff
@@ -104,6 +106,54 @@ async def qubes_features_request(self, vm, event, untrusted_features):
...
+            ):
+                continue
+            if _bootmode_name_key_re.match(untrusted_feature_key):
+                # Only 64 boot modes are allowed
+                if existing_bootmode_name_count >= 64:
+                    continue
+                bootmode_feature = untrusted_feature_key
+                bootmode_value = untrusted_feature_value
+                vm.features[bootmode_feature] = bootmode_value
+                existing_bootmode_name_count += 1
```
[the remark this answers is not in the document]
ArrayBolt3: Oooh, good catch! Will fix.

qubes/ext/core_features.py
```diff
@@ -104,6 +106,54 @@ async def qubes_features_request(self, vm, event, untrusted_features):
...
                 vm.features["qubes-agent-version"] = untrusted_value
+
+        # handle boot mode advertisement
+        existing_bootmode_name_list = [name for name in vm.features \
+            if _bootmode_name_key_re.match(name)]
+        existing_bootmode_name_count = len(existing_bootmode_name_list)
+        existing_bootmode_kernelopts_list = [mode for mode in vm.features \
+            if _bootmode_kernelopts_key_re.match(mode)]
+        existing_bootmode_kernelopts_count = len(
+            existing_bootmode_kernelopts_list)
```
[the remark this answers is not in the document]
ArrayBolt3: It is true I'm only using this for enforcing the boot mode limit. I was under the impression that a VM would have to explicitly unannounce a boot mode by setting it to be empty when I wrote this, but what you're talking about makes more sense and avoids potential deployment errors.

qubes/vm/qubesvm.py
```diff
@@ -216,6 +216,59 @@ def _setter_kbd_layout(self, prop, value):
...
+    """
+    subject = self
+    while hasattr(subject, "template"):
+        if hasattr(subject.template, "appvm_default_bootmode"):
+            bootmode_value = subject.template.appvm_default_bootmode
+            kernelopts = subject.features.check_with_template(
+                f"boot-mode.{bootmode_value}.kernelopts", None
+            )
+            if kernelopts is None:
+                subject.template.appvm_default_bootmode = qubes.property.DEFAULT
```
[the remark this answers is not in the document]
ArrayBolt3: The goal I was going for was basically when a boot mode stops existing, anything that refers to it needs to reset to defaults. It's not possible to do that when the boot mode is removed because that would require special handling of the boot mode features, but I was trying to implement it with the setter methods as much as possible to reduce the conscious work that had to be done by API consumers.

I'm happy to rewrite this so that all API consumers just do explicit checks and set things to what they should be if necessary. It's a bit more manual effort, but it sounds like I'm unintentionally playing with fire here.

qubes/vm/qubesvm.py
ArrayBolt3: Sounds good, I'll port it to `@property`. (I did like having it show up in qvm-prefs a bit, but it's not at all necessary, *and* it has an annoying space in the front that the libvirt template depends on so it looks bad in qvm-prefs' output anyway.)

qubes/vm/qubesvm.py
ArrayBolt3: The default boot mode being named was only important for the fully feature-based implementation. It's easier IMO to deal with an empty string than a special value since we're using properties now.

qubes/vm/qubesvm.py
ArrayBolt3: I guess that would be OK. I liked `active` personally but it's probably not necessary.

ArrayBolt3:
Thanks for the very thorough review! Will work through these issues and then request another review once ready.

qubes/vm/qubesvm.py
marmarek: What about user-visible name for the default boot mode? IIUC setting that is supported, right? With default name being empty that means feature request `boot-mode..name`, which feels wrong.

ArrayBolt3: I'm not sure that a user-visible name for the default is possible or desirable, since the frontend PR is using `utils.initialize_widget_for_property` with `allow_default=True`, so the default entry always looks like `default (internal-bootmode-id)`. I also don't know if it's really necessary, though it would be *nice* to have. I'll try to look into it more.

If there is a way to get a user-friendly boot mode name in the default entry, then `boot-mode.default.name` can be assigned to it. I agree, `boot-mode..name` feels wrong.

ArrayBolt3:
@marmarek What would you think about me reorganizing the boot mode feature names from `boot-mode.bootmodename.kernelopts` to `boot-mode.kernelopts.bootmodename` and the like? Right now I'm using regex matching to find boot mode related features, when I could be using `str.startswith()` if it was flipped around. That would probably come with performance advantages and less risk of security-related issues.

marmarek:
> @marmarek What would you think about me reorganizing the boot mode feature names from `boot-mode.bootmodename.kernelopts` to `boot-mode.kernelopts.bootmodename` and the like? Right now I'm using regex matching to find boot mode related features, when I could be using `str.startswith()` if it was flipped around. That would probably come with performance advantages and less risk of security-related issues.

Fine with me.

qubes/ext/core_features.py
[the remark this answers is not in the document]
ArrayBolt3: Fixed.

qubes/ext/core_features.py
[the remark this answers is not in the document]
ArrayBolt3: Fixed.

qubes/ext/core_features.py
```diff
@@ -104,6 +106,54 @@ async def qubes_features_request(self, vm, event, untrusted_features):
...
+                vm.features[bootmode_feature] = bootmode_value
+                existing_bootmode_kernelopts_count += 1
+            elif untrusted_feature_key == "boot-mode.active":
+                bootmode_feature = untrusted_feature_key
+                bootmode_value = untrusted_feature_value
+                if vm.active_bootmode == "":
+                    vm.active_bootmode = bootmode_value
+            elif untrusted_feature_key == "boot-mode.appvm-default":
+                bootmode_feature = untrusted_feature_key
+                bootmode_value = untrusted_feature_value
```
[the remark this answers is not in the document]
ArrayBolt3: fixed

qubes/ext/core_features.py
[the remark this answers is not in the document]
ArrayBolt3: Fixed, this turned out to be much less complicated now that we've switched to not needing explicit deletion requests for old boot modes.

qubes/ext/core_features.py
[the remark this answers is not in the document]
ArrayBolt3: Rewrote to work in a similar way to services.py.

qubes/vm/qubesvm.py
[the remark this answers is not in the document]
ArrayBolt3: Alright, there is no more property-setting side effects now. Seems everything works in my manual testing.

qubes/vm/qubesvm.py
ArrayBolt3: Switched to `@property`.

qubes/vm/qubesvm.py
ArrayBolt3: Fixed.

qubes/vm/qubesvm.py
ArrayBolt3: Renamed.

ArrayBolt3:
Still need to deal with pylint and black issues, the ext unit tests, and look at the default boot mode name stuff a bit more. Pushed what I had so far since it resolved most issues.

qubes/tests/ext.py
ArrayBolt3: Done, though I used the approach that was more common in `TC_00_CoreFeatures` since the announcement mechanism is in `core_features.py`, and the test mechanism there seems a bit more robust since it lets me verify things not only reach the right state but are done in the right order too.

ArrayBolt3:
This should now be ready for another review. I still haven't figured out the default boot mode name stuff, but everything else should now be dealt with.

marmarek:
> black issues

hint: `black -l80 qubes` will do the job for you

qubes/ext/core_features.py
```diff
@@ -104,7 +104,90 @@ async def qubes_features_request(self, vm, event, untrusted_features):
             untrusted_value = untrusted_features["qubes-agent-version"]
             if _version_re.fullmatch(untrusted_value):
                 vm.features["qubes-agent-version"] = untrusted_value
+
+        # handle boot mode advertisement
+        old_bootmode_info = {}
+        for feature_key, feature_val in vm.features.items():
+            if feature_key.startswith(
+                "boot-mode.kernelopts."
+            ) or feature_key.startswith("boot-mode.name"):
```
marmarek, line 113:
```suggestion
            ) or feature_key.startswith("boot-mode.name."):
```

qubes/ext/core_features.py
```diff
@@ -104,7 +104,90 @@ async def qubes_features_request(self, vm, event, untrusted_features):
...
+                old_bootmode_info[feature_key] = feature_val
+        new_bootmode_info = {}
+        untrusted_sanitized_features = {}
+        for (
+            untrusted_feature_key,
+            untrusted_feature_value,
+        ) in untrusted_features.items():
+            if not all(c in string.printable for c in untrusted_feature_value):
```
marmarek, line 121: First of all, `untrusted_features` is already filtered for quite limited char set - letters, digits and `"-._ "`: https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/api/misc.py#L61
But if you want to apply other limits anyway, you can probably do that in the next loop, after checking if the feature is relevant for this part of the code; there is no need to sanitize features that are irrelevant for this loop anyway.
Or if you hate to duplicate this line (for both kernelopts and name), then maybe add filtering on `.startswith("boot-mode.")`?

qubes/ext/core_features.py
```diff
@@ -104,7 +104,90 @@ async def qubes_features_request(self, vm, event, untrusted_features):
...
+                continue
+            untrusted_sanitized_features[untrusted_feature_key] = (
+                untrusted_feature_value
+            )
+        for (
+            untrusted_feature_key,
+            untrusted_feature_value,
+        ) in untrusted_sanitized_features.items():
+            if untrusted_feature_key.startswith("boot-mode.kernelopts."):
+                bootmode_name = untrusted_feature_key.split(".")[2]
```
marmarek, line 131: what if the boot mode id has dots? I don't see them being prohibited... maybe they should be?

qubes/ext/core_features.py
```diff
@@ -104,7 +104,90 @@ async def qubes_features_request(self, vm, event, untrusted_features):
...
+                for feature_key in old_bootmode_info:
+                    if feature_key not in new_bootmode_info:
+                        del vm.features[feature_key]
+                for feature_key, feature_val in new_bootmode_info.items():
+                    vm.features[feature_key] = feature_val
+        for (
+            untrusted_feature_key,
+            untrusted_feature_value,
+        ) in untrusted_sanitized_features.items():
+            if untrusted_feature_key == "boot-mode.active":
```
marmarek, line 167: There is no need to iterate over all just for those two keys. Simply do `if "boot-mode.active" in untrusted_features` or similar.

marmarek (changes requested):

qubes/ext/core_features.py
marmarek: or maybe you meant just `boot-mode.name` for the default boot mode name? in that case, it would need to be handled below too

qubes/ext/core_features.py
ArrayBolt3: Hmm, that's a good question. The spec says they should be, but that was originally because I was putting the name in the middle, like `boot-mode.vmreq.kernelopts`. Switching that around to `boot-mode.kernelopts.vmreq` means that a boot mode named `boot-mode.kernelopts....` or `boot-mode.kernelopts.a.b.c` or even just `boot-mode.kernelopts.` should be handled gracefully (though I'd need to test this). I'm not sure whether that qualifies as a bug or a feature, which would you prefer?

qubes/ext/core_features.py
ArrayBolt3: I still don't have default boot mode names implemented, that's still left unresolved from the earlier review. I'll leave some rationale behind leaving them out if I can't get them to work with qubes-manager.

ArrayBolt3:
@marmarek The problem I'm running into with trying to add a default boot mode name is that in qubes-builderv2, `initialize_widget_for_property`, the value used to fill in the default entry is the *value* of the property, not the associated name of the property. I think that could probably be fixed, and if so, adding a default boot mode name should be doable.

qubes/ext/core_features.py
ArrayBolt3: If this is already filtered, I don't think filtering it a second time is needed at all. I'll just remove it.

qubes/ext/core_features.py
marmarek: IMO it will be easier if dots are prohibited in boot mode id.

marmarek:
> @marmarek The problem I'm running into with trying to add a default boot mode name is that in qubes-builderv2, `initialize_widget_for_property`, the value used to fill in the default entry is the _value_ of the property, not the associated name of the property. I think that could probably be fixed, and if so, adding a default boot mode name should be doable.

I guess it can be extended to have extra param for default name? Or even have some mapping value<->display name, as that would be useful for other entries too. Whether to display actual boot mode id too, I'm not sure... maybe not? But then, the API should ensure display names are unique.

ArrayBolt3:
> I guess it can be extended to have extra param for default name?

I managed to implement this basically, but actually using it turned out to be a bit tricky. You can see what I did at https://github.com/ArrayBolt3/qubes-manager/commit/89d0d2d34bbd2ebfe97f7f65e92d45d1d6bb5269 (this branch is highly unstable since I'm using it as a generic catch-all to help me juggle Git things and work between two computers). Basically I have to figure out what the default item's name should be in `qubesmanager/settings.py`, then pass it to `qubesmanager/utils.py` `initialize_widget_for_property`. This means that there's two spots in the code that have to mess with the bootmode property default though, which isn't great. The duplication could be avoided if initialize_widget_for_property could be passed a function that would calculate the name for the default entry, but that would add substantial complexity to the already complex code.

> But then, the API should ensure display names are unique.

Good point, I hadn't thought of that. Probably that code would go in `qubes/ext/core_features.py`; if a VM tries to set multiple boot modes with the same name, reject the boot mode change attempt.

qubes/ext/core_features.py
ArrayBolt3: Alright, I got default boot mode names working! I went with `default` as the name like you suggested.

qubes/ext/core_features.py
ArrayBolt3: Removed.

qubes/vm/qubesvm.py
ArrayBolt3: Changed, the "no params" default boot mode is now `default`.

qubes/vm/qubesvm.py
ArrayBolt3: Resolved, now using `default`.

qubes/ext/core_features.py
ArrayBolt3: Still need to fix this, will do that on the next iteration.

ArrayBolt3:
Not quite ready for re-review yet, I should give this a good self-review first. It passes unit tests and manual testing, but I should read through the entire diff and see if I can spot ways of breaking this.

marmarek:
openQA test says setting default boot mode doesn't work, and I think I know why.

qubes/ext/core_features.py
```diff
@@ -104,6 +104,79 @@ async def qubes_features_request(self, vm, event, untrusted_features):
...
+            for feature_key in old_bootmode_info:
+                if feature_key not in new_bootmode_info:
+                    del vm.features[feature_key]
+            for feature_key, feature_val in new_bootmode_info.items():
+                vm.features[feature_key] = feature_val
+        if "boot-mode.active" in untrusted_features:
+            untrusted_feature_value = untrusted_features["boot-mode.active"]
+            if (
+                f"boot-mode.kernelopts.{untrusted_feature_value}" in vm.features
+                and vm.bootmode == ""
```
marmarek, line 165: This, and the other one below, wants to be "default" now. But also, maybe better check if property isn't in its default value by using `vm.property_is_default("bootmode")`?

ArrayBolt3: In this particular instance, there isn't really a difference between the property being in its default value and the property being set to `default`, since it's illegal to create a boot mode named `default`. But I think using the API as you show is clearer, so I'll go ahead and use it. (It'll also prevent problems if we end up changing the fallback boot mode name to something other than `default`.)

marmarek: There is a very subtle difference - if the user explicitly sets the bootmode variable to "default" using qvm-prefs, before template announces any boot modes support. Checking for literal "default" value will override this choice, while using `property_is_default` will leave it alone.

ArrayBolt3: That's true, though... if the user explicitly sets this to "default" using qvm-prefs, they've broken their system, that isn't ever supposed to be set explicitly. I'll add some code to prevent that.

marmarek: Why shouldn't that be supported? What if something goes wrong with all configured boot modes for some reason and user wants to debug it without extra kernel opts?

ArrayBolt3: Then the user should explicitly *unset* the property (easiest by setting the VM's template's appvm default boot mode to `default ()` and then setting the AppVM's boot mode to `default ()`). `default` is never a boot mode that actually exists, setting it as if it does exist will probably give pretty confusing results when the user opens Qube Manager next (I believe it will actually force-unset the property, thus undoing what the user tried to do).

qubes/ext/core_features.py
ArrayBolt3: Fixed.

ArrayBolt3:
Finished self-review, caught some issues. Also made the rtests a bit better and fixed the "dots in boot mode ID" problem.

qubes/ext/core_features.py
marmarek: But then, the template would override it back to its "preferred" mode next time it re-announce them, no?

marmarek:
Sorry for throwing ideas so late, but what do you think about this:
- when template request `boot-mode.active` feature, save it as a feature (potentially overriding existing value), instead of setting the property
- in `_default_bootmode` getter, look for this feature first (and then go iterating over templates chain)
- do similar thing with `boot-mode.appvm-default`

This way, template will be able to switch its boot mode, and also update default boot mode for its appvms. But only if user didn't explicitly set some other value.
This approach is more friendly to seamless updates, but on the other hand, may look a bit more magic to the user.

ArrayBolt3:
@marmarek No worries about late or not late giving of ideas, I'm fine with the feature taking longer to merge if it comes out better :)

I'm not sure I totally like the idea of letting the VM set its own boot mode unless the user explicitly set the boot mode some other way. For Whonix's use case, the boot modes will be set so that AppVMs use user mode, and the template uses sysmaint mode. If the AppVM could request its boot mode to be changed to something other than the default, it would be trivial for software within an AppVM to make the VM always boot in sysmaint mode instead. This would result in a rather dramatic change for Whonix on the desktop, since sysmaint mode looks nothing like user mode, but under Qubes, telling the difference might not be so easy. I suspect this could be used maliciously, though I haven't fully thought it through yet.

Hmm... for that matter, `boot-mode.active` is unset when falling back to the AppVM default boot mode. So this is a threat scenario that still applies to the current code. That's probably not good. (This turned out to be trickier than I thought!)

qubes/ext/core_features.py
marmarek: And generally, the default value should be just one of possible property values, chosen to be default. Not a magic value that cannot be set otherwise. In other words, `qvm-prefs vm property "$(qvm-prefs vm property)"` should always work

marmarek:
> letting the VM set its own boot mode

Not any VM, just TemplateVM. Feature request from AppVMs are ignored in the place you added boot mode support (see beginning of that function).
But StandaloneVM could change its boot mode this way (unless user set it explicitly ofc).

qubes/ext/core_features.py
ArrayBolt3: > And generally, the default value should be just one of possible property values, chosen to be default. Not a magic value that cannot be set otherwise. In other words, `qvm-prefs vm property "$(qvm-prefs vm property)"` should always work

Oof, I did not know this and didn't design with this in mind. The current design doesn't necessarily *depend* on this not being the case, but it definitely hasn't been tested well enough to be sure that it will work properly and safely if the user does this.

> But then, the template would override it back to its "preferred" mode next time it re-announce them, no?

True, but that's only if they open Qube Manager and it actually does unset the invalid `default` boot mode entry. If they don't, and it sticks around, then what will probably happen is the VM will be "stuck" in default mode, not allowing the template to change the active or AppVM default boot modes during advertisement. I guess I'll have to take a closer look at this and remove the faulty assumption from the code.

ArrayBolt3:
Ah, good point, I forgot about the restriction against AppVMs requesting features. I guess then this would be good to do.

ArrayBolt3:
@marmarek Implemented your suggestion w.r.t. making `boot-mode.active` and `boot-mode.appvm-default` into features that back their respective properties. Also attempted to fix the assumption that the user would never explicitly set `default` as a boot mode (though I haven't tested that this works as intended yet).

There is one problem left here that I'm not sure if it's a Qubes Manager issue or a qubes-core-admin issue, but it's basically a data consistency problem. In essence, if I open Qubes Manager, look at an AppVM's boot modes, then use CLI tools in dom0 to fiddle with the template's boot modes in ways that have a meaningful effect on the AppVM (such as changing which boot mode the default points to), I can sometimes get it so that the "Boot mode kernel opts" section shows the wrong kernel options (either showing the options for some other entry, or showing no options at all). It almost feels as if Qubes Manager is caching data about VMs, and that data is getting out-of-sync with the actual system state. Closing and re-opening the settings dialog doesn't fix this, I have to close and re-open Qubes Manager itself for things to resolve. I haven't figured out reliable reproduction steps *yet*, but I wanted to mention this here in case it gave you any ideas about what might be happening.

marmarek:
Some caching is involved there, and indeed that may be the problem. Events should invalidate respective cache parts, and in most cases those events are automatically generated when a value changes. But in case of properties with complicated default value, there may be case where event need to be triggered manually. I think it is the case when `boot-mode.active` / `boot-mode.appvm-default` features change. Try something like this:
```python
@qubes.events.handler(`feature-set:boot-mode.active`)
def on_feature_bootmode_active_set(self, event, name, newvalue, oldvalue=None):
    if newvalue == oldvalue:
        return
    if self.property_is_default("bootmode"):
        self.fire_event("property-reset:bootmode", name="bootmode")
```
and similar for appvm-default (but here with iterating over `getattr(self, "appvms", [])`)

ArrayBolt3:
what in the world... I ran rtests on this and did detailed manual testing, maybe I mangled something during Git gymnastics.

ArrayBolt3:
ugh, yep, failed to push a commit to my workjunk branch and so it didn't get tested.

ArrayBolt3:
Fixed the syntax error and some broken tests. Also checked very carefully to make sure all changes were pushed *and tested*.

ArrayBolt3:
Was able to get a list of steps that reliably reproduce the issue:

* Ensure that fedora-41-xfce and an AppVM based on it have *no* boot mode info (delete any features that start with `boot-mode.`, reset `bootmode` and `appvm_default_bootmode` properties to default).
* Open Qube Manager.
* Open and close the settings for fedora-41-xfce and the AppVM you chose earlier.
* In a Dom0 terminal, run:

```bash
qvm-features fedora-41-xfce boot-mode.name.mode1 'Mode One'
qvm-features fedora-41-xfce boot-mode.name.mode2 'Mode Two'
qvm-features fedora-41-xfce boot-mode.kernelopts.mode1 'mode1kern'
qvm-features fedora-41-xfce boot-mode.kernelopts.mode2 'mode2kern'
qvm-features fedora-41-xfce boot-mode.appvm-default 'mode1'
```

* Open the settings for the derived AppVM.
* Click the 'Advanced' tab.

Expected result: "Boot mode" is set to "default (Mode One) (current)", "Boot mode kernel opts" shows "mode1kern".

Actual result: "Boot mode" is set to "default (Mode One) (current)", "Boot mode kernel opts" is blank.

I tried patching my fork of qubes-core-admin with the following patch to see if it would fix it, however it does not, the above reproduction steps still give the actual result rather than the expected result:

```diff
diff --git a/qubes/vm/qubesvm.py b/qubes/vm/qubesvm.py
index e0e865f9..4f71e24e 100644
--- a/qubes/vm/qubesvm.py
+++ b/qubes/vm/qubesvm.py
@@ -1203,6 +1203,15 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.BaseVM):
             self._domain_stopped_event_received = False
             self._domain_stopped_event_handled = False
 
+    @qubes.events.handler("feature-set:boot-mode.active")
+    def on_feature_bootmode_active_set(
+        self, event, name, newvalue, oldvalue=None
+    ):
+        if newvalue == oldvalue:
+            return
+        if self.property_is_default("bootmode"):
+            self.fire_event("property-reset:bootmode", name="bootmode")
+
     @qubes.events.handler("property-set:label")
     def on_property_set_label(self, event, name, newvalue, oldvalue=None):
         # pylint: disable=unused-argument
diff --git a/qubes/vm/templatevm.py b/qubes/vm/templatevm.py
index fcbcc27b..293fcfe3 100644
--- a/qubes/vm/templatevm.py
+++ b/qubes/vm/templatevm.py
@@ -107,6 +107,24 @@ class TemplateVM(QubesVM):
         }
         super().__init__(*args, **kwargs)
 
+    @qubes.events.handler("feature-set:boot-mode.appvm-default")
+    def on_feature_bootmode_appvm_default_set(
+        self, event, name, newvalue, oldvalue=None
+    ):
+        if newvalue == oldvalue:
+            return
+        if self.property_is_default("appvm_default_bootmode"):
+            self.fire_event(
+                "property-reset:appvm_default_bootmode",
+                name="appvm_default_bootmode"
+            )
+            for appvm in getattr(self, "appvms", []):
+                if appvm.property_is_default("bootmode"):
+                    appvm.fire_event(
+                        "property-reset:bootmode",
+                        name="bootmode"
+                    )
+
     @qubes.events.handler(
         "property-set:default_user",
         "property-set:kernel",
```

marmarek:
Just to be sure - have you restarted qubesd after adding those events?
You can also add debug print in the loop in the appvm_default handler, to see if that event is fired - this is the one that should invalidate relevant cache in your scenario

ArrayBolt3:
I might have forgotten now that you mention it. Rebooting and retesting.

ArrayBolt3:
@marmarek Yes, this does occur even after restarting qubesd (I rebooted the machine entirely). So I guess it's time for more debugging.

ArrayBolt3:
Pushed the fix for the data consistency problem. I think the only problem I know of that's left is a label formatting issue in Qube Manager. The backend code should be ready for review again.

ArrayBolt3:
Did another push because black had to reformat some things.

marmarek:
pylint complains...

ArrayBolt3:
Fighting with event handler regression test issues, not ready for review yet...

ArrayBolt3:
@marmarek I tried to fix the pylint issues, which seems to have made it happy, but the regression tests now fail:

```
ERROR: qubes.tests.vm.qubesvm/TC_90_QubesVM/test_811_default_bootmode
    return func(*args, **kwds)
  File "/home/user/qubes-core-admin/qubes/tests/vm/qubesvm.py", line 3121, in test_811_default_bootmode
    vm.features["boot-mode.active"] = "testmode1"
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/home/user/qubes-core-admin/qubes/features.py", line 93, in __setitem__
    self.subject.fire_event(
    ~~~~~~~~~~~~~~~~~~~~~~~^
        "domain-feature-set:" + key, feature=key, value=value
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/user/qubes-core-admin/qubes/events.py", line 200, in fire_event
    sync_effects, async_effects = self._fire_event(
                                  ~~~~~~~~~~~~~~~~^
        event, kwargs, pre_event=pre_event
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/user/qubes-core-admin/qubes/events.py", line 169, in _fire_event
TypeError: QubesVM.on_feature_bootmode_active_set() missing 1 required positional argument: 'event'
```

I can't figure out why this happens though - `qubes/vm/qubesvm.py -> on_feature_bootmode_active_set` uses this "signature" (not sure what these are called in Python):

```
    @qubes.events.handler("domain-feature-set:boot-mode.active")
    def on_feature_bootmode_active_set(
        self, vm, event, feature, value, oldvalue=None
    ):
```

This is nearly identical to a signature in `qubes/ext/services.py -> on_domain_feature_set`:

```
    @qubes.ext.handler("domain-feature-set:*")
    def on_domain_feature_set(self, vm, event, feature, value, oldvalue=None):
```

How come the second one works but the first one doesn't? (The first one used to work without problems before adding `self` as the first argument, as required by `pylint`.)

marmarek:
> self, vm,

self _is_ vm, don't need to repeat it. It's the difference between event handler defined in the VM class itself (`@qubes.events.handler`) and in an extension (`@qubes.ext.handler`)

qubes/ext/core_features.py
ArrayBolt3: Fixed this some time ago, it should now be perfectly fine for the user to assign a boot mode of "default".

ArrayBolt3:
Fixed the issue with the failing regression tests.

ArrayBolt3:
@marmarek How would I run the code coverage test locally? It looks like there's some missed stuff I would have liked to have caught on my end earlier. Currently working on resolving that now.

marmarek:
The run-tests scripts does run with code coverage enabled. You can get the report with `coverage` tool (or `python3 -m coverage`)

ArrayBolt3:
qubesvm.py line 1010 should be unreachable code, thus why it shows as not covered. (`_default_bootmode()` shouldn't ever return something that isn't found by `self.features.check_with_template()`, except of course for the `default` value.) All the other code coverage issues should now be solved.

marmarek:
Black complains about one tiny thing...

ArrayBolt3:
Should be fixed, I'll try to get in the habit of running it before all pushes.

(aside, qubes-manager is NOT formatted by black, and I've caused myself major headaches more than once thinking I'm supposed to reformat it after changing it and then realizing I just changed like 90% of the code in a way that's tricky to revert if you had uncommitted, unstaged changes, or were silly enough to run `git add -A` after that. Is there a reason it isn't "blackened"?)

marmarek:
> Is there a reason it isn't "blackened"

It isn't _yet_

marmarek (approved):

PR #656, opened 2025-02-16
Add free-form text to qube for notes, comments, ...

alimirjamali:
Core and API part of adding free-form text to each qube for comments, notes, descriptions, remarks, reminders, etc.

fixes: https://github.com/QubesOS/qubes-issues/issues/899

### DIFF
--- a/qubes/api/admin.py
+++ b/qubes/api/admin.py
@@ -28,6 +28,8 @@ import string
 import subprocess
 import pathlib
 
+from ctypes import CDLL
+
 import libvirt
 import lxml.etree
 import importlib.metadata
@@ -52,6 +54,9 @@ from qubes.device_protocol import (
     AssignmentMode,
     DeviceInterface,
 )
+
+# To validate & sanitise UTF8 strings
+LIBQUBES_PURE = "libqubes-pure.so.0"
 
 
 class QubesMgmtEventsDispatcher:
@@ -2036,3 +2041,50 @@ class QubesAdminAPI(qubes.api.AbstractQubesAPI):
             "power_state": self.dest.get_power_state(),
         }
         return " ".join("{}={}".format(k, v) for k, v in state.items())
+
+    @qubes.api.method(
+        "admin.vm.notes.Get", no_payload=True, scope="local", read=True
+    )
+    async def vm_notes_get(self):
+        """Get qube notes"""
+        self.enforce(self.dest.name != "dom0")
+        self.fire_event_for_permission()
+        notes = self.dest.get_notes()
+        return notes
+
+    @qubes.api.method("admin.vm.notes.Set", scope="local", write=True)
+    async def vm_notes_set(self, untrusted_payload):
+        """Set qube notes"""
+        self.enforce(self.dest.name != "dom0")
+        self.fire_event_for_permission()
+        if len(untrusted_payload) > 256000:
+            raise qubes.exc.ProtocolError(
+                "Maximum note size is 256000 bytes ({} bytes received)".format(
+                    len(untrusted_payload)
+                )
+            )
+
+        # Sanitise the incoming utf8 notes with libqubes-pure
+        try:
+            libqubespure = CDLL(LIBQUBES_PURE)
+            notes = "".join(
+                [
+                    (
+                        c
+                        # first we check with our advanced unicode sanitisation
+                        if libqubespure.qubes_pure_code_point_safe_for_display(
+                            ord(c)
+                        )
+                        # validate tab and newline since qubespure excludes them
+                        or c in "\t\n"
+                        else "_"
+                    )
+                    for c in untrusted_payload.decode("utf8")
+                ]
+            )
+        except Exception as e:
+            raise qubes.exc.ProtocolError(
+                "Unable to sanitise qube notes: " + str(e)
+            )
+
+        self.dest.set_notes(notes)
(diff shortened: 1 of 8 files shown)


marmarek:
Black complains

alimirjamali:
> Black complains

Black complains are resolved. And the Codecov is advising on the required unittests. I guess I can proceed with the unittests? And I should assure that proper errors are raised if necessary.

qubes/vm/qubesvm.py
```diff
@@ -2585,6 +2585,40 @@ def kernelopts_common(self):
...
+    def set_notes(self, notes: str) -> bool:
+        """Write to notes file"""
+        try:
+            with open(
+                os.path.join(self.dir_path, self.notes_file),
+                "w",
+                encoding="ascii",
+            ) as fd:
+                fd.write(notes)
+                fd.close()
```
marmarek, line 2617: File is closed by the context manager already

qubes/vm/qubesvm.py
```diff
@@ -2585,6 +2585,40 @@ def kernelopts_common(self):
...
+        return "notes.txt"
+
+    def get_notes(self) -> str:
+        """Read the notes file and return its content"""
+        try:
+            with open(
+                os.path.join(self.dir_path, self.notes_file), encoding="ascii"
+            ) as fd:
+                return fd.read()
+        except EnvironmentError:  # pylint: disable=broad-except
```
marmarek, line 2604: FileNotFoundError?

qubes/vm/qubesvm.py
alimirjamali: Done

qubes/vm/qubesvm.py
alimirjamali: Done.

qubes/api/admin.py
```diff
@@ -2038,3 +2038,43 @@ async def vm_current_state(self):
...
+    @qubes.api.method(
+        "admin.vm.notes.Get", no_payload=True, scope="local", read=True
+    )
+    async def vm_notes_get(self):
+        """Get qube notes"""
+        self.enforce(self.dest.name != "dom0")
+        self.fire_event_for_permission()
+        try:
+            notes = self.dest.get_notes()
+        except Exception as e:
```
marmarek, line 2051: I'm not sure if this broad except is a good idea, it may make bug reports (if any) hard to  follow, as it will hide the backtrace, and for many built-in exception message is quite enigmatic (for example KeyError without context is hard to find).
get_notes() already has handing for expected errors (if some are missing, should be added there)

qubes/api/admin.py
```diff
@@ -2038,3 +2038,43 @@ async def vm_current_state(self):
...
+        allowed_chars = string.printable
+        notes = "".join(
+            [
+                c if c in allowed_chars else "_"
+                for c in untrusted_payload.decode("ascii")
+            ]
+        )
+        try:
+            self.dest.set_notes(notes)
+        except Exception as e:
```
marmarek, line 2077: same comment here

qubes/tests/api_admin.py
```diff
@@ -2130,6 +2130,50 @@ def test_450_property_reset(self):
...
+    def test_notes_set(self):
+        self.app.domains["test-vm1"].set_notes = unittest.mock.Mock()
+        with self.assertRaises(qubes.exc.ProtocolError):
+            payload = ("x" * 256001).encode()
+            self.call_mgmt_func(
+                b"admin.vm.notes.Set",
+                b"test-vm1",
+                payload=payload,
+            )
+        with self.assertRaises(qubes.exc.QubesException):
```
marmarek, line 2161: this should wrap only the last call_mgmt_func call, no?

marmarek (changes requested):

qubes/api/admin.py
alimirjamali: This should be done

qubes/api/admin.py
alimirjamali: This should be done as well.

qubes/tests/api_admin.py
alimirjamali: This should be fixed as well

marmarek (changes requested):
Something went wrong, multi-line notes don't work - looks like newline char is replaced with `_`. When you fix it, a test for this case would be useful too.

alimirjamali:
> Something went wrong, multi-line notes don't work - looks like newline char is replaced with `_`. When you fix it, a test for this case would be useful too.

It appears that `qubes_pure_string_safe_for_display` counts space as a safe character for display but does not count `\n\t` as safe characters.

marmarek (approved):

qubes/api/admin.py
```diff
@@ -2036,3 +2041,50 @@ async def vm_current_state(self):
...
+
+        # Sanitise the incoming utf8 notes with libqubes-pure
+        try:
+            libqubespure = CDLL(LIBQUBES_PURE)
+            notes = "".join(
+                [
+                    (
+                        c
+                        # first we check with our advanced unicode sanitisation
+                        if libqubespure.qubes_pure_string_safe_for_display(
```
marmarek, line 2075: Is there any reason why not simply `qubes_pure_code_point_safe_for_display(ord(c))`?

marmarek (changes requested):

qubes/api/admin.py
alimirjamali: No actual reason. The suggested change is applied

marmarek (approved):

PR #658, opened 2025-02-17
Add the feature to prohibit starting of a qube

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/9622

### DIFF
--- a/qubes/vm/qubesvm.py
+++ b/qubes/vm/qubesvm.py
@@ -369,6 +369,7 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.BaseVM):
         .. event:: domain-start-failed (subject, event, reason)
 
             Fired when :py:meth:`start` method fails.
+            or if domain has a `prohibit-start` feature.
             *reason* argument is a textual error message.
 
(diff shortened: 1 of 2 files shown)


qubes/vm/qubesvm.py
```diff
@@ -1302,6 +1302,14 @@ async def start(
 
             await self._ensure_shutdown_handled()
 
+            if self.features.get("prohibit-start", False):
+                await self.fire_event_async(
+                    "domain-start-failed", reason="Qube start is prohibited!"
```
marmarek, line 1307: This event used to be fired only after `domain-pre-start`. It isn't explicitly documented, but I think the new behavior wants mentioning in the event documentation (qubesvm.py).

And also, the message (and the exception below) should IMO include the reason taken from the feature value.

marmarek (changes requested):
And please also add a tests. Since unit test may require mocking most of the start method, it can be in integration tests (for example qubes/tests/integ/basic.py)

qubes/vm/qubesvm.py
alimirjamali: Done

PR #660, opened 2025-02-21
Preload disposables

ben-grande:

---

Untested and incomplete.


qubes/api/internal.py
```diff
@@ -361,3 +361,65 @@ async def suspend_post(self):
...
+    @qubes.api.method(
+        "internal.dispvm.preload.Create", scope="local", write=True
+    )
+    async def dispvm_preload_create(self):
+        """
+        Method called to create a preloaded DispVM.
+
+        :return:
+        """
+        dispvm_template = self.arg
```
marmarek, line 374: `self.dest`?

qubes/api/internal.py
```diff
@@ -361,3 +361,65 @@ async def suspend_post(self):
...
+        preload_dispvm = dispvm_template.features.get("preload-dispvm", None)
+        ## TODO: should this function receive disposable name as argument
+        ##       or call admin.vm.CreateDisposable?
```
marmarek, line 377: IMO makes more sense to create here.

There should be also somewhere a counter how many preloaded disposables should be and if that limit (which may be 0) is reached already, do nothing here.

qubes/api/internal.py
```diff
@@ -361,3 +361,65 @@ async def suspend_post(self):
...
+    async def dispvm_preload_use(self):
+        """
+        Method called when usage of preloaded DispVM is requested.
+
+        To use the first preloaded disposable, provide as argument the
+        disposable template, else, provide as argument the disposable name.
+
+        :return:
+        """
+        used_dispvm = self.arg
```
marmarek, line 398: `self.dest`?

qubes/api/internal.py
```diff
@@ -361,3 +361,65 @@ async def suspend_post(self):
...
+            raise qubes.exc.QubesException("Invalid argument provided")
+        preload_dispvm = dispvm_template.features.get("preload-dispvm", None)
+        if not preload_dispvm:
+            return
+
+        preload_dispvm = preload_dispvm.split(" ")
+        if use_first:
+            used_dispvm_name = preload_dispvm[1:]
+        else:
+            used_dispvm_name = used_dispvm.name
```
marmarek, line 418: This should also check if the selected dispvm is actually one of preloaded ones.
But also, what is the use case of pointing at specific DispVM, instead of its template?

qubes/api/internal.py
```diff
@@ -361,3 +361,65 @@ async def suspend_post(self):
...
+            preload_dispvm += " " + dispvm
+        else:
+            preload_dispvm = dispvm
+        dispvm_template.features["preload-dispvm"] = preload_dispvm
+        dispvm.features["internal"] = True
+
+    @qubes.api.method(
+        "internal.dispvm.preload.Use", scope="local", write=True
+    )
+    async def dispvm_preload_use(self):
```
marmarek, line 389: This should probably be an utility function somewhere, not an API method exposed to external callers.

qubes/vm/dispvm.py
```diff
@@ -187,11 +187,37 @@ def __init__(self, app, xml, *args, **kwargs):
...
     @qubes.events.handler("domain-load")
     def on_domain_loaded(self, event):
         """When domain is loaded assert that this vm has a template."""  # pylint: disable=unused-argument
         assert self.template
 
+    @qubes.events.handler("domain-started")
+    def on_domain_started(self, event, **kwargs):
+        """Pause preloaded domains as soon as they start."""
+        if self.is_domain_preloaded():
+            self.pause()
```
marmarek, line 211: This is easy thing to do, but requires testing if it isn't too early. Some services (especially gui-agent) may still be starting.
There is an `qubes.WaitForSession` service that may help (ensure to use async handler to not block qubesd while waiting on it).
And also, the event is called `domain-start`.

qubes/vm/dispvm.py
```diff
@@ -187,11 +187,37 @@ def __init__(self, app, xml, *args, **kwargs):
...
+
+    @qubes.events.handler("domain-unpaused")
+    def on_domain_unpaused(self):
+        """Mark unpaused preloaded domains as used."""
+        if self.is_domain_preloaded():
+            self.app.qubesd_call(
```
marmarek, line 217: This doesn't exist in core-admin - it's only in core-admin-client. Here, you are running inside qubesd already, you can simply call appropriate function directly.

qubes/api/internal.py
marmarek: But also, who will call it? If something from inside qubesd, then it doesn't need to be an "API" method, just a function somewhere, probably next to the dispvm class.

qubes/vm/dispvm.py
```diff
@@ -187,11 +189,130 @@ def __init__(self, app, xml, *args, **kwargs):
...
+        while True:
+            ## TODO: Is there existing Qubes code that checks available memory
+            ##       before starting a qube?
+            memory = getattr(self, "memory", 0)
+            available_memory = (
+                psutil.virtual_memory().available / (1024 * 1024)
+            )
+            threshold = 1024 * 5
+            if memory >= (available_memory - threshold):
+                await DispVM.create_preloaded_dispvm(self)
```
ben-grande, line 287: This is wrong, should have called `QubesAdminAPI.create_disposable` with `preload` argument.

qubes/vm/dispvm.py
```diff
@@ -187,11 +189,130 @@ def __init__(self, app, xml, *args, **kwargs):
...
+    async def on_domain_preloaded_dispvm_used(self, event, delay=5):
+        """When preloaded DispVM is used or after boot, preload another one."""  # pylint: disable=unused-argument
+        ## TODO: fire event after start of all qubes that set to autostart.
+        await asyncio.sleep(delay)
+        while True:
+            ## TODO: Is there existing Qubes code that checks available memory
+            ##       before starting a qube?
+            memory = getattr(self, "memory", 0)
+            available_memory = (
+                psutil.virtual_memory().available / (1024 * 1024)
```
marmarek, line 283: I get what you mean, but this will not work. This looks only at free memory in dom0, not the whole system. And even if it would look more globally, qmemman tries to allocate available memory as much as possible. Only qmemman knows how much "free" memory you really have, and currently there is no API to query that...

qubes/vm/dispvm.py
marmarek: > What is preloaded dispvm has `gui` feature disabled in case the user is using the disposable to be qubes-builder-dvm for example, that doesn't require a GUI? What can be done in this case?

Very good question. In that case I guess start even is all we have. Maybe with some (configurable?) delay?

qubes/vm/dispvm.py
```diff
@@ -187,11 +189,130 @@ def __init__(self, app, xml, *args, **kwargs):
...
+            self.klass == "AppVM"
+            and getattr(self, "template_for_dispvms", False)
+        ):
+            dispvm_template = dispvm
+        else:
+            raise qubes.exc.QubesException("Invalid argument provided")
+
+        preload_dispvm = dispvm_template.features.get("preload-dispvm", None)
+        if not preload_dispvm:
+            return
```
marmarek, line 253: This looks wrong - this shouldn't happen, as `use_preloaded_dispvm` should only be called on a preloaded dispvm, at this point there should be something in this feature. This should at least be a warning, if not even an exception.

qubes/vm/dispvm.py
```diff
@@ -187,11 +189,130 @@ def __init__(self, app, xml, *args, **kwargs):
...
+
+        preload_dispvm = preload_dispvm.split(" ")
+        if self.klass == "DispVM":
+            if dispvm.name not in preload_dispvm:
+                raise qubes.exc.QubesException("DispVM is not preloaded")
+            used_dispvm = dispvm.name
+        else:
```
marmarek, line 260: same comment as above about the class

qubes/vm/dispvm.py
```diff
@@ -187,11 +189,130 @@ def __init__(self, app, xml, *args, **kwargs):
...
+            used_dispvm = preload_dispvm[0]
+            dispvm = self.app.domains[used_dispvm]
+
+        unused_dispvm = preload_dispvm.remove(used_dispvm)
+        dispvm_template.features["preload-dispvm"] = unused_dispvm
+        dispvm.features["internal"] = False
+        ## TODO: confirm if this guess of event code is correct
+        self.app.fire_event_for_permission(dispvm_template=dispvm_template)
```
marmarek, line 268: This is relevant only for admin api methods, not needed here.

qubes/vm/dispvm.py
```diff
@@ -187,11 +189,130 @@ def __init__(self, app, xml, *args, **kwargs):
...
+        dispvm_template.fire_event_async("domain-preloaded-dispvm-used")
```
marmarek, line 269: _async needs `await`. And also, it may be useful to add dispvm (name? object?) as an event parameter.
And please document the event (there is specific sphinx syntax for events, see others for example).

qubes/vm/dispvm.py
```diff
@@ -187,11 +189,130 @@ def __init__(self, app, xml, *args, **kwargs):
...
+        Create preloaded DispVM.
+
+        Template from which the VM should be created.
+
+        :return:
+        """
+        preload_dispvm_max = int(
+            self.features.check_with_template("preload-dispvm-max", 0)
+        )
+        if preload_dispvm_max == 0:
```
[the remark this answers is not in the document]
marmarek: This check probably should be in `from_appvm` when `preload=True`. And also similar check for `>0`, if there aren't enough preloaded dispvms already (and if there are, probably raise an exception, not just return).

marmarek (changes requested):
I think the overall structure is getting close, but there are still several details. And tests missing, but this you know.

qubes/vm/dispvm.py
```diff
@@ -187,11 +190,170 @@ def __init__(self, app, xml, *args, **kwargs):
...
+        appvm.features["preload-dispvm"] = preload_dispvm
+        self.features["internal"] = False
+        await appvm.fire_event_async(
+            "domain-preloaded-dispvm-used", dispvm=self
+        )
+
+    @qubes.events.handler(
+        "domain-preloaded-dispvm-used", "domain-preloaded-dispvm-autostart"
+    )
+    async def on_domain_preloaded_dispvm_used(self, event, delay=5, **kwargs):  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
ben-grande: Will fix.

linux/systemd/qubes-preload-dispvm.service
```diff
@@ -0,0 +1,17 @@
...
+Description=Preload Qubes DispVMs
+After=qubes-vm@.service
+# TODO: or 'Requires='? Should a failure to autostart sys-(usb|net) make this
+# unit be skipped?
+Wants=qubes-vm@.service
+ConditionKernelCommandLine=!qubes.skip_autostart
+
+[Service]
+Type=oneshot
+Environment=DISPLAY=:0
```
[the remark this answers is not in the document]
marmarek: Yes, things have changed since 2013, now qvm-start only sends an API call to qubesd, it doesn't start those services as its own children anymore.

linux/systemd/qubes-vm@.service
```diff
@@ -1,12 +1,12 @@
 [Unit]
 Description=Start Qubes VM %i
 After=qubesd.service qubes-meminfo-writer-dom0.service
+Before=qubes-vm@.service
```
marmarek, line 4:
```suggestion
Before=qubes-preload-dispvm.service
```

qubes/tests/api_admin.py
```diff
@@ -3716,6 +3716,45 @@ def test_642_vm_create_disposable_not_allowed(self, storage_mock):
...
+        self.assertTrue(self.app.save.called)
+
+    @unittest.mock.patch("qubes.storage.Storage.create")
+    def test_643_vm_create_disposable_preload_autostart(self, mock_storage):
+        mock_storage.side_effect = self.dummy_coro
+        self.vm.template_for_dispvms = True
+        self.app.domains[self.vm].fire_event = self.emitter.fire_event
+        self.vm.features["preload-dispvm-max"] = "1"
+        self.app.default_dispvm = self.vm
+        # TODO: how to mock start of a qube that we don't have its object?
```
marmarek, line 3743: you can also patch the start method at the class level, like `unittest.mock.patch("qubes.vm.dispvm.DispVM.start`. A heavier hammer, but should be fine in this test context.

qubes/vm/dispvm.py
```diff
@@ -187,11 +189,115 @@ def __init__(self, app, xml, *args, **kwargs):
...
+    ):  # pylint: disable=unused-argument
+        """Pause preloaded domains as soon as basic services started."""
+        # TODO:
+        # Marek: Test if pause isn't too early. Some services (especially:
+        #   gui-agent) may still be starting.  qubes.WaitForSession service may
+        #   help (ensure to use async handler to not block qubesd while waiting
+        #   on it).
+        # TODO:
+        #   Test if pause isn't too late, what if application autostarts, will
+        #   it open before the qube is paused?
```
marmarek, line 262: Yes, it will. Theoretically there is an "invisible" mode for gui-daemon for situation like this (it was used for very old implementation of DispVM that also kinda preloaded it).  But there is no support for flipping it in runtime, gui-daemon needs to be restarted for that, so that's a broader change to use it in this version. Maybe later, I'd say it's okay to ignore this issue for now.

qubes/tests/api_admin.py
ben-grande: Did this. Thanks.

qubes/vm/dispvm.py
ben-grande: Add this comment there. Nothing to do now.

ben-grande:
Removed draft but still not ready yet, but Gitlab CI fails to fetch amended and force pushed commits when using the Github API.

Having problems with events not being fired, trying to understand it.

linux/aux-tools/preload-dispvm
```diff
@@ -0,0 +1,21 @@
...
+    appvms = [
+        qube
+        for qube in domains
+        if int(qube.features.get("dispvm-preload-max", 0)) > 0
+        and qube.klass == "AppVM"
+        and getattr(qube, "template_for_dispvms", False)
+    ]
+    method = "admin.vm.CreateDisposable"
+    for qube in appvms:
+        qube.qubesd_call("dom0", method, "preload-autostart")
```
marmarek, line 17: This looks wrong, as the first argument is the call target. So, it calls the method on dom0 several times, instead of each disposable template...
See how qubesd_call is used in qubesadmin.vm.QubesVM. But since that uses private attribute, I guess it needs a wrapper (and then make all `self.qubesd_call(self._method_dest, ...)` use that new wrapper?)

ben-grande: https://github.com/QubesOS/qubes-core-admin-client/blob/main/qubesadmin/vm/__init__.py#L436-L441

```python3
            if self._method_dest.startswith('$dispvm:'):
                method_dest = self._method_dest[len('$dispvm:'):]
            else:
                method_dest = 'dom0'
            dispvm = self.app.qubesd_call(method_dest,
                'admin.vm.CreateDisposable')
```

I guess it can always be the disposable template in this case?

```python3
    for qube in appvms:
        qube.qubesd_call(qube.name, method, "preload-autostart")
```

marmarek: Yes, the latter should work too.

marmarek:
This didn't went far:
```
Mar 19 17:31:15.017096 dom0 qubesd[4569]: unhandled exception while calling src=b'dom0' meth=b'admin.vm.Start' dest=b'sys-firewall' arg=b'' len(untrusted_payload)=0
Mar 19 17:31:15.017096 dom0 qubesd[4569]: Traceback (most recent call last):
Mar 19 17:31:15.017096 dom0 qubesd[4569]:   File "/usr/lib/python3.13/site-packages/qubes/api/__init__.py", line 333, in respond
Mar 19 17:31:15.017096 dom0 qubesd[4569]:     response = await self.mgmt.execute(
Mar 19 17:31:15.017096 dom0 qubesd[4569]:                ^^^^^^^^^^^^^^^^^^^^^^^^
Mar 19 17:31:15.017096 dom0 qubesd[4569]:         untrusted_payload=untrusted_payload
Mar 19 17:31:15.017096 dom0 qubesd[4569]:         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Mar 19 17:31:15.017096 dom0 qubesd[4569]:     )
Mar 19 17:31:15.017096 dom0 qubesd[4569]:     ^
Mar 19 17:31:15.017096 dom0 qubesd[4569]:   File "/usr/lib/python3.13/site-packages/qubes/api/admin.py", line 947, in vm_start
Mar 19 17:31:15.017096 dom0 qubesd[4569]:     await self.dest.start()
Mar 19 17:31:15.017096 dom0 qubesd[4569]:   File "/usr/lib/python3.13/site-packages/qubes/vm/dispvm.py", line 426, in start
Mar 19 17:31:15.017096 dom0 qubesd[4569]:     await super().start(**kwargs)
Mar 19 17:31:15.017096 dom0 qubesd[4569]:   File "/usr/lib/python3.13/site-packages/qubes/vm/qubesvm.py", line 1522, in start
Mar 19 17:31:15.017096 dom0 qubesd[4569]:     await self.fire_event_async(
Mar 19 17:31:15.017096 dom0 qubesd[4569]:         "domain-start", start_guid=start_guid
Mar 19 17:31:15.017096 dom0 qubesd[4569]:     )
Mar 19 17:31:15.017096 dom0 qubesd[4569]:   File "/usr/lib/python3.13/site-packages/qubes/events.py", line 234, in fire_event_async
Mar 19 17:31:15.017096 dom0 qubesd[4569]:     sync_effects, async_effects = self._fire_event(
Mar 19 17:31:15.017096 dom0 qubesd[4569]:                                   ~~~~~~~~~~~~~~~~^
Mar 19 17:31:15.017096 dom0 qubesd[4569]:         event, kwargs, pre_event=pre_event
Mar 19 17:31:15.017096 dom0 qubesd[4569]:         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Mar 19 17:31:15.017096 dom0 qubesd[4569]:     )
Mar 19 17:31:15.017096 dom0 qubesd[4569]:     ^
Mar 19 17:31:15.017096 dom0 qubesd[4569]:   File "/usr/lib/python3.13/site-packages/qubes/events.py", line 169, in _fire_event
Mar 19 17:31:15.017096 dom0 qubesd[4569]:     effect = func(self, event, **kwargs)
Mar 19 17:31:15.017096 dom0 qubesd[4569]: TypeError: DispVM.on_domain_started_dispvm() got an unexpected keyword argument 'start_guid'
```

marmarek:
And this:
```
Mar 19 17:31:05.021673 dom0 qubesd[4569]: vm.sys-firewall: Activating the sys-firewall VM
Mar 19 17:31:05.040131 dom0 qubesd[4569]: Uncaught exception from domain-unpaused handler for domain sys-firewall
Mar 19 17:31:05.040131 dom0 qubesd[4569]: Traceback (most recent call last):
Mar 19 17:31:05.040131 dom0 qubesd[4569]:   File "/usr/lib/python3.13/site-packages/qubes/app.py", line 1624, in _domain_event_callback
Mar 19 17:31:05.040131 dom0 qubesd[4569]:     vm.fire_event("domain-unpaused")
Mar 19 17:31:05.040131 dom0 qubesd[4569]:     ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
Mar 19 17:31:05.040131 dom0 qubesd[4569]:   File "/usr/lib/python3.13/site-packages/qubes/events.py", line 200, in fire_event
Mar 19 17:31:05.040131 dom0 qubesd[4569]:     sync_effects, async_effects = self._fire_event(
Mar 19 17:31:05.040131 dom0 qubesd[4569]:                                   ~~~~~~~~~~~~~~~~^
Mar 19 17:31:05.040131 dom0 qubesd[4569]:         event, kwargs, pre_event=pre_event
Mar 19 17:31:05.040131 dom0 qubesd[4569]:         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Mar 19 17:31:05.040131 dom0 qubesd[4569]:     )
Mar 19 17:31:05.040131 dom0 qubesd[4569]:     ^
Mar 19 17:31:05.040131 dom0 qubesd[4569]:   File "/usr/lib/python3.13/site-packages/qubes/events.py", line 169, in _fire_event
Mar 19 17:31:05.040131 dom0 qubesd[4569]:     effect = func(self, event, **kwargs)
Mar 19 17:31:05.040131 dom0 qubesd[4569]: TypeError: DispVM.on_domain_unpaused() takes 1 positional argument but 2 were given
```

marmarek:
> * TC_04_DispVM: [test_002_cleanup](https://openqa.qubes-os.org/tests/133233#step/TC_04_DispVM/1) (error)
>   `{message : RuntimeWarning("coroutine 'DispVM.on_domain_unpaused' wa...`
> 
> * TC_04_DispVM: [test_003_cleanup_destroyed](https://openqa.qubes-os.org/tests/133233#step/TC_04_DispVM/2) (error)
>   `{message : RuntimeWarning("coroutine 'DispVM.on_domain_unpaused' wa...`

This and other similar failures here look like missing `await` somewhere.

qubes/tests/vm/mix/dvmtemplate.py
```diff
@@ -0,0 +1,173 @@
...
+        self.domains[vm.name] = vm
+        self.domains[vm] = vm
+        return vm
+
+
+## TODO: finish
+class TC_00_DVMTemplateMixin(
+    qubes.tests.vm.qubesvm.QubesVMTestsMixin,
+    qubes.tests.QubesTestCase,
+    qubes.tests.TestEmitter,
```
marmarek, line 59: The last one is not supposed to be used for the test case class, but rather for test VM/App objects.

ben-grande: Fixed that part, but the remainder of the text is incomplete.

ben-grande:
PipelineRetryFailed

ben-grande:
PipelineRetryFailed

ben-grande:
PipelineRetryFailed

ben-grande:
PipelineRetryFailed

ben-grande:
PipelineRetryFailed

ben-grande:
PipelineRetryFailed

qubes/tests/integ/dispvm.py
```diff
@@ -221,6 +225,118 @@ def test_010_simple_dvm_run(self):
         finally:
             self.loop.run_until_complete(dispvm.cleanup())
 
+    def test_011_dvm_run_preload_invalid_max(self):
+        # TODO: couldn't make the assert raise on unit tests.
+        cases_invalid = ["a", "-1", "1 1"]
+        for value in cases_invalid:
+            with self.subTest(value=value):
+                with self.assertRaises(qubes.exc.QubesValueError):
+                    self.appvm.features["preload-dispvm-max"] = value
```
marmarek, line 234: Did you mean `self.disp_base` instead of `self.appvm`? Same applies to few more cases below.

qubes/tests/integ/dispvm.py
```diff
@@ -221,6 +225,118 @@ def test_010_simple_dvm_run(self):
...
+        try:
+            self.assertEqual(self.disp_base.get_feat_preload(), [dispvm.name])
+            self.assertTrue(dispvm.is_preloaded())
+            self.assertTrue(dispvm.features.get("internal", False))
+            self.assertTrue(dispvm.is_paused())
+            # TODO: how to assert call was made to the desired qube?
+            # I saw qubes-core-admin-client/qubesadmin/tests/tools/qvm_run.py
+            # but it has extra things in tests/__init__.py (expected_calls and
+            # service_calls_).
+            ret = qvm_run.main(self.qvm_run_preload_args, app=self.app)
```
marmarek, line 279: This _will_ deadlock... Anything that wants to interact with qubesd needs to be called via asyncio, otherwise qubesd won't respond. In practice, it's probably easier to call this via asyncio subprocess instead of import + function call, as for the latter you'd need a separate thread and there are several pitfalls with threads.

qubes/tests/integ/dispvm.py
```diff
@@ -221,6 +225,118 @@ def test_010_simple_dvm_run(self):
...
+        dispvm = self.loop.run_until_complete(
+            qubes.vm.dispvm.DispVM.from_appvm(self.disp_base, preload=True)
+        )
+        self.disp_base.features["preload-dispvm-max"] = False
+        try:
+            self.assertEqual(self.disp_base.get_feat_preload(), [dispvm.name])
+            self.assertTrue(dispvm.is_preloaded())
+            self.assertTrue(dispvm.features.get("internal", False))
+            self.assertTrue(dispvm.is_paused())
+            # TODO: how to assert call was made to the desired qube?
```
marmarek, line 275: Register event handler on that object? You can do that not only via decorator, but also via `add_handler` function. So, something like (untested):
```python
test_ok = False
def test_handler(*argfs, **kwargs):
    test_ok = Trhe
dispvm.add_handler('domain-unpause', test_handler)
...
self.assertTrue(test_ok)
```

Alternatively, adjust qvm-run params to print hostname or something like this then capture it and  compare if you got expected value.

qubes/tests/integ/dispvm.py
```diff
@@ -221,6 +225,118 @@ def test_010_simple_dvm_run(self):
...
+            # I saw qubes-core-admin-client/qubesadmin/tests/tools/qvm_run.py
+            # but it has extra things in tests/__init__.py (expected_calls and
+            # service_calls_).
+            ret = qvm_run.main(self.qvm_run_preload_args, app=self.app)
+            self.assertEqual(ret, 0)
+            self.assertEqual(self.disp_base.get_feat_preload(), [])
+            # TODO: these tests are too late, when qvm_run returns, the qube
```
marmarek, line 282: If you change the call to use not qubes.GetDate, but something that waits for input (even qubes.VMShell), then you can have `qvm-run -p` started (via asyncio, as noted above), check what you need and only then send EOF to finalize the service call.
Note that starting the process doesn't mean you instantly get the dispvm, but if you go with the hostname idea, you can wait for the service to print it, then do checks, and then send EOF.

qubes/vm/mix/dvmtemplate.py
```diff
@@ -34,6 +38,88 @@ class DVMTemplateMixin(qubes.events.Emitter):
         doc="Should this VM be allowed to start as Disposable VM",
     )
 
+    def get_feat_preload(self) -> list:
+        feature = "preload-dispvm"
+        # TODO: fix, mypy throws:
+        # error: "DVMTemplateMixin" has no attribute "features"  [attr-defined]
```
marmarek, line 44: mypy doesn't understand mixins... there are two options:
- add `assert isinstance(self, BaseVM)` (or similar)
- ignore (as you did now)

qubes/vm/dispvm.py
```diff
@@ -187,11 +189,131 @@ def __init__(self, app, xml, *args, **kwargs):
...
+                self.run_service_for_stdio(
+                    "qubes.WaitForSession",
+                    user=self.default_user,
+                    stdout=subprocess.DEVNULL,
+                    stderr=subprocess.DEVNULL,
+                ),
+                timeout=gui_timeout,
+            )
+        except asyncio.TimeoutError:
+            ## TODO: should timeout be treated as an error/qubes.exc?
```
marmarek, line 294: Throwing an exception from this event handler, will cause VM to be killed. I'm not sure if that's a good idea, especially with hardcoded timeouts. If not killing, it will pause a VM that potentially is still in starting, right? If that's harmless then just log a message but don't kill. But if that causes some issues (for example, systemd on unpause considers various service start timeouts already expired), then probably better kill and maybe make timeout adjustable (or derive from qrexec_timeout property?)

qubes/tests/vm/dispvm.py
```diff
@@ -45,7 +46,7 @@ def add_new_vm(self, cls, **kwargs):
         return vm
 
 
-class TC_00_DispVM(qubes.tests.QubesTestCase):
+class TC_00_DispVM(qubes.tests.QubesTestCase, qubes.tests.TestEmitter):
```
marmarek, line 49: TestEmitter isn't supposed to be used for testcase class.

qubes/tests/vm/dispvm.py
```diff
@@ -69,6 +70,15 @@ def setUp(self):
         self.app.domains[self.appvm.name] = self.appvm
         self.app.domains[self.appvm] = self.appvm
         self.addCleanup(self.cleanup_dispvm)
+        self.emitter = qubes.tests.TestEmitter()
+        # TODO: Ben: cleanup
+        # self.appvm_emitter = qubes.tests.TestEmitter()
+        # self.app.domains[self.appvm].fire_event = self.emitter.fire_event
```
marmarek, line 76: This should work for checking if events are fired, but won't fire actual handlers. But looking at the unit test results, it seems you've got actual event handlers working too:
```
qubes.tests.vm.dispvm/TC_00_DispVM/test_000_from_appvm_preload_only ... ok
started
ERROR (QubesValueError: QubesValueError("Invalid preload-dispvm value: non qube(s): 'disp42'"))
ERROR (QubesValueError: Invalid preload-dispvm value: non qube(s): 'disp42')
qubes.tests.vm.dispvm/TC_00_DispVM/test_000_from_appvm_preload_reject_max ... started
ok
```

qubes/tests/integ/dispvm.py
ben-grande: Fixed.

marmarek:
Looks like unit tests hanged, specifically `qubes.tests.api_admin/TC_00_VMs/test_643_vm_create_disposable_preload_autostart ...` (or the one after?)

qubes/tests/run.py
```diff
@@ -303,10 +304,19 @@ def test_6_unexpected_success(self):
     return runner.run(suite).wasSuccessful()
 
 
+epilog="""\
+When running only specific tests, write their names like in the log format:
+MODULE+"/"+CLASS+"/"+FUNCTION.
+Example: qubes.tests.basic/TC_00_Basic/test_000_create
+
+Some integration tests require the environment variable QUBES_TEST_TEMPLATES to
```
marmarek, line 312: This is not true, none requires it (and if something requires this variable set, it's a bug that should never hit the main branch). You can use this variable to limit on what templates test are run (but when listing specific tests, it isn't really necessary as test names will include template name already, so you can select template this way instead).

ben-grande: Tests fails to load and be listed when QUBES_TEST_TEMPLATES is not set:

```
$ grep _FailedTest /tmp/out
cli (unittest.loader._FailedTest.cli)
qubes.tests.integ.basic (unittest.loader._FailedTest.qubes.tests.integ.basic)
qubes.tests.integ.grub (unittest.loader._FailedTest.qubes.tests.integ.grub)
qubes.tests.integ.devices_block (unittest.loader._FailedTest.qubes.tests.integ.devices_block)
qubes.tests.integ.qrexec (unittest.loader._FailedTest.qubes.tests.integ.qrexec)
qubes.tests.integ.qrexec_perf (unittest.loader._FailedTest.qubes.tests.integ.qrexec_perf)
qubes.tests.integ.dom0_update (unittest.loader._FailedTest.qubes.tests.integ.dom0_update)
qubes.tests.integ.vm_update (unittest.loader._FailedTest.qubes.tests.integ.vm_update)
qubes.tests.integ.network (unittest.loader._FailedTest.qubes.tests.integ.network)
qubes.tests.integ.network_ipv6 (unittest.loader._FailedTest.qubes.tests.integ.network_ipv6)
qubes.tests.integ.dispvm (unittest.loader._FailedTest.qubes.tests.integ.dispvm)
qubes.tests.integ.vm_qrexec_gui (unittest.loader._FailedTest.qubes.tests.integ.vm_qrexec_gui)
qubes.tests.integ.audio (unittest.loader._FailedTest.qubes.tests.integ.audio)
qubes.tests.integ.mime (unittest.loader._FailedTest.qubes.tests.integ.mime)
qubes.tests.integ.salt (unittest.loader._FailedTest.qubes.tests.integ.salt)
qubes.tests.integ.backup (unittest.loader._FailedTest.qubes.tests.integ.backup)
qubes.tests.integ.backupdispvm (unittest.loader._FailedTest.qubes.tests.integ.backupdispvm)
```

marmarek: And that is a problem to fix, not to workaround by setting this variable. Tests that fails to load this way is a blocking failure, I won't merge a PR that introduces such regression.

Try to import one of those modules in python interpreter manually and see what exactly fails.

ben-grande: Seems to be fixed now, will remove the comment.

qubes/api/internal.py
```diff
@@ -113,6 +115,7 @@ def get_system_info(cls, app):
         system_info = {
             "domains": {
                 domain.name: {
+                    "features": dict(domain.features.items()),
```
marmarek, line 118: You setup events only for "internal" feature, not all of them, so changing others will lead to an outdated cache. I know others are not used right now, but it's easy to miss this when somebody would want to use it. I see two solutions:
- add events to invalidate cache on other feature changes too
- include only `internal` feature here (maybe even as a single entry, not a `features` dict, to not mislead that the list is complete; or maybe `some-feature` or something)

I slightly prefer the second option, as it will invalidate cache less often, but I'm okay with the first option too.

ben-grande: I went for the second approach.

ben-grande:
PipelineRetryFailed

ben-grande:
PipelineRetry

qubes/tests/api_admin.py
```diff
@@ -3787,6 +3806,101 @@ def test_642_vm_create_disposable_not_allowed(self, storage_mock):
...
+
+    @unittest.mock.patch("qubes.vm.dispvm.DispVM.start")
+    @unittest.mock.patch("qubes.storage.Storage.create")
+    def test_643_vm_create_disposable_preload_autostart(
+        self, mock_storage, mock_dispvm_start
+    ):
+        mock_storage.side_effect = self.dummy_coro
+        mock_dispvm_start.side_effect = self.dummy_coro
+        self.vm.template_for_dispvms = True
+        self.app.domains[self.vm].fire_event = self.emitter.fire_event
```
marmarek, line 3839: `fire_event_async`
? but then it will suppress the actual handlers so not dispvm will be preloaded...

ben-grande: I will try with that and with handlers.

marmarek (changes requested):
There should be a high-level documentation of this feature somewhere. Maybe a docstring of the DispVM class? or `doc/qubes-vm/dispvm.rst`? It should explain the lifetime of a preloaded dispvm and point a reader (developer) where to look for the implementation of individual steps.

And also, git history will want a cleanup at some point. At least a few initial commits should be squashed together. The ideal situation is that `git blame` points at the commit explaining why a given line was added/changed, not a chain of several fixes of that line within a single PR. But also there shouldn't be unrelated changes in a single commit. I'm not expecting it to be ideal, but there is a bit of simple cleanup that can improve this PR significantly.

qubes/vm/mix/dvmtemplate.py
```diff
@@ -34,6 +38,131 @@ class DVMTemplateMixin(qubes.events.Emitter):
...
+            return
+
+        # New value can be bigger than maximum permitted as long as it is
+        # smaller than its old value.
+        if new_len > max(preload_dispvm_max, old_len):
+            raise qubes.exc.QubesValueError(
+                f"{error_prefix} can't increment: qube count ({new_len}) "
+                f"is bigger than old count ({old_len}) and also bigger than "
+                f"preload-dispvm-max ({preload_dispvm_max})"
+            )
```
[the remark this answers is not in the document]
ben-grande: Added as a note of todos.

qubes/vm/dispvm.py
```diff
@@ -251,10 +382,39 @@ async def from_appvm(cls, appvm, **kwargs):
...
+                    app.save()
+                    return dispvm
+                await asyncio.sleep(0.1)
+
         dispvm = app.add_new_vm(
             cls, template=appvm, auto_cleanup=True, **kwargs
         )
         await dispvm.create_on_disk()
+        if preload:
+            await dispvm.preload()
```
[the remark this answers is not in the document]
ben-grande: Added as a note of todos.

qubes/vm/mix/dvmtemplate.py
```diff
@@ -34,6 +38,131 @@ class DVMTemplateMixin(qubes.events.Emitter):
...
+            self.features["preload-dispvm"] = " ".join(new_preload or [])
+            for unwanted_disp in excess:
+                if unwanted_disp in self.app.domains:
+                    dispvm = self.app.domains[unwanted_disp]
+                    asyncio.ensure_future(dispvm.cleanup())
+        clean_preload = new_preload
+        for unwanted_disp in new_preload:
+            if unwanted_disp not in self.app.domains:
+                clean_preload.remove(unwanted_disp)
+        if clean_preload < new_preload:
```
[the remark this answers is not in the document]
marmarek: I don't think it does what you think it does:
```python
>>> [1,1,1,1] > [2,2]
False
```

qubes/vm/mix/dvmtemplate.py
```diff
@@ -34,6 +38,131 @@ class DVMTemplateMixin(qubes.events.Emitter):
...
+        return False
+
+    def remove_preload_excess(self, max_preload=None):
+        if max_preload is None:
+            max_preload = self.get_feat_preload_max()
+        max_preload = int(max_preload)
+        old_preload = self.get_feat_preload()
+        if not old_preload:
+            return
+        new_preload = old_preload[:max_preload]
```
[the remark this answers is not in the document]
marmarek: Mostly concern about OS getting unstable after a long pause (large time jump, maybe a host suspend in the meantime etc). Keeping fresher DispVMs seems more reliable. But on the other hand, this case is rare anyway (only on configuration change), so maybe it doesn't matter.

qubes/tests/api_admin.py
ben-grande: `fire_event_async` does suppress the actual handlers, but `fire_event`... it is suppressing something that lets the test work.

```
        self.vm.features["preload-dispvm-max"] = "1"
        retval = self.call_mgmt_func(
            b"admin.vm.CreateDisposable", b"dom0", arg=b"preload-autostart"
        )
```

If I remove the `fire_event` suppression, this fails as it preloads 2 qubes. I don't understand how that happens, because if I sleep:

```
        self.vm.features["preload-dispvm-max"] = "1"
        import time
        time.sleep(10)
        retval = self.call_mgmt_func(
            b"admin.vm.CreateDisposable", b"dom0", arg=b"preload-autostart"
        )
```

It still happens. No qube is preloaded after changing `preload-dispvm-max` no matter the delay, but when it runs the call to the admin API, it tries to preload 2 qubes. Only happens on unittest.

On another note, I found a way to check if the event `"domain-preloaded-dispvm-autostart"` was fired.

marmarek: > `fire_event_async` does suppress the actual handlers, but `fire_event`... it is suppressing something that lets the test work.

What `fire_event_async` suppresses? Maybe you forgot `await` when using it somewhere?

ben-grande: I commented the code for you to see on the next push.

qubes/vm/mix/dvmtemplate.py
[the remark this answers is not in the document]
ben-grande: Done.

qubes/vm/dispvm.py
```diff
@@ -187,11 +189,130 @@ def __init__(self, app, xml, *args, **kwargs):
...
+            raise qubes.exc.QubesException(
+                "Failed to run QUBESRPC qubes.WaitForSession during preload "
+                "startup"
+            )
+        finally:
+            # TODO: ben: this finally block is harmful, pausing a qube that
+            # couldn't run the test services doesn't help to later try to use
+            # them, instead, they should be purged from the list, not marked as
+            # used to not create an infinite loop, but just cleanup() and
+            # remove from the list.
```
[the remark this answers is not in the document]
ben-grande: Fixed by removing the finally statement, exceptions that raise there will trigger `domain-start-failed` and later on `cleanup()`, `domain-shutdown` and it will remove the list without marking it as used.

qubes/vm/dispvm.py
```diff
@@ -187,11 +189,130 @@ def __init__(self, app, xml, *args, **kwargs):
             self.features.update(template.features)
             self.tags.update(template.tags)
 
+    def is_preloaded(self) -> bool:
```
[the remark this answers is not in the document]
ben-grande: Done.

qubes/tests/integ/dispvm.py
```diff
@@ -221,6 +241,163 @@ def test_010_simple_dvm_run(self):
...
+
+    async def _test_012_dvm_run_preload_race_more(self):
+        self.disp_base.features["preload-dispvm-max"] = "5"
+        for _ in range(100):
+            if len(self.disp_base.get_feat_preload()) == 5:
+                break
+            await asyncio.sleep(1)
+        last_disp_name = self.disp_base.get_feat_preload()[4]
+        last_disp = self.app.domains[last_disp_name]
+        for _ in range(50):
```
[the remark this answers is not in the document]
ben-grande: Done.

qubes/vm/mix/dvmtemplate.py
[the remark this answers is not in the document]
ben-grande: I think it doesn't matter here because it only does that when trimming (reducing the maximum allowed). The only effective trim for that issue is using `remove_preload_excess(0)` (triggered by the `autostart` event), exposed to the Admin API as `admin.vm.CreateDisposable+preload-autostart`.

Other than that, I don't think that reducing the number of preloaded disposables should care about large time jumps on long pause. I think this issue is similar to how to later preload qubes when it couldn't preload now because of not having enough memory and should be resolved in the same way.

ben-grande:
> And also, git history will want a cleanup at some point

Understood. I am doing multiple unclean commits because it is easier to revert something or to review a specific change with quick iterations. By the end, I will probably squash them all together because most of the history was not a fully working setup that had to change multiple times, unless I see that there is a clear distinction that I should make to create new commits or you say that 1 commit with many line changes is not something you want.

qubes/vm/mix/dvmtemplate.py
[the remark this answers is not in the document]
ben-grande: > and maybe other places?

I don't see other places impacting that. The `await dispvm.create_on_disk()` problem was a nice catch that I didn't thought of before.

qubes/vm/mix/dvmtemplate.py
```diff
@@ -68,6 +197,65 @@ def __on_property_set_template(self, event, name, newvalue, oldvalue=None):
...
+            memory = getattr(self, "memory", 0) * 1024 * 1024
+            unrestricted_preload = int(available_memory / memory)
+            can_preload = min(unrestricted_preload, want_preload)
+            if skip_preload := want_preload - can_preload:
+                self.log.warning(
+                    "Not preloading '%d' disposable(s) due to insufficient "
+                    "memory",
+                    skip_preload,
+                )
+            if can_preload == 0:
```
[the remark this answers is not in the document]
ben-grande: Your recommendations were fruitful, now it tries to preload:

1. When a preloaded is used or
2. When it asks to use a disposable without asking for a preload

I didn't add on the closing/shutdown of the preloaded disposable because of the intended behavior I want for the shutdown:

1. If the qube is preloaded and shuts down, user don't want qubes to be preloaded anymore.
2. If the qube was preloaded and shuts down, it was already marked as used before.

qubes/vm/dispvm.py
```diff
@@ -39,7 +41,86 @@ def _setter_template(self, prop, value):
...
+    distinction is made.
+
+    **Design**:
+
+    - **Fast**: No delays, skip interrupt (pause or suspend) if qube is
+      requested before it.
+
+    - **Easy**: Bootstrap must only require a single feature to be set,
+      *preload-dispvm-max*, while usage must be transparent, as in the same
+      workflow to use unnamed disposables.
```
marmarta, line 61:
```suggestion
    - **Fast**: no delays from user perspective. Pause/suspend should be skipped 
    if this qube is requested before interrupt can be performed.

    - **Easy-to-use**: preloading requires a single feature (*preload-dispvm-max*) 
    and its use must be transparent, indistinguishable from working with 
    normal (non-preloaded) unnamed disposable qubes.
```

qubes/vm/dispvm.py
```diff
@@ -39,7 +41,86 @@ def _setter_template(self, prop, value):
...
+
+      - Must be able to resume from an invalid list when starting, the event
+        must occur even if the list is outdated by cleaning it up prior to
+        returning a qube from it.
+
+      - Cloning, renaming or deriving a qube from a template must not clone
+        properties that would be invalid on the target.
+
+      - Interruption or usage of the qube must only happen after all basic
+        services have been started on the qube.
```
marmarta, line 81:
```suggestion
    - **Reliable**:

      - Avoid race conditions and use-before-setup: preloading and marking a 
      preloaded qube as used must be synchronous

     - Recovery from failed start: the system must attempt to preload qubes even if
     previous pre-loading attempts failed, regardless of whether preloaded disposable
     qubes have been requested on this session. If current qube list is invalid, it should 
     be cleaned up.
     
      - No qube operation (in particular cloning, renaming or creating a standalone based 
      on a template) should result in properties that are invalid on the target

      - Full start: a preloaded disposable can only be interrupted (paused/suspended) 
      or used after all basic services in it have started 
```

qubes/vm/dispvm.py
```diff
@@ -39,7 +41,86 @@ def _setter_template(self, prop, value):
...
+
+    - **Prevents misuse**:
```
marmarta, line 83:
```suggestion
    - **Prevents accidental tampering**:
```

qubes/vm/dispvm.py
```diff
@@ -39,7 +41,86 @@ def _setter_template(self, prop, value):
...
+        application that can list a preloaded qube must have an event handler
+        for the *internal* feature and optionally the *is_preload* property,
+        reacting to the setting of those attributes to *True* with hiding the
+        qube itself from being listed or it's applications from being clickable.
+
+      - The interruption allows having an event to handle its end (unpause or
+        resume), which must mark the qube as used, even if the qube has not been
+        requested for use. Use of interruption for disabling scheduling to the
+        hypervisor is a second thought and must not be favored over it's benefit
+        of detecting when a qube is used.
```
marmarta, line 97:
```suggestion
      - Preloaded qubes have the *internal* feature set when they are created. This
      feature hides the qube from GUI tools and discourages user tampering. It
      is unset when the qube is resumed and starts being in use.  Remember to 
      validate if all GUI applications correctly react to setting and removing the
      *internal* feature (optionally, the *is_preload* property can be helpful)

      - If the qube is unpaused/resumed, it should always be marked as used, even
      if it was not requested. The goal of pause/suspend in case of preloaded disposables
      is mostly detecting whether a qube was used or not, and not managing resource 
      consumption; thus, even with abundant system memory, they should not be unpaused
      or resumed without being requested.
```

qubes/vm/dispvm.py
```diff
@@ -39,7 +41,86 @@ def _setter_template(self, prop, value):
...
+
+      - **Startup**: Awaits basic qube services to be started followed with
+        domain interruption.
+
+      - **Request**: Qube is removed from the preload list, in case the
+        *startup* did not reach interruption yet, that stage is skipped.
+
+    - **Used**: Interruption has ended if applicable, qube is marked as used,
+      GUI applications treats the qube as any other unnamed disposable and the
+      qube object is returned to the caller if requested.
```
marmarta, line 112:
```suggestion
    **Stages**:

    - **Preload**: The qube is created and marked as preloaded. Qube is not visible
    in GUI applications.

    - **Startup**: begins qube startup, starts normal qube services, attempts to 
    interrupt (suspend/pause); the qube is not visible in GUI applications

     - **Request**: the qube is removed from preload list. If *startup* has not yet reached
     interrupt, interrupting is skipped. The qube is not visible in GUI applications

    - **Used**: Qube has been unpaused/resumed (if applicable) and it has been
     marked as used. GUI applications treat the qube as any other unnamed disposable 
     and the  qube object is returned to the caller if requested.
```

qubes/vm/dispvm.py
```diff
@@ -39,7 +41,86 @@ def _setter_template(self, prop, value):
...
+    **Outstanding bugs**:
+
+    - GUI applications set to autostart can appear on the screen before the qube
+      is allowed to be used and also causes UX confusion of why an application
+      appears on the screen but it does not react to GUI events as the qube is
+      interrupted.
+    - Can't interrupt qubes that have GUI applications before the GUI session
+      has started as the GUI daemon does not recognize an interrupted qube to
+      treat it differently.
+    """
```
marmarta, line 123:
```suggestion
    **Outstanding bugs**:

    - GUI applications set to autostart can appear on the screen before the qube
      is allowed to be used (those applications will be non-interactive 
      due to interruption)
    - Can't interrupt qubes that have GUI applications running before the GUI session
      has started (GUI daemon cannot handle an interrupted qube).
    """
```

marmarta (commented):
I've done some rewriting the docstring to be clearer; you tend to have really long sentences that can be a bit confusing. Please check if I have not misunderstood the contents tho, my understanding here is deeply imperfect

qubes/vm/dispvm.py
ben-grande: > thus, even with abundant system memory,

 Changed to *system resources* because `pause` "blocks" CPU scheduling but not memory.

qubes/vm/dispvm.py
ben-grande: 1. What is your preferred of capital letters or not after the sections?
2. What is your preferred use of use *the* to refer to the qube?

Example:

```markdown
- **Preload**: The qube
- **Startup**: begins qube
- **Request**: the qube
- **Used**: Qube
```

ben-grande: > The qube is not visible in GUI applications

Do you think it is important to mention this on all stages?

qubes/vm/dispvm.py
ben-grande: > those applications will be non-interactive due to interruption

Unfortunately, there is a brief moment where the user can interact with the applications before it is interrupted, which may be a cause of confusion for the user as suddenly the GUI will freeze/pause.

ben-grande: > have GUI applications running

It has to do with any GUI application not being able to open, it does not matter if they are running before the qube is paused, also they can't possibly show as there is no DISPLAY yet.

Changed the text to be simpler.

ben-grande (commented):
Please let me know what you think of the documentation fixes.

ben-grande:
>  Please check if I have not misunderstood the contents tho, my understanding here is deeply imperfect

You understand most of it, all the important parts, so that is great.

qubes/vm/dispvm.py
marmarta: you're right, in **used** it should be "the qube"; in startup, 'qube' is not an independent noun, it's used as a modifier for startup, which feels like it should have a 'the' because it's a general phase, not a specific instance of it.

I mention the GUI visibility because it feels like an important distinction between phases, but I think it's also fine to only put it in the used phase, maybe with additional "Starting from this moment" or "Only in this phase"

qubes/vm/dispvm.py
marmarta: then maybe "(those applications will be interactive only for a brief moment and then suddenly freeze)"?

marmarta:
docs look good :)

qubes/vm/dispvm.py
ben-grande: Fixed.

ben-grande:
> docs look good :)

Would you like more?

- https://github.com/QubesOS/qubes-core-admin-client/pull/354]

Your inputs were valuable.

DemiMarie:
Does preloading a qube block the qubesd event loop?  That’s a really bad idea.  Preloading a qube should happen asynchronously, but the qube being preloaded should be marked as unusable until it is ready, and attempting to use a qube being preloaded should wait (asynchronously) until the qube is fully ready.

marmarek:
> Does preloading a qube block the qubesd event loop?  

It doesn't.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

qubes/vm/dispvm.py
```diff
@@ -243,18 +460,71 @@ async def from_appvm(cls, appvm, **kwargs):
...
+            # As well help informing that unpause/resume was done before
+            # requesting the qube.
+            appvm.remove_preload_from_list([dispvm.name])
+            dispvm.features["preload-dispvm-request"] = True
+            tries, sleep = 1200, 0.1
+            for _ in range(tries):
+                if dispvm.features.get(
+                    "preload-dispvm-skip-interrupt", None
+                ) or (paused := dispvm.is_paused()):
+                    if paused:
```
marmarek, line 502: If you get here with just the first condition true (unlikely, but possible), `paused` will be undefined, or contains value from previous loop iteration. You should initialize it first.

But also, this loop isn't nice... Maybe better use asyncio features for it, for example an event that is woken up if `preload-dispvm-request` is set. Something like:
```
if not dispvm.is_paused():
    dispvm.preload_finished = asyncio.Event()
    await dispvm.preload_finished.wait()
else:
    await dispvm.unpause()
```
and instead of setting `preload-dispvm-skip-interrupt`, do `dispvm.preload_finished.set()`.

You may need to add something like `self.preload_finished = None` in `__init__` of this class. And maybe use `preload_finished` being set as a signal to not pause it (instead of the `preload-dispvm-request` feature)? But that I'm not sure.

qubes/tests/integ/dispvm.py
```diff
@@ -221,6 +240,240 @@ def test_010_simple_dvm_run(self):
...
+            self.loop.run_until_complete(asyncio.sleep(1))
+        else:
+            self.fail("didn't preload in time")
+        old_preload = self.disp_base.get_feat_preload()
+        proc = self.loop.run_until_complete(
+            asyncio.create_subprocess_exec("/usr/lib/qubes/preload-dispvm")
+        )
+        self.loop.run_until_complete(asyncio.wait_for(proc.wait(), timeout=30))
+        preload_dispvm = self.disp_base.get_feat_preload()
+        self.loop.run_until_complete(proc.wait())
```
marmarek, line 469: Since you added `proc.wait()` above already, this isn't needed here. If the above would hit a timeout, it won't reach this line anyway, since `wait_for` raises `TimeoutError`.

qubes/vm/qubesvm.py
```diff
@@ -406,13 +406,29 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.LocalVM):
             :param subject: Event emitter (the qube object)
             :param event: Event name (``'domain-start-failed'``)
 
+        .. event:: domain-pre-paused (subject, event)
+
+            Fired at the beginning of :py:meth:`paused` method and before
```
marmarek, line 411:
```suggestion
            Fired at the beginning of :py:meth:`pause` method and before
```

qubes/vm/qubesvm.py
```diff
@@ -406,13 +406,29 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.LocalVM):
...
+            ``libvirt_domain.suspend()``.
+
```
marmarek, line 413: Since all places where this (and the next) event is fired is aN async function, better allow async handlers too.

qubes/vm/qubesvm.py
```diff
@@ -406,13 +406,29 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.LocalVM):
...
             Fired when the domain has been paused.
 
             :param subject: Event emitter (the qube object)
             :param event: Event name (``'domain-paused'``)
 
+        .. event:: domain-pre-unpaused (subject, event)
+
+            Fired at the beginning of :py:meth:`unpaused` method and before
+            ``libvirt_domain.resume()``.
+
```
marmarek, line 428: See above about async.

qubes/vm/qubesvm.py
```diff
@@ -1365,7 +1381,7 @@ async def start(
                     f"Qube start is prohibited. Rationale: {prohibit_rationale}"
                 )
 
-            self.log.info("Starting {}".format(self.name))
+            self.log.info("Starting qube")
```
marmarek, line 1384: AFAIR there are cases where logger name is not easily accessible, so better keep the name, at least in this message.

marmarek (changes requested):

qubes/tests/integ/dispvm.py
ben-grande: Remnants removed.

qubes/tests/integ/dispvm.py
```diff
@@ -221,6 +240,240 @@ def test_010_simple_dvm_run(self):
...
+            stdout=asyncio.subprocess.PIPE,
+        )
+        stdout, _ = await asyncio.wait_for(proc.communicate(), timeout=60)
+        return stdout.decode()
+
+    async def _test_012_run_preload(self):
+        appvm = self.disp_base
+        dispvm = appvm.get_feat_preload()[0]
+        dispvm = self.app.domains[dispvm]
+        self.assertEqual(appvm.get_feat_preload(), [dispvm.name])
```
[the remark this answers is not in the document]
ben-grande: Removed.

qubes/tests/integ/dispvm.py
```diff
@@ -221,6 +240,240 @@ def test_010_simple_dvm_run(self):
...
+        proc = await asyncio.create_subprocess_exec(
+            *self.preload_cmd,
+            stdout=asyncio.subprocess.PIPE,
+        )
+        stdout, _ = await asyncio.wait_for(proc.communicate(), timeout=60)
+        return stdout.decode()
+
+    async def _test_012_run_preload(self):
+        appvm = self.disp_base
+        dispvm = appvm.get_feat_preload()[0]
```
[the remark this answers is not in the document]
ben-grande: Sure! Fixed.

qubes/vm/dispvm.py
ben-grande: > Something like:

Perfect.

> and instead of setting `preload-dispvm-skip-interrupt`, do `dispvm.preload_finished.set()`.

This would complicate the tests, because I can check if the event for setting `preload-dispvm-skip-interrupt` was called, and if it was not, the events `domain-(un)paused` must have been called.

> You may need to add something like `self.preload_finished = None` in `__init__` of this class. And maybe use `preload_finished` being set as a signal to not pause it (instead of the `preload-dispvm-request` feature)?

That can work also well on tests with property events.

---

Will do this in one batch later.

linux/aux-tools/preload-dispvm
```diff
@@ -0,0 +1,28 @@
...
+
+import asyncio
+import qubesadmin
+
+
+def get_max(qube):
+    return int(qube.features.get("preload-dispvm-max", 0) or 0)
+
+
+async def main():
```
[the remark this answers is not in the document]
marmarek: Ah, right, https://github.com/QubesOS/qubes-issues/issues/4719. For now you can workaround it with threads: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor

qubes/vm/dispvm.py
marmarek: > This would complicate the tests, because I can check if the event for setting `preload-dispvm-skip-interrupt` was called, and if it was not, the events `domain-(un)paused` must have been called.

You can check if event was set as you do have that object available in the test, no?

ben-grande: I think I can check for the property being set, yeah, my misinterpretation from before.

linux/aux-tools/preload-dispvm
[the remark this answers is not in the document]
ben-grande: This is fixed.

qubes/api/admin.py
```diff
@@ -1280,25 +1290,35 @@ async def _vm_create(
...
         else:
-            dispvm_template = self.dest
+            appvm = self.dest
 
-        self.fire_event_for_permission(dispvm_template=dispvm_template)
-
-        dispvm = await qubes.vm.dispvm.DispVM.from_appvm(dispvm_template)
+        self.fire_event_for_permission(dispvm_template=appvm)
+        if preload_autostart:
+            asyncio.ensure_future(
```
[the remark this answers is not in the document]
ben-grande: This is fixed.

marmarek:
You probably noticed, but one of the unit tests is not happy; and pylint complains about one too long line...

qubes/vm/mix/dvmtemplate.py
```diff
@@ -34,6 +39,221 @@ class DVMTemplateMixin(qubes.events.Emitter):
...
+        old_preload = self.get_feat_preload()
+
+        # Qubesd restart: Preloading was requested but not delivered.
+        preload_delivery_incomplete = [
+            qube
+            for qube in self.app.domains
+            if getattr(qube, "template", None) == self
+            and qube.features.get("preload-dispvm-requested", False)
+            and not qube.preload_requested
+            and not qube.features.get("preload-dispvm-used", False)
```
marmarek, line 99: Used is simply not on the preload-dispvm list anymore, no? But also, I don't think you need `preload-dispvm-requested` feature - it's used here to distinguish in-flight dispvm requests, but if you make this cleanup run only on qubesd startup (for example domain-load event for a given disposable template), you don't need that check, as you surely don't have any in-flight requests on startup.

Limiting unfortunate qubesd restart cleanup to just qubesd startup time (instead of every dispvm request) would be a good idea anyway.

ben-grande: > Used is simply not on the preload-dispvm list anymore, no?

Yes, not on the list anymore.

> But also, I don't think you need `preload-dispvm-requested` feature - it's used here to distinguish in-flight dispvm requests, but if you make this cleanup run only on qubesd startup (for example domain-load event for a given disposable template), you don't need that check, as you surely don't have any in-flight requests on startup.

That check is to cleanup in-flight requests from previous qubesd startups. The qube may be already running, but as it was requested, it is not on the list anymore and there needs to be a way to know that the qube was requested to not let it linger on the system. It will not occupy the preload list but will occupy system resources and will never be used.

> Limiting unfortunate qubesd restart cleanup to just qubesd startup time (instead of every dispvm request) would be a good idea anyway.

Agreed, implementing it on `domain-load`.

ben-grande:
```python
and qube.features.get("preload-dispvm-requested", False)
and not qube.preload_requested
and not qube.features.get("preload-dispvm-used", False)
```

- Feature `preload-dispvm-requested` being `True` indicates that the qube was once requested.
- With the above plus the property `preload_requested` being `False` indicates that it was not requested in this cycle.
- With the above plus the feature `preload-dispvm-used` being `False` indicates that the qube was never delivered.

If I remove the feature `preload-dispvm-requested`, the remaining checks are not enough to know if the qube was ever requested, it can be that the qube is still being preloaded and was not requested.

marmarek: But if you check that during qubesd startup, any dispvm during preload process likely needs to be cleaned - regardless if it was requested or not. At that point, any in-progress preload would have been from previous startup, and after restart nothing would complete it (pause then done etc).

qubes/vm/mix/dvmtemplate.py
```diff
@@ -34,6 +39,223 @@ class DVMTemplateMixin(qubes.events.Emitter):
...
+                "Removing absent qube(s) from preloaded list: '%s'",
+                ", ".join(absent),
+            )
+            self.features["preload-dispvm"] = " ".join(clean_preload or [])
+
+        # Preloading began but qubesd restarted before preloading completed.
+        preload_creation_incomplete = [
+            self.app.domains[qube]
+            for qube in self.get_feat_preload()
+            if qube in self.app.domains
```
marmarek, line 121: You have `clean_preload` already that is filtered on this condition.

ben-grande: Fixed.

qubes/vm/mix/dvmtemplate.py
ben-grande: > any dispvm during preload process likely needs to be cleaned 

But that is taken care elsewhere.

https://github.com/ben-grande/qubes-core-admin/blob/50e857daa71138c60780887cb40c3bc0b60c6b5e/qubes/vm/mix/dvmtemplate.py#L117

The idea is:

1. `began` preloading but did not reach `complete`, remove
2. Reached `complete` but was not `requested`, kept
3. `requested` but not `used`, remove

marmarek: IIUC the two cases are because you need two different methods of detecting if dispvm was during preloading depending whether it was requested or not (because requested is removed from dispvm-preload list), right? Maybe it could be unified by having a feature like `preload-dispvm-in-progress` that is removed only when preloading is completed (in the same place where you set `preload-dispvm-completed`), regardless of whether it was requested or not? Maybe then just one feature would be enough to handle all the cleanup cases?

qubes/tests/integ/dispvm.py
```diff
@@ -235,8 +235,12 @@ async def run_preload_proc(self):
             *self.preload_cmd,
             stdout=asyncio.subprocess.PIPE,
         )
-        stdout, _ = await asyncio.wait_for(proc.communicate(), timeout=60)
-        return stdout.decode()
+        try:
+            stdout, _ = await asyncio.wait_for(proc.communicate(), timeout=60)
+            return stdout.decode()
+        except asyncio.TimeoutError:
+            proc.terminate()
```
marmarek, line 242: you may want to add `await proc.wait()` after this, to cleanup remains of the process (not doing so may result in zombie processes, which openqa will also report as `cleanup_loop` failure, even if the current issue is caused by something else)

marmarek:
> * TC_20_DispVM_debian-12-xfce: [test_014_dvm_run_preload_nogui](https://openqa.qubes-os.org/tests/141614#step/TC_20_DispVM_debian-12-xfce/5) (error)
>       `qubes.exc.QubesValueError: Qube GUI is 'False' and does not support...`

Uhm, looks like I forgot to include core-agent-linux PR in this test...

marmarek (approved):

PR #661, opened 2025-02-22
Reproduce bug in qrexec tests

DemiMarie:
The qrexec tests are broken and fail intermittently.  This patch makes the bug obvious: by adding a sleep, it causes the test to fail right away, reliably.

### DIFF
--- a/qubes/tests/integ/qrexec.py
+++ b/qubes/tests/integ/qrexec.py
@@ -496,7 +496,7 @@ class TC_00_QrexecMixin(object):
         self.create_remote_file(
             self.testvm2,
             "/etc/qubes-rpc/test.Argument",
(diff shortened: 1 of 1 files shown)


marmarek:
I finally got the correct stderr logged (https://github.com/QubesOS/qubes-core-admin/pull/670) and got this:
```
2025-04-12 18:47:10.164 qrexec-fork-server[1073]: exec.c:165:exec_qubes_rpc2: execve: Exec format error
```
It's a progress I guess...

marmarek:
Ok, this is clearly bug in tests - they create rpc services without shebang, so it works only when called via shell - which not always is the case.

marmarek:
Fix: https://github.com/QubesOS/qubes-core-admin/pull/675

PR #664, opened 2025-02-27
tests: integration tests for boot mode

marmarek:
Right now they mostly duplicate what unit tests are already testing. It
will make more sense for boot mode with in-vm kernel, and/or Whonix
package (instead of dummy boot modes).

QubesOS/qubes-issues#9750

Related to https://github.com/QubesOS/qubes-core-admin/pull/653
FYI @ArrayBolt3


marmarek:
So, the test said `=` is not allowed in feature request values... PR incoming

marmarek:
@ArrayBolt3 I'll also merge this one as is. It did found some issues that unit tests missed before, so IMO it's still useful (even though duplicates some parts). And later it can be extended to test in-vm kernel support.

PR #665, opened 2025-02-27
Allow equal sign and coma in feature requests

marmarek:
It is necessary for boot mode feature, as a common character in
kernelopts

QubesOS/qubes-issues#9750

### DIFF
--- a/qubes/api/misc.py
+++ b/qubes/api/misc.py
@@ -58,7 +58,7 @@ class QubesMiscAPI(qubes.api.AbstractQubesAPI):
             for key in keys
         }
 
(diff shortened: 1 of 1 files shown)


marmarek:
@ArrayBolt3 any other characters missing, that you'd see used in kernelopts for a boot mode?

ArrayBolt3:
Wrote a quick script to tell me all of the characters in Kicksecure's kernel command line, one character that *might* be needed in the future but I don't believe is needed now is the comma (`,`). That's used in arguments like `l1tf=full,force`. We're probably not going to try to cram all of Kicksecure's hardening kernel options into a boot mode, but in the event we need something along these lines in the future, that would be handy.

PR #669, opened 2025-04-02
tests: simple end-to-end remote qrexec test

marmarek:
Test vm1 -> relay -> vm2-remote call path. Do it on a single host by having
relay service that transform names (source to source-remote and 
target-remote to target). For this to work, there need to be matching pairs
of VMs with and without -remote suffix, and appropriate policy for both
outgoing and incoming connections.

QubesOS/qubes-issues#9015

### DIFF
--- a/qubes/tests/__init__.py
+++ b/qubes/tests/__init__.py
@@ -1146,16 +1146,16 @@ class SystemTestCase(QubesTestCase):
                 pass
         # break dependencies
         for vm in vms:
-            if vm.default_dispvm in vms:
-                vm.default_dispvm = None
-            if vm.netvm in vms:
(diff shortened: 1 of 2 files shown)


marmarek:
@fepitre I tried this. It doesn't work yet. First I hit the bug I commented in the qrexec PR. But then the source on remote side is not set - it appears as the connection is coming from the relay, not vm1-remote.

fepitre:
> @fepitre I tried this. It doesn't work yet. First I hit the bug I commented in the qrexec PR. But then the source on remote side is not set - it appears as the connection is coming from the relay, not vm1-remote.

Yes, I did not added the translation to source being the vm calling qrexec originally.

PR #673, opened 2025-04-09
Assure paused domains prior to suspending remain paused

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/1611

### DIFF
--- a/qubes/api/internal.py
+++ b/qubes/api/internal.py
@@ -22,12 +22,16 @@
 
 import asyncio
 import json
+import os
 import subprocess
 
 import qubes.api
 import qubes.api.admin
 import qubes.vm.adminvm
 import qubes.vm.dispvm
+
+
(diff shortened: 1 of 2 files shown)


alimirjamali:
pipelineretry

marmarek:
Paused qube should not be unpaused, even for a short time. It might be paused for several reasons, and allowing it to execute is a bad idea. Yes, it may become quite unhappy after host suspend (mostly due to time jump), but that's why it's called "emergency pause" in the domains widget, to not encourage using it.

alimirjamali:
> Paused qube should not be unpaused, even for a short time.

Yes. I remember it from  #616

I will read the `suspend` and `pause` functions and see the differences.

BTW, any idea on why the CI/CD did not run on this PR?

alimirjamali:
PipelineRetry

marmarek (approved):

PR #680, opened 2025-05-16
Late GUID for preloaded disposables

ben-grande:
- Autostarted applications don't show when preloading; and
- Preloaded qubes paused before the session will have a working GUI when
  the session starts.

For: https://github.com/qubesos/qubes-issues/issues/9940
For: https://github.com/qubesos/qubes-issues/issues/9907
Requires: https://github.com/QubesOS/qubes-core-admin-client/pull/359
Requires: https://github.com/QubesOS/qubes-core-admin/pull/689

### DIFF
--- a/qubes/tests/integ/dispvm.py
+++ b/qubes/tests/integ/dispvm.py
@@ -29,10 +29,21 @@ from distutils import spawn
 from unittest.mock import patch, mock_open
 import asyncio
 import sys
+import logging
 
 import qubes.config
 import qubes.tests
 import qubesadmin.exc
+
+# nose will duplicate this logger.
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+handler = logging.StreamHandler()
+formatter = logging.Formatter(
+    "%(asctime)s: %(levelname)s: %(funcName)s: %(message)s"
+)
+handler.setFormatter(formatter)
+logger.addHandler(handler)
 
 
 class TC_04_DispVM(qubes.tests.SystemTestCase):
@@ -183,11 +194,16 @@ class TC_04_DispVM(qubes.tests.SystemTestCase):
 
 class TC_20_DispVMMixin(object):
     def setUp(self):  # pylint: disable=invalid-name
+        logger.info("start")
         super(TC_20_DispVMMixin, self).setUp()
         if "whonix-g" in self.template:
             self.skipTest(
                 "whonix gateway is not supported as DisposableVM Template"
             )
+        self.app.add_handler("domain-add", self._on_domain_add)
+        self.addCleanup(
+            self.app.remove_handler, "domain-add", self._on_domain_add
+        )
         self.init_default_template(self.template)
         self.disp_base = self.app.add_new_vm(
             qubes.vm.appvm.AppVM,
@@ -201,12 +217,16 @@ class TC_20_DispVMMixin(object):
         self.preload_cmd = [
             "qvm-run",
             "-p",
+            "--no-color-output",
+            "--no-color-stderr",
             f"--dispvm={self.disp_base.name}",
             "--",
             "qubesdb-read /name | tr -d '\n'",
         ]
+        logger.info("end")
 
     def tearDown(self):  # pylint: disable=invalid-name
+        logger.info("start")
         if "gui" in self.disp_base.features:
             del self.disp_base.features["gui"]
         old_preload = self.disp_base.get_feat_preload()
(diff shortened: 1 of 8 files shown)


qubes/vm/dispvm.py
```diff
@@ -341,43 +332,33 @@ async def on_domain_started_dispvm(
...
-        gui = bool(self.guivm and self.features.get("gui", True))
-        service = "qubes.WaitForSession"
-        if not gui:
-            # https://github.com/QubesOS/qubes-issues/issues/9964
-            # service = "qubes.WaitForRunningSystem"
-            rpc = "qubes.WaitForRunningSystem"
-            path = "/run/qubes-rpc:/usr/local/etc/qubes-rpc:/etc/qubes-rpc"
-            service = '$(PATH="' + path + '" command -v ' + rpc + ")"
+        # https://github.com/QubesOS/qubes-issues/issues/9964
+        rpc = "qubes.WaitForRunningSystem"
```
marmarek, line 337: This needs a better handling of old (R4.2) templates. Right now (with this PR), if `qubes.WaitForRunningSystem` is not available, preloading fails always, regardless of GUI or not. For old template, it should either fallback to the previous version (no late-gui approach, and only complain about missing `qubes.WaitForRunningSystem` if there is no gui either), or it should refuse to enable preloading in the first place.

ben-grande: Does the changes on `vm/mix/dvmtemplate.py` deal with this?

marmarek: Ah, right, all good then :)

ben-grande:
> Inclusion of [QubesOS/qubes-core-agent-linux#571](https://github.com/QubesOS/qubes-core-agent-linux/pull/571) is obligatory for all tests now. There was no version bump since [QubesOS/qubes-core-agent-linux@8a5e873](https://github.com/QubesOS/qubes-core-agent-linux/commit/8a5e873a95b6c0ea90f5581803032619b9eb958d), therefore templates didn't have a package to upgrade.

Maybe the idea of this PR https://github.com/QubesOS/qubes-core-agent-linux/pull/578 was for there to be a package upgrade?

marmarek:
> Maybe the idea of this PR [QubesOS/qubes-core-agent-linux#578](https://github.com/QubesOS/qubes-core-agent-linux/pull/578) was for there to be a package upgrade?

Yes, exactly. I needed something to put the `openqa-pending` label on.

qubes/vm/dispvm.py
```diff
@@ -341,43 +332,33 @@ async def on_domain_started_dispvm(
...
         try:
             self.log.info(
                 "Preload startup waiting '%s' with '%d' seconds timeout",
                 service,
                 timeout,
             )
-            runner = self.run_service_for_stdio if gui else self.run_for_stdio
             await asyncio.wait_for(
-                runner(
+                self.run(
```
ben-grande: There is a high chance that this is leaking, changing it to `run_for_stdio` doesn't make it leak and also is the correct one to use... I remember some comment of yours about this, probably on #660, but forgot to integrate in this PR.

ben-grande:
```
# test_013_dvm_run_preload_gui
# failure: 

# timestamp 2025-06-19T21:55:08.880827
    return func(*args, **kwds)
  File "/usr/lib/python3.13/site-packages/qubes/tests/integ/dispvm.py", line 428, in test_013_dvm_run_preload_gui
    self.loop.run_until_complete(self._test_013_dvm_run_preload_gui())
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/asyncio/base_events.py", line 719, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/qubes/tests/integ/dispvm.py", line 436, in _test_013_dvm_run_preload_gui
    await self.run_preload()
  File "/usr/lib/python3.13/site-packages/qubes/tests/integ/dispvm.py", line 352, in run_preload
    self.assertTrue(
    ~~~~~~~~~~~~~~~^
        self._test_event_was_handled(
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            dispvm_name, "domain-feature-set:preload-dispvm-completed"
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        )
        ^
    )
    ^
AssertionError: None is not true
```

This is fixed.

ben-grande:
Latest commits didn't fix the problem, just adding some debugging, but currently fails on more tests. Unfinished.

ben-grande:
The other errors appears to be only because that test failed, although it should not have impacted other tests.

qubes/tests/integ/dispvm.py
```diff
@@ -233,10 +251,70 @@ async def no_preload(self):
...
+            await asyncio.sleep(1)
+        else:
+            if fail_on_timeout:
+                self.fail("didn't preload in time")
+        if not wait_completion:
+            logger.info("end")
+            return
+        preload_dispvm = self.disp_base.get_feat_preload()
+        preload_unfinished = preload_dispvm
+        for _ in range(60):
```
marmarek: This is not very nice place to add handlers... If you are unlucky, some events may get fired before you manage to add handlers (for example `domain-feature-set:preload-dispvm-in-progress` as it's fired quite early). Maybe you could instead add registering those handlers in `domain-add` event? Something like:
```python

def _register_handlers(self, vm):
    # all the vm.add_handler(...) calls

def _on_domain_add(self, app, event, vm):
    if (some condition checking if preloaded? or maybe do unconditionally?):
        self._register_handlers(vm)

async def wait_preload(...):
    self.app.add_handler("domain-add", self._on_domain_add)
    # and then call _register_handlers for already existing dispvms
```

ben-grande: > unconditionally?

Better, domain is added then it is marked as preloaded. For VMs that are irrelevant, the event checker won't run.

ben-grande:
> * TC_20_DispVM_debian-12-xfce: [test_013_dvm_run_preload_gui (assert_type=False, qube='test-inst-dvm', event='domain-preload-dispvm-start', test=True)](https://openqa.qubes-os.org/tests/144550#step/TC_20_DispVM_debian-12-xfce/4) (failure)
>       `~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^... AssertionError: True is not false`

Fixed by:

```python3
self._test_event_handler_remove(appvm, "domain-preload-dispvm-start")
```
before using the preload.

qubes/tests/integ/dispvm.py
```diff
@@ -215,28 +235,99 @@ def tearDown(self):  # pylint: disable=invalid-name
...
+                break
+            await asyncio.sleep(1)
+        else:
+            if fail_on_timeout:
+                self.fail("last preloaded didn't complete in time")
+        logger.info("end")
+
     async def run_preload_proc(self):
+        logger.info("start")
+        # TODO: ben: to marek: DEVNULL may work because differently than None,
```
marmarek: I'm okay with DEVNULL too

ben-grande: Done.

marmarek (approved):

PR #681, opened 2025-05-19
Allow VM boot modes to change a VM's default user

ArrayBolt3:
Pretty straightforward, adds another feature category `boot-mode.default-user.<boot-mode-name>`. Regression tests pass on my end, and the feature itself works in testing. I'm able to remove the qrexec overrides for Whonix-Workstation, add a default-user feature for sysmaint mode, and have sysmaint mode appear to function properly. Software updates also work again as dom0 can now run commands in VMs as root again.

### DIFF
--- a/qubes/ext/core_features.py
+++ b/qubes/ext/core_features.py
@@ -108,9 +108,11 @@ class CoreFeatures(qubes.ext.Extension):
         # handle boot mode advertisement
         old_bootmode_info = {}
         for feature_key, feature_val in vm.features.items():
-            if feature_key.startswith(
-                "boot-mode.kernelopts."
-            ) or feature_key.startswith("boot-mode.name."):
+            if (
+                feature_key.startswith("boot-mode.kernelopts.")
+                or feature_key.startswith("boot-mode.name.")
+                or feature_key.startswith("boot-mode.default-user.")
+            ):
                 old_bootmode_info[feature_key] = feature_val
         new_bootmode_info = {}
         new_bootmode_names = []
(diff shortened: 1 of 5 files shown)


ArrayBolt3:
The one line missing coverage appears to be a change of a `self.default_user` reference to a `self.get_default_user()` call. The original line wasn't covered either.

qubes/ext/core_features.py
```diff
@@ -136,19 +138,22 @@ async def qubes_features_request(self, vm, event, untrusted_features):
...
+            bootmode_name = bootmode_key_parts[2]
+            if bootmode_name == "":
+                continue
+            if (
+                f"boot-mode.kernelopts.{bootmode_name}" not in new_bootmode_info
+            ) and bootmode_name != "default":
+                continue
+            if untrusted_feature_key.startswith(
+                "boot-mode.name."
+            ) or untrusted_feature_key.startswith("boot-mode.default-user."):
```
marmarek, line 156: You allow setting `boot-mode.default-user.default`, but it's later ignored in `get_default_user()`. Please make it consistent (if you have some good argument for it, I can be convinced to allowing default user for the default boot mode too; otherwise don't allow setting it here).

qubes/tests/vm/qubesvm.py
marmarek, line 1: Maybe also worth extending integration test in qubes/tests/integ/basic.py, to check if really new default user got used?

marmarek (changes requested):

marmarek:
Besides the comments above, it looks fine and together with whonix developers repo updates seems to fix the issue.

ArrayBolt3:
This hasn't fallen off my radar, I've just been having to tackle other work responsibilities and have had very limited time (too limited to look at this) for a few days. I'm scheduled to resume work on Whonix and Kicksecure again starting today, and this is my top priority to look at, so expect something later today.

qubes/ext/core_features.py
ArrayBolt3: Fixed, test added for this edge case.

qubes/tests/vm/qubesvm.py
ArrayBolt3: Good idea, implemented.

ArrayBolt3:
@marmarek Should be ready for re-review.

qubes/tests/integ/basic.py
```diff
@@ -513,6 +513,32 @@ async def _test_bootmode(self, tpl, vm):
...
+        )
+        await vm.run_for_stdio(
+            "chmod +x " "/etc/qubes/post-install.d/50-test.sh", user="root"
+        )
+        await vm.run_service_for_stdio("qubes.PostInstall", user="root")
+        await vm.shutdown(wait=True)
+
+        await vm.start()
+        user_id = (await vm.run_for_stdio("id -un"))[0].decode()
+        self.assertEquals(user_id, "altuser")
```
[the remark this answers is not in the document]
ArrayBolt3: Ah, that's because the tests aren't being run under dom0. That's going to be a bit trickier to do, I'll try to run them under dom0 locally though.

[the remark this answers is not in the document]
ArrayBolt3: Kk. I should still have a dom0 testing environment for in the future though, so my time setting it up won't be wasted, but I will go ahead and let openQA handle it for this time around. Thank you!

marmarek (approved):

marmarek:
@ArrayBolt3 do you remember why it isn't plugged into `default_user` property, but a bunch of specific places instead of using that property directly? I think there was some reason for it, just can't find it now...

ArrayBolt3:
Unfortunately I don't remember anymore. I searched Matrix chat history and found some context, but not enough to fully piece together the original rationale.

PR #690, opened 2025-06-25
New option disable snapshots to use a private volume directly

Guiiix:
This PR is related to issue https://github.com/QubesOS/qubes-issues/issues/8767

A new option **disable snapshots** can be set on volumes (lvm, file, reflink and zfs supported). When enabled on a private volume, the volume is used directly instead of doing a snapshot and use it.

The option can be enabled by setting ``revisions_to_keep = -1``

Using the volume directly saves space but requires features like backup, dispvm or cloning to be disabled when the volume is in use.

### DIFF
--- a/qubes/backup.py
+++ b/qubes/backup.py
@@ -517,10 +517,15 @@ class Backup:
             summary_line += fmt.format(size_to_human(vm_size))
 
             if qid != 0 and vm_info.vm.is_running():
-                summary_line += (
-                    " <-- The VM is running, backup will contain "
-                    "its state from before its start!"
-                )
+                if [volume for volume in vm_info.vm.volumes.values()
+                           if volume.snapshots_disabled]:
+                    summary_line += (
+                        " <-- The VM is running and private volume snapshots "
+                        "are disabled. Backup will fail!"
+                    )
+                else:
+                    summary_line += " <-- The VM is running, backup will " \
+                                    "contain its state from before its start!"
 
             summary += summary_line + "\n"
 
@@ -823,6 +828,15 @@ class Backup:
             backup_app.domains[qid].features["backup-content"] = True
             backup_app.domains[qid].features["backup-path"] = vm_info.subdir
             backup_app.domains[qid].features["backup-size"] = vm_info.size
+
+            # VM running private volumes without snapshoting them
+            # (revision_to_keep = -1) must be powered off to be backup
+            if backup_app.domains[qid].is_running:
+                for volume in backup_app.domains[qid].volumes.values():
+                    if volume.snapshots_disabled:
+                        raise qubes.exc.QubesVMNotHaltedError(
+                            backup_app.domains[qid]
+                        )
         backup_app.save()
         del backup_app
 
--- a/qubes/backup.py
+++ b/qubes/backup.py
@@ -517,15 +517,20 @@ class Backup:
             summary_line += fmt.format(size_to_human(vm_size))
 
             if qid != 0 and vm_info.vm.is_running():
-                if [volume for volume in vm_info.vm.volumes.values()
-                           if volume.snapshots_disabled]:
+                if [
+                    volume
+                    for volume in vm_info.vm.volumes.values()
+                    if volume.snapshots_disabled
+                ]:
                     summary_line += (
                         " <-- The VM is running and private volume snapshots "
                         "are disabled. Backup will fail!"
                     )
                 else:
-                    summary_line += " <-- The VM is running, backup will " \
-                                    "contain its state from before its start!"
+                    summary_line += (
+                        " <-- The VM is running, backup will "
+                        "contain its state from before its start!"
+                    )
 
             summary += summary_line + "\n"
 
--- a/qubes/storage/__init__.py
+++ b/qubes/storage/__init__.py
@@ -809,6 +809,12 @@ class Storage:
 
     async def start(self):
         """Execute the start method on each volume"""
+        for vol in self.vm.volumes.values():
+            if (vol.source and vol.source.snapshots_disabled
+                    and vol.source.is_running()):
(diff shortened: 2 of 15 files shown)


rustybird:
Thank you for tackling this! It should make a big difference with database workloads.

For file-reflink IMO it would be neater if the snapshot_disabled volume only had a _path_dirty file without a _path_clean file, instead of the other way around:

That would prevent some combinations which don't make sense, e.g. accidentally exporting this kind of volume or still having an existing _path_precache file or existing revision files on disk. (Setting revisions_to_keep to -1 should delete/prune them *after* renaming _path_clean to _path_dirty, so that it fails early in case of a severely broken volume with a missing _path_clean file.)

It would also keep path = _path_dirty = "the thing that's connected to Xen if the VM is running" constant in all cases. As a user, I always found the indirection a bit confusing for lvm_thin.

rustybird:
> That would prevent some combinations which don't make sense, e.g. accidentally exporting this kind of volume

Sorry, I didn't get that export *should* be possible for a snapshot_disabled volume if it's stopped!

Okay that's a point in favor of the way you did it, because if export instead returned _path_dirty to a caller that doesn't immediately open the file, and then the user re-enables snapshots and starts the volume, later on the caller would use the actual live data. :(

qubes/storage/file.py
```diff
@@ -490,9 +490,14 @@ async def start(self):
...
             _remove_if_exists(self.path_cow)
         if not os.path.exists(self.path_cow):
-            create_sparse_file(self.path_cow, self.size)
+            if not self.snapshots_disabled:
+                create_sparse_file(self.path_cow, self.size)
         if not self.snap_on_start:
             _check_path(self.path)
+
+        if self.snapshots_disabled:
+            return
```
marmarek:
```suggestion
            return self
```

qubes/backup.py
```diff
@@ -520,8 +520,8 @@ def get_backup_summary(self):
                 if [volume for volume in vm_info.vm.volumes.values()
                            if volume.snapshots_disabled]:
                     summary_line += (
-                        " <-- The VM is running, backup will contain "
-                        "its state from before its start!"
+                        " <-- The VM is running and private volume snapshots "
+                        "are disabled. Backup will fail!"
```
marmarek, line 527: Squash this commit into "backup: prevent backup of running VM when private volume snapshots are disabled"?

marmarek (changes requested):
Ok, I've read the whole thing, and it looks pretty solid! I have just one minor request below.

I'll do some testing now...

marmarek:
With LVM pool, I was allowed to clone a running AppVM that has revisions_to_keep=-1 on its private volume...

qubes/storage/__init__.py
```diff
@@ -513,6 +515,27 @@ def _not_implemented(self, method_name):
...
+            and not self.snap_on_start
+            and self.save_on_stop
+        )
+
+    @property
+    def state_file(self) -> str:
+        return os.path.join(
+            VOLUME_STATE_DIR,
+            VOLUME_STATE_PREFIX
+            + f"{self.pool.name}__{self.vid}".replace("-", "--").replace(
```
rustybird: `_` is an [allowed character for pool names](https://github.com/QubesOS/qubes-core-admin/blob/v4.3.29/qubes/api/admin.py#L772) and for vids, so this is not injective.

```suggestion
            + f"{self.pool.name}:{self.vid}".replace("-", "--").replace(
```

qubes/storage/__init__.py
```diff
@@ -813,6 +869,9 @@ async def stop(self):
             )
             for name, vol in self.vm.volumes.items()
         )
+        for vol in self.vm.volumes.values():
+            if os.path.exists(vol.state_file):
+                os.unlink(vol.state_file)
```
rustybird:
```suggestion
            qubes.utils.remove_file(vol.state_file)
```

Guiiix:
With the last commit, the file is renamed ``-dirty`` at startup and ``-clean`` at shutdown. In this way, the dirty volume is the one connected to Xen, but when the VM is not running, we have a clean volume just like when revisions_to_keep is >= 0. Are you okay with this?

rustybird:
Nice, that actually feels like the most natural way to fit in the feature.

Then _update_precache() should not call _copy_file() if snapshot_disabled, so resizing a volume won't create a precache only to remove it again on start. Other than that, the file-reflink changes LGTM (untested, I'm still on R4.2)

qubes/storage/__init__.py
```diff
@@ -863,6 +921,20 @@ async def import_data_end(self, volume, success):
             self.get_volume(volume).import_data_end(success=success)
         )
 
+    async def import_volume(self, dst_volume, src_volume):
+        """Helper function to import data from another volume"""
+        src_volume = self.get_volume(src_volume)
```
marmarek: Source volume for import usually is from another VM, so get_volume() won't work if you give it just a name. You can either require also source vm object in arguments, or simply require Volume object and don't support name in `src_volume` arg - in which case the get_volume call is unnecessary.

Guiiix: Ah I see. That makes sense, thank you!

Guiiix:
> Nice, that actually feels like the most natural way to fit in the feature.
> 
> ~Then _update_precache() should not call _copy_file() if snapshots_disabled, so resizing a volume won't create a precache only to remove it again on start.~ Oh you already did that too. Perfect! The file-reflink changes LGTM (untested, I'm still on R4.2)

Thank you, if you want to try it on R4.2, you can use this branch:
https://github.com/Guiiix/qubes-core-admin/tree/disable_snapshot_4.2

I am on R4.2 too :)

marmarek (approved):

PR #695, opened 2025-07-07
 Add Virtual DNS for IPv6 (#558)

### LINKED ISSUE #10038: IPv6 DNS support

3nprob:
- #558 squashed and rebased on main

No changes beyond conflict resolution and formatting lint

#### Related
- https://github.com/QubesOS/qubes-core-agent-linux/pull/592

Fixes https://github.com/QubesOS/qubes-issues/issues/10038

### DIFF
--- a/qubes/tests/api_admin.py
+++ b/qubes/tests/api_admin.py
@@ -243,27 +243,63 @@ netvm default=True type=vm \n'''
         self.assertEqual(value, expected)
 
     def test_028_vm_property_get_list(self):
+        self.vm.features['ipv6'] = '1'
         self.vm.provides_network = True
         value = self.call_mgmt_func(
             b'admin.vm.property.Get',
             b'test-vm1',
             b'dns')
         self.assertEqual(value, 'default=True type=str 10.139.1.1 10.139.1.2')
+        value = self.call_mgmt_func(
+            b'admin.vm.property.Get',
+            b'test-vm1',
+            b'dns6')
+        self.assertEqual(value, 'default=True type=str ')
+        self.vm.features['supported-feature.ipv6dns'] = '1'
+        value = self.call_mgmt_func(
+            b'admin.vm.property.Get',
+            b'test-vm1',
+            b'dns6')
+        self.assertEqual(value, 'default=True type=str fd09:24ef:4179::a8b:1 fd09:24ef:4179::a8b:2')
 
     def test_029_vm_property_get_list_none(self):
+        self.vm.features['ipv6'] = '1'
         value = self.call_mgmt_func(
             b'admin.vm.property.Get',
             b'test-vm1',
             b'dns')
         self.assertEqual(value, 'default=True type=str ')
+        value = self.call_mgmt_func(
+            b'admin.vm.property.Get',
+            b'test-vm1',
+            b'dns6')
+        self.assertEqual(value, 'default=True type=str ')
+        self.vm.features['supported-feature.ipv6dns'] = '1'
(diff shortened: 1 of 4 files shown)


qubes/tests/api_admin.py
```diff
@@ -341,24 +341,50 @@ def test_027_vm_property_get_all(self):
...
+        self.vm.features["ipv6"] = "1"
         self.vm.provides_network = True
         value = self.call_mgmt_func(
             b"admin.vm.property.GetDefault", b"test-vm1", b"dns"
         )
         self.assertEqual(value, "type=str 10.139.1.1 10.139.1.2")
+        value = self.call_mgmt_func(
+            b"admin.vm.property.Get", b"test-vm1", b"dns6"
+        )
+        self.assertEqual(value, "type=str ")
```
marmarek: This and the one below lacks `default=True ` prefix (see test failure message).

marmarek:
I've added issue reference to PR descriptions, please include it also in a commit message next time you push a change (in both PRs).

qubes/tests/api_admin.py
3nprob: Fixed but there's still this to resolve:

```
  File "/usr/lib64/python3.12/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/gitlab-runner/builds/QubesOS/qubes-core-admin/qubes/tests/firewall.py", line 153, in test_007_ipv4_invalid_network
    with self.assertRaises(ValueError):
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: ValueError not raised
EXPECTED: qubes.tests.firewall/TC_02_DstHost/test_015_ipv6_invalid_network
  File "/usr/lib64/python3.12/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/gitlab-runner/builds/QubesOS/qubes-core-admin/qubes/tests/firewall.py", line 203, in test_015_ipv6_invalid_network
    with self.assertRaises(ValueError):
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: ValueError not raised
```

marmarek: expected failures are... expected, you don't need to do anything about them

qubes/tests/api_admin.py
```diff
@@ -341,24 +341,50 @@ def test_027_vm_property_get_all(self):
...
     def test_029_vm_property_get_list_none(self):
+        self.vm.features["ipv6"] = "1"
         value = self.call_mgmt_func(
-            b"admin.vm.property.Get", b"test-vm1", b"dns"
+            b"admin.vm.property.Get", b"test-vm1", b"dns6"
+        )
+        self.assertEqual(value, "default=True type=str ")
+        self.vm.features["supported-feature.ipv6dns"] = "1"
+        value = self.call_mgmt_func(
+            b"admin.vm.property.Get", b"test-vm1", b"dns6"
```
ArrayBolt3, line 368: This discards the IPv4 DNS test. I'm not sure that's intentional or desirable.

PR #686, opened 2025-06-07
Handle global disposable preload feature

ben-grande:
For: https://github.com/QubesOS/qubes-issues/issues/1512
For: https://github.com/QubesOS/qubes-desktop-linux-manager/pull/262

PR #700, opened 2025-07-11
Autostart global disposable preload

ben-grande:

---

Missed this on https://github.com/QubesOS/qubes-core-admin/pull/686.


marmarek (changes requested):
In some places you check klass to be AppVM, and in others name to be "dom0", please make it consistent, one way or the other.

PR #694, opened 2025-07-04
Cleanup qmemman

ben-grande:
Plan:

- pylint
- remove code comments according to [guidelines](https://www.qubes-os.org/doc/coding-style/#general-programming-style-guidelines)
- renaming variables to have an explanatory name according to [guidelines](https://www.qubes-os.org/doc/coding-style/#general-typographic-conventions)
- document the functions

PR #702, opened 2025-07-16
Set less memory to domain before pausing for preload

ben-grande:
For: https://github.com/QubesOS/qubes-issues/issues/9917

Requires (based on top of): https://github.com/QubesOS/qubes-core-admin/pull/694

---

Not working as intended, there still is work/debugging to be done.


qubes/tools/qmemmand.py
```diff
@@ -224,9 +224,10 @@ def handle(self):
                         resp = "OK"
                 elif len(data_args) == 2:
                     resp = "FAIL"
-                    domid, memory = str(data_args[0]), int(data_args[1])
-                    # TODO: ben: create the function to set memory to a domain.
-                    if system_state.mem_set(domid, memory):
```
marmarek: what was the issue with using `system_state.mem_set`? Using under `if` looks weird, as it always returns None, but otherwise looks correct?
What you might need to add is waiting for the domain to actually balloon down to the requested size - possibly with a simple loop that calls `refresh_mem_actual` and checks mem_actual, with some timeout.

ben-grande: I didn't test with `mem_set()` because I saw that `do_balloon` was calling it with a loop, so `mem_set()` was there to give me hint of what to look next, I didn't test with only it.

I reused `do_balloon()` in order to avoid code duplication, but a lot of functions there use the class object `self.dom_dict` and that might cause some interference. I will try a new function with a simple loop, `do_balloon()` might indeed be overkill for a single domain changing values. I thought the xenfree value was necessary to be sure the memory was returned to Xen as well as reusing the waiting time and transfer speed to check again.

marmarek:
This needs a rebase now

marmarek (approved):

PR #705, opened 2025-07-21
Update Whonix-Workstation file open and dispvm open tests

ArrayBolt3:
Whonix-Workstation now presents a qrexec confirmation prompt to the user before opening files in a DispVM or existing Whonix-Workstation VM. Explicitly confirm these actions in the appropriate tests.

This is my initial attempt at implementing option 2 from https://github.com/QubesOS/qubes-core-admin-addon-whonix/pull/22#issuecomment-3094587696, to get OpenQA's tests to pass. I don't have a system set up to allow me to run rtests on dom0, so I have no clue if this works, we'll need to wait for another OpenQA run to tell us that (or I'll need to overhaul my testing system).

### DIFF
--- a/qubes/tests/integ/dispvm.py
+++ b/qubes/tests/integ/dispvm.py
@@ -879,6 +879,41 @@ class TC_20_DispVMMixin(object):
         else:
             raise KeyError(window_title)
 
+    def _whonix_ws_dispvm_confirm(self, action_str):
+        try:
+            winid = self.wait_for_window(
+                "qrexec-policy-agent",
+                search_class=True,
+                include_tray=False,
+                timeout=5,
+            )
+        except Exception:
+            return (
+                False,
+                "Failed to find qrexec confirmation window for "
+                f"{action_str} action",
+            )
+        try:
+            subprocess.run(
+                [
+                    "bash",
+                    "-c",
+                    "--",
+                    f"xdotool windowfocus {winid}; sleep 1.1; "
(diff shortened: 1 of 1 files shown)


qubes/tests/integ/dispvm.py
```diff
@@ -879,6 +879,41 @@ def _handle_editor(self, winid, copy=False):
         else:
             raise KeyError(window_title)
 
+    def _whonix_ws_dispvm_confirm(self, action_str):
+        try:
+            winid = self.wait_for_window(
+                "Operation execution",
+                search_class=True,
```
marmarek, line 886: search_class=True means to look based on WM_CLASS property instead of the title. You can do either here, but with search_class=True, you need to actually use window class (xprop says it's `qrexec-policy-agent`)

ArrayBolt3: Ah, that makes sense. I'll switch to the class name then.

marmarek:
The exception handling is a bit unusual (with returning tuple - looks like golang?), but it works, so lets keep it. For future, you can simply call `self.fail()` in the helper function directly.

marmarek (approved):

ArrayBolt3:
Ah, I thought it might be safe to use `self.fail()` in the helper function, but wasn't entirely sure and so rewrote it to use the more Go-like error handling in the hopes of getting it to work the first time. I'll try to remember that for future reference, thank you!

PR #706, opened 2025-07-22
Fix preload autostart integration test

ben-grande:
For: https://github.com/QubesOS/qubes-core-admin/pull/700


qubes/tests/integ/dispvm.py
```diff
@@ -612,13 +612,24 @@ def test_017_preload_autostart(self):
...
         proc = self.loop.run_until_complete(
             asyncio.create_subprocess_exec("/usr/lib/qubes/preload-dispvm")
         )
         self.loop.run_until_complete(asyncio.wait_for(proc.wait(), timeout=30))
-        self.assertEqual(self.disp_base.get_feat_preload(), [])
+        preload_dispvm = self.disp_base.get_feat_preload()
+        self.assertEqual(len(old_preload), preload_max)
+        self.assertEqual(len(preload_dispvm), preload_max)
+        self.assertTrue(
+            set(old_preload).isdisjoint(preload_dispvm),
```
marmarek, line 632: Why is it supposed to be disjoint? I don't see anything using or killing preloaded between collecging old_preload and preload_dispvm? You just unset the feature on the disposable template, but it's still set via global property at that time, so it shouldn't matter, no?

ben-grande: This unsets the feature:

```
        proc = self.loop.run_until_complete(
            asyncio.create_subprocess_exec("/usr/lib/qubes/preload-dispvm")
        )
```

In other words, event `domain-preload-dispvm-autostart` does it.

marmarek: Ok, this could use a comment then, next to calling that script here.

ben-grande: I added as a doctstring because there is another disjoint on the block before and I prefer to consider both blocks with a single comment.

PR #707, opened 2025-07-22
Refresh preloaded disposables on outdated volumes

ben-grande:
Fixes: https://github.com/QubesOS/qubes-issues/issues/10026

---

Missing unit tests.

### DIFF
--- a/qubes/vm/dispvm.py
+++ b/qubes/vm/dispvm.py
@@ -471,46 +471,80 @@ class DispVM(qubes.vm.qubesvm.QubesVM):
 
         if not preload and appvm.can_preload():
             # Not necessary to await for this event as its intent is to fill
-            # gaps and not relevant for this run.
+            # gaps and not relevant for this run. Delay to not affect this run.
             asyncio.ensure_future(
                 appvm.fire_event_async(
-                    "domain-preload-dispvm-start", reason="there is a gap"
+                    "domain-preload-dispvm-start",
+                    reason="there is a gap",
+                    delay=5,
                 )
             )
 
         if not preload and (preload_dispvm := appvm.get_feat_preload()):
-            dispvm = app.domains[preload_dispvm[0]]
-            dispvm.log.info("Requesting preloaded qube")
-            # The property "preload_requested" offloads "preload-dispvm" and
-            # thus avoids various race condition:
-            # - Decreasing maximum feature will not remove the qube;
-            # - Another request to this function will not return the same qube.
-            dispvm.features["preload-dispvm-in-progress"] = True
-            appvm.remove_preload_from_list([dispvm.name])
-            dispvm.preload_requested = True
-            app.save()
-            timeout = int(dispvm.qrexec_timeout * 1.2)
-            try:
-                if not dispvm.features.get("preload-dispvm-completed", False):
-                    dispvm.log.info(
-                        "Waiting preload completion with '%s' seconds timeout",
-                        timeout,
+            dispvm = None
+            for item in preload_dispvm:
+                qube = app.domains[item]
+                if any(vol.is_outdated() for vol in qube.volumes.values()):
+                    qube.log.warning(
+                        "Requested preloaded qube but it is outdated, trying "
+                        "another one if available"
                     )
-                    async with asyncio.timeout(timeout):
-                        await dispvm.preload_complete.wait()
-                if dispvm.is_paused():
-                    await dispvm.unpause()
-                else:
-                    dispvm.use_preload()
+                    # The gap is filled after the delay set by the
+                    # 'domain-shutdown' of its ancestors. Not refilling now to
+                    # deliver a disposable faster.
+                    appvm.remove_preload_from_list([qube.name])
+                    # Delay to not  affect this run.
+                    asyncio.ensure_future(
+                        qube.delay(delay=2, coros=[qube.cleanup()])
+                    )
+                    continue
+                dispvm = qube
+                break
+            if dispvm:
+                dispvm.log.info("Requesting preloaded qube")
+                # The property "preload_requested" offloads "preload-dispvm"
+                # and thus avoids various race condition:
+                # - Decreasing maximum feature will not remove the qube;
+                # - Another request to this function will not return the same
+                #   qube.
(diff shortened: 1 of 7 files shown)


qubes/vm/dispvm.py
```diff
@@ -472,45 +472,73 @@ async def from_appvm(cls, appvm, preload=False, **kwargs):
...
-                else:
-                    dispvm.use_preload()
+                    # The gap is filled after the delay set by the
+                    # 'domain-shutdown' of its ancestors. Not refilling now to
+                    # deliver a disposable faster.
+                    appvm.remove_preload_from_list([qube.name])
+                    # TODO: ben: cleanup with delay? Delay can help because the
+                    #   cleanup is not relevant to this run.
+                    asyncio.ensure_future(qube.cleanup())
+                dispvm = qube
```
marmarek, line 501: This loop doesn't stop checking. There should be some break/continue to stop at first non-outdated.

ben-grande: Fixed.

qubes/vm/dispvm.py
```diff
@@ -472,45 +472,74 @@ async def from_appvm(cls, appvm, preload=False, **kwargs):
...
+                break
```
marmarek, line 502: Should it be under `else:`? otherwise you will try to use a dispvm that you just scheduled to be cleaned up...
Or maybe add continue after scheduling cleanup?

ben-grande: I didn't test the latest version since the TODOs yet, focusing on the benchmark PR to test delays on different stages (somehow put a delay in those places).

ben-grande: Anyway. fixed.

qubes/vm/dispvm.py
```diff
@@ -471,46 +471,80 @@ async def from_appvm(cls, appvm, preload=False, **kwargs):
...
         if not preload and appvm.can_preload():
             # Not necessary to await for this event as its intent is to fill
-            # gaps and not relevant for this run.
+            # gaps and not relevant for this run. Delay to not affect this run.
             asyncio.ensure_future(
                 appvm.fire_event_async(
-                    "domain-preload-dispvm-start", reason="there is a gap"
+                    "domain-preload-dispvm-start",
+                    reason="there is a gap",
+                    delay=5,
```
ben-grande, line 479: This is also relevant on refresh because the old preloads might be removed already and there is a gap, but because of the delay in `refresh_preload`, there is a higher chance of this event being called. Therefore, add a delay to not attempt preload.

qubes/vm/dispvm.py
```diff
@@ -471,46 +471,80 @@ async def from_appvm(cls, appvm, preload=False, **kwargs):
...
-                    await dispvm.unpause()
-                else:
-                    dispvm.use_preload()
+                    # The gap is filled after the delay set by the
+                    # 'domain-shutdown' of its ancestors. Not refilling now to
+                    # deliver a disposable faster.
+                    appvm.remove_preload_from_list([qube.name])
+                    # Delay to not  affect this run.
+                    asyncio.ensure_future(
+                        qube.delay(delay=2, coros=[qube.cleanup()])
```
ben-grande, line 498: 2 seconds seems to be enough according to qubesd logs.

qubes/vm/dispvm.py
```diff
@@ -471,46 +471,80 @@ async def from_appvm(cls, appvm, preload=False, **kwargs):
...
+                except asyncio.TimeoutError:
+                    dispvm.log.warning(
+                        "Requested preloaded qube but failed to finish "
+                        "preloading after '%d' seconds, falling back to normal "
+                        "disposable",
+                        int(timeout),
+                    )
+                    # Delay to not affect this run.
+                    asyncio.ensure_future(
+                        dispvm.delay(delay=2, coros=[dispvm.cleanup()])
```
ben-grande, line 541: 2 seconds seems to be enough according to qubesd logs.

qubes/vm/mix/dvmtemplate.py
```diff
@@ -370,10 +392,32 @@ def can_preload(self) -> bool:
...
+            self.remove_preload_from_list([qube.name])
+        if outdated:
+            tasks = [self.app.domains[qube].cleanup() for qube in outdated]
+            asyncio.ensure_future(asyncio.gather(*tasks))
+            # Delay to not overload the system with cleanup+preload.
+            asyncio.ensure_future(
+                self.fire_event_async(
+                    "domain-preload-dispvm-start",
+                    reason="of outdated volume(s)",
+                    delay=4,
```
ben-grande, line 413: This delay gives enough time for the old preloads to be cleaned up before trying to preload. It is intended to avoid overload of the system because cleanup and preload can be a bit intensive when combined with many preloaded disposables. There is no perfect delay here, basing of what my tests told were enough, but my hardware is not the same as any other user's hardware.

qubes/vm/qubesvm.py
```diff
@@ -2900,6 +2902,15 @@ def get_pref_mem(self):
 
         return qubes.qmemman.algo.pref_mem(domain) / 1024
 
+    async def delay(self, delay: float | int, coros: list[Awaitable]) -> None:
+        self.log.info(
```
marmarek: log.debug?

ben-grande: Fixed.

qubes/vm/dispvm.py
```diff
@@ -471,46 +471,80 @@ async def from_appvm(cls, appvm, preload=False, **kwargs):
...
-            timeout = int(dispvm.qrexec_timeout * 1.2)
-            try:
-                if not dispvm.features.get("preload-dispvm-completed", False):
-                    dispvm.log.info(
-                        "Waiting preload completion with '%s' seconds timeout",
-                        timeout,
+            dispvm = None
+            for item in preload_dispvm:
+                qube = app.domains[item]
+                if any(vol.is_outdated() for vol in qube.volumes.values()):
```
ben-grande, line 487: Just to be clear of why this is needed, receiving the event `domain-shutdown`, as far as I know, does not enforce that it runs before anything else, including running before this section.

PR #710, opened 2025-07-30
Add available memory threshold to preload event

ben-grande:
For: https://github.com/QubesOS/qubes-issues/issues/10027


ben-grande:
I did consider a dynamic management depending on the threshold but I thought that would be too much. The management would be done on `domain-feature-delete` and `domain-feature-set` to dynamically add or reduce depending on the threshold. It will not be easy because it involves checking for every dvm template that has the preload feature and set some heuristics of what to do in each case... therefore, I went with the simpler option.

The openqa test low_mem is working.

qubes/tests/integ/dispvm.py
```diff
@@ -498,20 +498,38 @@ async def _test_012_preload_low_mem(self):
...
         with patch("builtins.open", side_effect=mock_open_mem):
-            preload_max = 2
+            logger.info("low mem standard")
+            self.disp_base.features["preload-dispvm-max"] = str(preload_max)
+            await self.wait_preload(
+                preload_max, fail_on_timeout=False, timeout=15
+            )
+            self.assertEqual(1, len(self.disp_base.get_feat_preload()))
+            # Nothing will be done here, just to prepare to the next test.
+            self.disp_base.features["preload-dispvm-max"] = str(preload_max - 1)
```
marmarek, line 522: Shouldn't it wait for cleanup to complete before going further with the test?

ben-grande: Not necessary as no cleanup will happen. It didn't preload 2, it preloaded only 1.

```python3
            self.assertEqual(1, len(self.disp_base.get_feat_preload()))
            # Nothing will be done here, just to prepare to the next test.
```

And if it preloaded anything other than 1, it would fail the test.

marmarek (approved):

PR #711, opened 2025-08-01
Fail early on missing service for preloaded disposables

ben-grande:
Occurs when the global max preload feature, set on dom0, with the default-dispvm being an unsupported template, which can happen on old ISOs, old templates, downgraded templates or packages. When setting the dom0 feature or when switching the default-dispvm, there is no supported-rpc validation. Handle gracefully in case the qube is the default-dispvm and the global max feature is set.

Without this validation, it tries to preload and fails on startup check of the same supported-rpc, but this time, running it as a service.

---

Untested.

Without it, doesn't break anything, but a minor annoyance of failing the `qubes-preload-dispvm.service` and `qubesd` logs and the preloads trying to start to just fail. This has more chance of happening with the global preload feature as it doesn't check if the default-dispvm supports the necessary RPC service.

### DIFF
--- a/qubes/vm/mix/dvmtemplate.py
+++ b/qubes/vm/mix/dvmtemplate.py
@@ -124,8 +124,7 @@ class DVMTemplateMixin(qubes.events.Emitter):
             raise qubes.exc.QubesValueError("Qube does not support qrexec")
 
         service = "qubes.WaitForRunningSystem"
-        supported_service = "supported-rpc." + service
-        if not self.features.check_with_template(supported_service, False):
+        if not self.supports_preload():
             raise qubes.exc.QubesValueError(
                 "Qube does not support the RPC '%s'" % service
             )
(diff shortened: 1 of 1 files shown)


qubes/vm/mix/dvmtemplate.py
```diff
@@ -265,6 +264,13 @@ async def on_domain_preload_dispvm_used(
             event_log += " because %s" % str(kwargs.get("reason"))
         self.log.info(event_log)
 
+        service = "qubes.WaitForRunningSystem"
+        if not self.supports_preload():
+            raise qubes.exc.QubesValueError(
+                "Qube does not support the RPC '%s' but tried to preload, "
+                "check if template is outdated" % service
+            )
```
marmarek, line 290: This should be after possibly removing old preloaded below, no? Otherwise, you may end up with outdated preloaded disposables, especially when connected with https://github.com/QubesOS/qubes-core-admin/pull/707 (if it just lost that service support for some reason).

marmarek (changes requested):

qubes/vm/mix/dvmtemplate.py
ben-grande: If you meant `remove_preload_excess`, it doesn't remove preloads with outdated volumes. If it lost that service support for some reason, it would still be handled by `refresh_preload()` of #707 

1. outdated removed from list
2. outdated cleanup scheduled
3. fresh ones scheduled with a delay

When the scheduled event to `domain-preload-dispvm-start` runs, which tries to create new preloads, it raises the exception above that informing why it can't preload anymore.

ben-grande:
Resolved merge conflict.

PR #718, opened 2025-08-13
Add global preload to template gathering

ben-grande:
Affects suspend, which uses this function.

### DIFF
--- a/linux/aux-tools/preload-dispvm
+++ b/linux/aux-tools/preload-dispvm
@@ -12,39 +12,43 @@ import concurrent.futures
 import qubesadmin
 
 
-def get_max(qube):
-    return int(qube.features.get("preload-dispvm-max", 0) or 0)
+def get_preload_max(qube) -> int | None:
+    value = qube.features.get("preload-dispvm-max", None)
+    return int(value) if value else value
 
 
 async def main():
     app = qubesadmin.Qubes()
     domains = app.domains
     default_dispvm = getattr(app, "default_dispvm", None)
+    global_max = get_preload_max(domains["dom0"])
     appvms = [
         qube
         for qube in domains
-        if get_max(qube) > 0
-        and (
-            (
-                qube.klass == "AppVM"
-                and getattr(qube, "template_for_dispvms", False)
+        if (
+            qube.klass == "AppVM"
+            and getattr(qube, "template_for_dispvms", False)
+            and (
+                (qube != default_dispvm and get_preload_max(qube))
+                or (
+                    (qube == default_dispvm and global_max)
+                    or (global_max is None and get_preload_max(qube))
+                )
             )
-            or (qube.name == "dom0" and default_dispvm)
         )
     ]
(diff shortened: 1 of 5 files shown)


qubes/api/internal.py
```diff
@@ -65,47 +65,95 @@ class SystemInfoCache:
...
         cls.register_events_vm(vm)
 
     @classmethod
-    def on_domain_delete(cls, subject, event, vm):
+    def on_domain_delete(
+        cls,
+        subject,
+        event,
+        vm,
+    ):
```
marmarek: Most (all?) of those reformatting looks bogus (black bug? wrong arguments?) - the line was way under 80 chars, so there is no need for splitting. And black doesn't attempt to do it for me...

ben-grande: Strange indeed. Which black version do you have?

Fedora 42 black from repos:
```
% git rev-parse HEAD
2249d3871bef5494969c192f44afd7217be4a976

% black --version
black, 24.4.2 (compiled: no)
Python (CPython) 3.13.7

% black -l80 --check qubes/api/internal.py 
All done! ✨ 🍰 ✨
1 file would be left unchanged.
```

Debian with black installed via pip (latest)
```
% git rev-parse HEAD
2249d3871bef5494969c192f44afd7217be4a976

% black --version
black, 25.1.0 (compiled: yes)
Python (CPython) 3.11.2

% black -l80 qubes/api/internal.py
All done! ✨ 🍰 ✨
1 file left unchanged.
```

ben-grande: It passed CI with black `24.4.2` on https://gitlab.com/QubesOS/qubes-core-admin/-/jobs/11051605265.

It is a bogus change indeed..., I will amend the commit with only the relevant chunks of code from `qubes/api/internal.py` and see what the black from CI throws.

ben-grande: New test didn't fail: https://gitlab.com/QubesOS/qubes-core-admin/-/jobs/11136244926

ben-grande: And last test: https://gitlab.com/QubesOS/qubes-core-admin/-/jobs/11136313453

Why black wanted a reformatted version before I still don't know.

marmarek:
> system_tests_backup
> 
>     * TC_00_Backup: [test_010_selective_restore](https://openqa.qubes-os.org/tests/150994#step/TC_00_Backup/5) (failure)
>       `: VM test-inst-dvm - property template not properly restored... `

This recently started failing. Last test run on current-testing (no PRs) is green: https://openqa.qubes-os.org/tests/150611, so it's most likely some of the included PRs to blame (not sure which one).

marmarek:
Technically it could be also due to some commits merged into main, but not released yet. Given some are about backups, it's likely./

ben-grande:
Yes, strange, I don't know how that happened. I remember seeing the results all green once, but the comment history does not show that...

Anyway, https://github.com/QubesOS/qubes-core-admin/pull/726

PR #727, opened 2025-08-30
Don't apply PM overrides to thunderbolt when PCI hotplug enabled

dsg22:
These are only needed because Qubes disabled PCI hotplugging in kernel config. If the user compiled a kernel with it enabled, the kernel does PM better.


DemiMarie:
Would it be possible for the Linux kernel to handle power management even with PCI hotplug disabled?

dsg22:
> See description of the patch adding this option: https://github.com/QubesOS/qubes-linux-kernel/blob/main/xen-pciback-pm-suspend.patch

It doesn't really explain the problem that motivated the patch though, beyond that it doesn't work "for some, not yet fully clear, reasons". But I guess maybe we still don't know what's causing it. If so, it's not possible to answer @DemiMarie's question.

PR #728, opened 2025-09-01
Do not register 'testclass' device class outside of tests

marmarek:
Now that more places (especially qvm-clone and ansible) properly lists
available device classes, it's no longer harmless to include 'testclass'
on production system. There is a mechanism to replace some entry point
with another for tests (`substitute_entry_points` context manager), but
unfortunately it can't be used here, as device classes are enumerated at
the qubes.api.admin import time, not test run time (as it's the case for
storage classes). Exclude it normally, and include only if
QUBES_TEST=1 variable is set - which is done in run-tests script.

The 'testclass' devices are used only in unit tests, so it's okay
to not include them in integration tests.

Reported at https://github.com/QubesOS/qubes-ansible/pull/6#issuecomment-3229326890

### DIFF
--- a/run-tests
+++ b/run-tests
@@ -45,6 +45,7 @@ unset dir
 
 PYTHONPATH=${TESTPYTHONPATH}${PYTHONPATH+":${PYTHONPATH}"}
 export PYTHONPATH
+export QUBES_TEST=1
 
 "${PYTHON}" setup.py egg_info --egg-base "${TESTPYTHONPATH}"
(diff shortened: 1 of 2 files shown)


DemiMarie:
Would it be possible to unset this environment variable in the `qubesd` entry point?

marmarek:
No, if you set it, you probably wanted it set. systemd service doesn't magically gets random env variables.

marmarek:
Oh, and this doesn't even matter for qubesd - it's about `setup.py` only - so during package build.

PR #729, opened 2025-09-03
Inform how to debug failed preload startup

ben-grande:
The "domain-start-failed" has to be fired before the domain is killed, as the qube is unregistered from clients when it is removed/powered off.

Fixes: https://github.com/QubesOS/qubes-issues/issues/10200

### DIFF
--- a/qubes/vm/qubesvm.py
+++ b/qubes/vm/qubesvm.py
@@ -1548,6 +1548,10 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.LocalVM):
 
             except Exception as exc:  # pylint: disable=bare-except
                 self.log.error("Start failed: %s", str(exc))
+                # let anyone receiving domain-pre-start know that startup failed
+                await self.fire_event_async(
+                    "domain-start-failed", reason=str(exc)
+                )
                 # This avoids losing the exception if an exception is
                 # raised in self.kill(), because the vm is not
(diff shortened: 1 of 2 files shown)


marmarek:
> But if I used only the service name `qubes.WaitForRunningSystem`

I would say it's still okay to use shorter name, it's an implementation detail how it's called.
Hopefully #9964 will be fixed one day...

qubes/vm/qubesvm.py
```diff
@@ -1548,18 +1548,17 @@ async def start(
 
             except Exception as exc:  # pylint: disable=bare-except
                 self.log.error("Start failed: %s", str(exc))
+                # let anyone receiving domain-pre-start know that startup failed
+                await self.fire_event_async(
+                    "domain-start-failed", reason=str(exc)
+                )
```
marmarek, line 1554: Does moving it actually helps anything? This code (intentionally) does not wait until all clients process an event, so there is no guarantee that qui-domains will still see the VM. Or is it about order of events if kill triggers `domain-delete` event?

ben-grande: It helps as without that, the event will never reach the clients as `domain-shutdown` and `domain-delete` from the `self.kill()` will reach first, thus it will stop tracking the qube. Now that the event is fired before the kill(), it can reach the qui-domains to emit the `domain-start-failed` notification.

PR #731, opened 2025-09-11
Prevent backing up dom0 it itself (in home dir)

alimirjamali:
resolves: https://github.com/QubesOS/qubes-issues/issues/10127

### DIFF
--- a/qubes/backup.py
+++ b/qubes/backup.py
@@ -28,6 +28,7 @@ import grp
 import itertools
 import logging
 import os
+from pathlib import Path
 import pwd
 import shutil
 import stat
@@ -452,6 +453,15 @@ class Backup:
         if 0 in [vm.qid for vm in self.vms_for_backup]:
             local_user = grp.getgrnam("qubes").gr_mem[0]
             home_dir = pwd.getpwnam(local_user).pw_dir
+
+            # Checking if target is not user home directory in dom0
+            if self.target_dir in ["", "~"] or Path(
+                self.target_dir
+            ).is_relative_to(home_dir):
+                raise qubes.exc.QubesException(
+                    "Can not backup dom0 home directory to itself!"
+                )
+
             # Home dir should have only user-owned files, so fix it now
             # to prevent permissions problems - some root-owned files can
             # left after 'sudo bash' and similar commands
--- a/qubes/tests/integ/backup.py
+++ b/qubes/tests/integ/backup.py
@@ -29,6 +29,8 @@ import shutil
 import sys
 
 import asyncio
+import grp
+import pwd
 
 import qubes
 import qubes.backup
@@ -198,7 +200,9 @@ class BackupTestsMixin(object):
         if target is None:
             target = self.backupdir
         try:
-            backup = qubes.backup.Backup(self.app, vms, **kwargs)
+            backup = qubes.backup.Backup(
+                self.app, vms, target_dir=target, **kwargs
+            )
         except qubes.exc.QubesException as e:
             if not expect_failure:
                 self.fail("QubesException during backup_prepare: %s" % str(e))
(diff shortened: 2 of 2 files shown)


alimirjamali:
> both test failed, with different message this time:

Applied the fix.

qubes/backup.py
```diff
@@ -452,6 +453,15 @@ def get_files_to_backup(self):
         if 0 in [vm.qid for vm in self.vms_for_backup]:
             local_user = grp.getgrnam("qubes").gr_mem[0]
             home_dir = pwd.getpwnam(local_user).pw_dir
+
+            # Checking if target is not user home directory in dom0
+            if self.target_dir in ["", "~", None] \
+                    or Path(self.target_dir) == Path(home_dir) \
+                    or Path(home_dir) in Path(self.target_dir).parents:
```
marmarek: `Path(self.target_dir).is_relative_to(home_dir)`?

alimirjamali: It appears that it can replace both of conditions. I will update the code (and unittests) in few seconds.

alimirjamali: This is done. I modified the unittest to check both the `$HOMEDIR` and `$HOMEDIR/somedir`

qubes/tests/integ/backup.py
```diff
@@ -575,6 +575,14 @@ def test_100_backup_dom0_no_restore(self):
         self.make_backup([self.app.domains[0]])
         # TODO: think of some safe way to test restore...
 
+    def test_101_backup_dom0_to_homedir(self):
+        # Test of backup into dom0 home itself results in error...
+        with self.assertRaises(qubes.exc.QubesException):
+            self.make_backup([self.app.domains[0]])
+        with self.assertRaises(qubes.exc.QubesException):
+            self.backupdir = os.environ["HOME"]
```
marmarek: This may not work, as tests are running as root...

alimirjamali: > This may not work, as tests are running as root...

I am considering revisiting the idea of having this test as a part of integration tests. It is a small test which might be better be performed as ordinary tests.

alimirjamali:
> Both tests still fail the same way. It looks like `self.target_dir` is not set yet at this point (it's `None`).

Actually it was not `None`. It was undefined. It should be fixed now.

qubes/tests/integ/backup.py
```diff
@@ -198,7 +198,12 @@ def make_backup(self, vms, target=None, expect_failure=False, **kwargs):
         if target is None:
             target = self.backupdir
         try:
-            backup = qubes.backup.Backup(self.app, vms, **kwargs)
+            backup = qubes.backup.Backup(
+                self.app,
+                vms,
+                target_dir=target,
```
marmarek: In this case, you don't need to set it below, right?

qubes/backup.py
```diff
@@ -452,6 +453,14 @@ def get_files_to_backup(self):
         if 0 in [vm.qid for vm in self.vms_for_backup]:
             local_user = grp.getgrnam("qubes").gr_mem[0]
             home_dir = pwd.getpwnam(local_user).pw_dir
+
+            # Checking if target is not user home directory in dom0
+            if self.target_dir in ["", "~", None] \
```
marmarek: Based on earlier error, `None` shouldn't actually be refused, as it doesn't really mean to backup into user home - it means you don't know where is the target.

qubes/tests/integ/backup.py
alimirjamali: yes

qubes/backup.py
alimirjamali: That is also done. But I believe in practice, `None` is never passed to `qubes.backup.Backup` constructor as the value of target directory. Since the Admin API always expects presence of target directory [here](https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/api/admin.py#L1821).

qubes/tests/integ/backup.py
alimirjamali: I moved the main tests to `api_admin`. This should be performed at usual CI/CD tests. No need to wait for openQA anymore.

alimirjamali:
openQA failed again. I had to pass except_failure=True to make_backup

PR #751, opened 2025-11-11
Use ProtocolError on invalid requests

### LINKED ISSUE #10345: Specific exception against PermissionDenied when API argument is invalid

A broader view of:

- https://github.com/qubesos/qubes-issues/issues/5316

Considering:

- https://github.com/QubesOS/qubes-core-admin/pull/644

Just making some comments to make sure we are aligned.

> I think so, yes, but some of the uses of `enforce()` are for problems that a client cannot realistically know of before making the call. These should raise a better exception that explains exactly what went wrong.

> * Things like “no such snapshot” should use neither `PermissionDenied` nor `ProtocolError`.  Instead, they should use a custom exception type.

```python3
@qubes.api.method("admin.vm.volume.Revert", scope="local", write=True)
async def vm_volume_revert(self, untrusted_payload):
    self.enforce(self.arg in self.dest.volumes.keys())
    untrusted_revision = untrusted_payload.decode("ascii").strip()
    del untrusted_payload
                                                                      
    snapshots = volume.revisions
    self.enforce(untrusted_revision in snapshots)
    revision = untrusted_revision

    self.fire_event_for_permission(volume=volume, revision=revision)
```
Are both of these `enforce()` `ProtocolError` if it fails? If the client consulted `.List` before, it should know the volume names. If the volume or revision don't exist anymore, is still correct to call this `ProtocolError`? This validation seems more close to an HTTP `404`. There are some qubes exceptions for `*NotFoundError`. So in these cases I will use those. Maybe it can be derived from `ProtocolError`.

I see that using `*NotFoundError` [may be leaky](https://github.com/QubesOS/qubes-issues/issues/5316#issuecomment-574217836) if fired after the validation that could possibly raise the custom exception. I don't see `untrusted_*` arguments being passed to `fire_event_for_permission`. Should it accept untrusted arguments that will be sent to extension? It is a breaking change as extension should be adapted if they consider the arguments trusted.

---

```python3
@qubes.api.method(
    "admin.vm.tag.Remove", no_payload=True, scope="local", write=True
)
async def vm_tag_remove(self):
    qubes.vm.Tags.validate_tag(self.arg)

    try:
        self.dest.tags.remove(self.arg)
    except KeyError:
        raise qubes.exc.QubesTagNotFoundError(self.dest, self.arg)
    self.app.save()
```
The `validate_tag` raises `ValueError`, but it is a `ProtocolError`. Maybe it should be `QubesValueError` and that exception should derive from `ProtocolError`?

- It validates the tag by checking the characters.
- Fires event for permission
- Attempts to remove the tag and throw custom exception if the volume doesn't exist

From the above example:

- The example of reverting volumes above validates things before asking for permission, I presume this is because `QubesTagNotFoundError` prints the tag in the exception. If the intention is to [prevent data leak](https://github.com/QubesOS/qubes-issues/issues/5316#issuecomment-574217836), then there is a lot of places that validate arguments against what is present on the system and throws some exception that may or may not be `PermissionDenied`. Therefore, some guidelines must be put in place to prevent this.
- Why was all the tag validation not done before checking for permission? Other cases do it like that, is it unintended to do validation after firing event for permission when it can be done before?
    ```python3
    @qubes.api.method(
        "admin.pool.Remove", no_payload=True, scope="global", write=True
    )
    async def pool_remove(self):
        self.enforce(self.dest.name == "dom0")
        self.enforce(self.arg in self.app.pools.keys())

    ```

- Last code block, I would modify `enforce` to accept passing the exception as an argument:

```python3
def enforce(predicate: bool, exception: Exception = ProtocolError) -> None
    if not predicate:
        raise exception

...

self.enforce(
    self.arg in self.dest.tags,
    qubes.exc.QubesTagNotFoundError(self.dest, self.arg)
)
```

But then what is the value of the `enforce()` method?
```python3
if self.arg not in self.dest.tags:
    qubes.exc.QubesTagNotFoundError(self.dest, self.arg)
```

---

```python3
@qubes.api.method(
    "admin.backup.Execute",
    no_payload=True,
    scope="global",
    read=True,
    execute=True,
)
async def backup_execute(self):
    self.enforce(self.dest.name == "dom0")
    self.enforce(self.arg)
    self.enforce("/" not in self.arg)

    profile_path = os.path.join(
        qubes.config.backup_profile_dir, self.arg + ".conf"
    )
    if not os.path.exists(profile_path):
        raise qubes.exc.PermissionDenied(
            "Backup profile {} does not exist".format(self.arg)
        )
```

> * Dropping the connection without sending a response indicates an unexpected internal error in dom0.  If any details are available, they will be in dom0 logs.  In HTTP, this would be 500 Internal Server Error.

Should I add explanation to every exception (include enforce) I touch, even if minimal to not leak information? If I don't add an explanation, it seems like a server error.

### The solution you'd like

Do not leak system data but still return useful information if client has permissions.

Anyone that may type the name of the template wrong when attempting to create a qube, that raises `PermissionDenied` on invalid template without saying why.

- https://github.com/qubesos/qubes-issues/issues/5316

This case applies to a great range of the API.

_No response_

### LINKED ISSUE #10689: Log dom0 qrexec calls

Sometimes it is useful to know which Qrexec calls are being made by dom0. Because dom0 holds the qrexec daemon, it skips some steps on policy evaluation, skipping log specifically. This makes it difficult to know which Qrexec calls are being made by dom0.

### The solution you'd like

https://github.com/QubesOS/qubes-issues/issues/10648#issuecomment-3907465680

> > > Ideally, there would be a way to log dom0 qrexec calls instead of having to run a GUI domain for collecting the calls...
> > 
> > 
> > Sometimes for debugging I add logging in `qubes/api/__init__.py:QubesDaemonProtocol.respond()`. But it produces quite a lot of logs, so I don't think it should be there by default...
> 
> What about requiring some sort of opt-in to enable the extra logging? It would be easy to only add it if an environment variable is set.

Qmemman uses `/etc/qubes/qmemman.conf`, so maybe using another TOML file is expected. With your variant, I supposed that you'd set `Environment(File)?=` in a drop-in for `qubesd.service`, which would also work.

I am thinking of:

- Command-line option
- Configuration file if not set on the command-line.

Developers who want to debug Qrexec calls

_No response_

ben-grande:
The exceptions ProtocolError and PermissionDenied are both raised by qubesd to indicate various problems with requests. However, both cause clients to print the extremely unhelpful "Got empty response from qubesd" message.

ProtocolError must be used only for client *programming* errors (bugs), not for problems that the client could not have detected before making the request.

PermissionDenied must be used only for service authorization denials.

Therefore, a API handler can be arranged as:

- Validation: ProtocolError otherwise
- fire_event_for_permission(): PermissionDenied if unauthorized
- Action

Ideally, validation that may leak existence of a system property should be done after asking for administrative permission, but then it would not be possible to pass only safe values to "admin-api:" events.

If we are already leaking existence of a property, it makes sense to provide a useful exception class for it.

Fixes: https://github.com/QubesOS/qubes-issues/issues/10345
Fixes: https://github.com/qubesos/qubes-issues/issues/10689

---

It is a draft, just posting here for CI. It is more like an idea than a finished version. Logging messages should be okay now, it is only from trusted input.

From some open TODOs:

```
% rg 'TODO: ben'
qubes/api/admin.py
460:        # TODO: ben: info-leak: revision existence
522:        # TODO: ben: info-leak: pool existence
525:        # TODO: ben: info-leak: volume existence
799:        # TODO: ben: info-leak: pool existence
841:        # TODO: ben: info-leak: pool existence
849:        # TODO: ben: dangerous logging any ASCII value?
949:        # TODO: ben: info-leak: label existence
963:        # TODO: ben: info-leak: label existence
975:        # TODO: ben: info-leak: label existence
1008:        # TODO: ben: info-leak: label existence
1309:                # TODO: ben: info-leak: template existence
1341:                # TODO: ben: info-leak: label existence
1392:        # TODO: ben: info-disclosure: learns if qube exists
```

It seems that most checks that are done before `admin-permission` events may leak some data but not sure of its sensitiveness.

- `revision` is a date
- `volumes` are always the same
- `pool` and `labels` have a default that most people use
- `template` and `qube` existence, not so nice

But without these checks, there is no good error message...


qubes/exc.py
```diff
@@ -27,6 +27,40 @@ class QubesException(Exception):
     """Exception that can be shown to the user"""
 
 
+class ProtocolError(AssertionError):
+    """Raised deliberately by handlers to indicate a malformed client request.
+
+    Client programming errors. The client made a request that it should have
+    known better than to send in the first place. Includes things like passing
+    an argument or payload to a service documented to not take one. The only
+    way that a correctly behaving client can get ProtocolError is qubesd is
```
DemiMarie, line 36:
```suggestion
    way that a correctly behaving client can get ProtocolError is qubesd is if
```

qubes/exc.py
```diff
@@ -27,6 +27,40 @@ class QubesException(Exception):
...
+    The request is valid, but the client does not have permission to perform
+    the operation. Clients in dom0 should usually not get this error. In HTTP,
+    this would be 403 Forbidden.
+
+    This does not provide any useful information to the client making the
+    request. Therefore, it should only be raised if there is a client
+    *programming* error, such as passing an argument to a request that does not
+    take an argument. It should not be used to reject requests that are valid,
+    but which qubesd is refusing to process. Instead, raise a subclass of
+    :py:class:`QubesException` with a useful error message.
```
DemiMarie:
```suggestion
```

qubes/exc.py
```diff
@@ -322,3 +352,48 @@ class UnexpectedDeviceProperty(QubesException, ValueError):
...
+
+
+class QubesPoolNotFoundError(KeyError):
+    """Pool does not exist."""
+
+
+class QubesVolumeNotFoundError(KeyError):
+    """Pool does not exist."""
+
+
```
DemiMarie, line 394: Catching all of these requires catching `KeyError`, which seems a bit weird to me.  Would `StoragePoolException` be a better superclass?

qubes/api/admin.py
```diff
@@ -1871,7 +1996,10 @@ async def _load_backup_profile(self, profile_name, skip_passphrase=False):
                 )
                 # make it foolproof against "echo passphrase" implementation
                 passphrase = passphrase.strip()
-                self.enforce(b"\n" not in passphrase)
+                self.enforce(
+                    b"\n" not in passphrase,
+                    reason="Passphrase must not contain newline character",
+                )
```
DemiMarie, line 2329: Should there also be a check for NUL, or does qrexec do this implicitly?

qubes/exc.py
ben-grande: Thanks.

qubes/exc.py
ben-grande: And then make `class StoragePoolException(QubesException, KeyError)`? If yes, I agree.

qubes/api/admin.py
ben-grande: Tested, it allows null, `QubesVM.run_service`, which `run_service_for_stdio` uses, has `filter_esc=False` as default. If set to `True`, it substitutes by `_` which would become an invalid password if user has other prohibited characters in the UTF-8 set. This is a hard one if we want to allow UTF-8 characters in passphrases... switching it off now might break existing workflows, maybe allow the safe to print set you did on qrexec?

ben-grande: `scrypt` allows NUL and escape character as a passphrase. Tested with reading the passphrase from a file, to not have to deal with bash's handling. `scrypt` does prohibits `\n` though.

The `qubes.backup.Backup.backup_do`, encodes the passphrase in UTF-8, I don't know if conversion from higher sets to a lower set always works. Later, on `launch_scrypt`, does write the passphrase to the pty.

qubes/exc.py
ben-grande: Is this deleting the paragraph? Github interface doesn't make it obvious.

I tried to make sure that only sanitized and already trusted information, as well as non private information that the caller doesn't know, is sent in the exception message. There are various cases where this could go wrong, especially in the part about revealing more information to the caller than what it should have, but a proper error message should be done any way to avoid hard to debug errors such as [Got empty response from qubesd](https://github.com/qubesos/qubes-issues/issues/6739).

qubes/exc.py
ben-grande: > Would `StoragePoolException` be a better superclass?

No because it is a superclass of `QubesException`. Search for comments above each of these exceptions in `qubes/api/*`, I mention that it may reveal xistence of revisions, volumes and pools.

The issue is that we are doing checks to pass sanitized data to `fire_event_for_permission` for the `qubes.ext.admin`, this can reveal information about the system that the `admin-extension` could want to block. The current state is that Qrexec policy may not be restrictive enough, and the `admin-extension` is something to be a more precise ACL, as it is already in qubesd, and not on qrexec level.

ben-grande:
This PR got bigger than what I anticipated. I will try to restrict myself to:

1. Choosing ProtocolError and PermissionDenied correctly
2. Returning exceptions to the clients so problems can be debugged
3. Minimizing information leak of object existence before admin-permission (hard to do a on most cases, because some information needs to be sanitized before passing to `admin_for_permission()`).

---

For future PRs, somethings to tackle would be improving API security: https://github.com/QubesOS/qubes-doc/pull/1607/commits/753fa4ec304e7655336dce21c87dba4325a13ace, which, if no GSoC applicant takes it, might be assigned directly to me. I think this becomes more important when the non-dom0 GUI domain is streamlined.

ben-grande:
Oh, an this PR, because it changes the exception class, will break a lot of exception handling on qubesadmin and things that use it.

qubes/api/__init__.py
```diff
@@ -343,62 +436,46 @@ async def respond(self, src, meth, dest, arg, *, untrusted_payload):
...
             self.app.log.warning(
-                "permission denied for call %s+%s (%s → %s) "
-                "with payload of %d bytes",
+                exc_fmt,
                 meth,
                 arg,
                 src,
                 dest,
                 len(untrusted_payload),
+                exc_name,
```
ben-grande: Once review checks that no private information of untrusted value is passed to the exception, we should change to log the reason of ProtocolError in dom0. Currently, it is returned to the caller, but if debugging only in dom0, it is not possible to know what happened.

I believe that only safe values are in the message of ProtocolError, but a careful review on this would be nice.

qubes/api/__init__.py
```diff
@@ -231,49 +293,80 @@ def fire_event_for_permission(self, **kwargs):
...
+                )
+            )
         coefficient = 1
         if allow_negative and untrusted_size.startswith(b"-"):
             coefficient = -1
             untrusted_size = untrusted_size[1:]
+        name_cap = name.capitalize()
         if not untrusted_size.isdigit():
-            raise qubes.exc.ProtocolError("Size must be ASCII digits (only)")
+            raise ProtocolError(name_cap + " size contains non ASCII digits")
```
marmarek: In some uses of this function it isn't really a size - for example "revisions to keep" use this function too.

qubes/api/__init__.py
```diff
@@ -231,49 +293,80 @@ def fire_event_for_permission(self, **kwargs):
...
         if len(untrusted_size) >= 20:
-            raise qubes.exc.ProtocolError("Sizes limited to 19 decimal digits")
+            raise ProtocolError(
+                name_cap + " size is bigger than 19 decimal digits"
```
marmarek: here too

qubes/tools/qubesd.py
```diff
@@ -38,6 +38,19 @@ def sighandler(loop, signame, servers, app):
     help="Enable verbose error logging (all exceptions with full "
     "tracebacks) and also send tracebacks to Admin API clients",
 )
+parser.add_argument(
+    "--disable-debug-libvirt",
+    action="store_true",
+    default=False,
+    help="Disables verbose logging of libvirtaio, specifically the "
```
marmarek: Maybe go one step further and require extra option (on top of --debug) to enable this logging? IOW change the default (and rename). libvirtaio logs are rarely needed.

doc/qubes-api.rst
```diff
@@ -0,0 +1,138 @@
+:py:mod:`qubes.api` -- API
+=================================
+
+API sanitization is very important to maintain the AdminVM and everything it
+controls, secure.
+
+**Simple rules**:
+
+* Handle exceptions gracefully. Unhandled exceptions have the traceback only in
+  the AdminVM, not allowing clients to know how what happened. More information
```
marmarek:
```suggestion
  the AdminVM, not allowing clients to know what happened. More information
```

marmarek (changes requested):
I checked also the "info-leak" todo comments, they look acceptable to me, since they apply to a situation when the user already has at least partial access to the method, and the information is not very sensitive.

marmarek:
There is one line too long, otherwise looks good now

ben-grande:
Add unittests for coverage.

ben-grande:
PipelineRetry

marmarek (approved):

PR #782, opened 2026-02-26
Allow VMs to request the anon-timezone feature

### LINKED ISSUE #8381: Stop leaking dom0 timezone to Qubes-Whonix
### Qubes OS release

R4.2

### Brief summary

Qubes VMs leak timezone.

Reported by @[chessjazz](https://forums.whonix.org/u/chessjazz).

### Steps to reproduce

    qubesdb-read /qubes-timezone

### Expected behavior

No command available to leak dom0 timezone.

### Actual behavior

Dom0 timezone can be leaked in VM if malware is running inside the VM.

### Additional information

For issue tracking.

* issue caused by Qubes-Whonix: no
* affects Qubes-Whonix: yes, because Whonix sets timezone to UTC as it should be hidden. (It doesn't leak to remote websites but malware with local code execution could read dom0 timezone.)
* only relevant for Whonix: Dunno if there are also other users who would prefer not to leak this information to VMs.

### Suggested solution

If `qvm-features` or similar mechanism has `whonix-ws 1`, `whonix-gw 1`, `notimezone 1`, then don't write `/qubes-timezone` to  qubesdb.

ArrayBolt3:
This will allow virtual machines to hide dom0's timezone from themselves irreversibly. To prevent malware within a VM from undoing the feature once it's set, allow VMs to request the feature but do not let them drop it thereafter.

Fixes: https://github.com/QubesOS/qubes-issues/issues/8381

### DIFF
--- a/qubes/ext/core_features.py
+++ b/qubes/ext/core_features.py
@@ -194,6 +194,12 @@ class CoreFeatures(qubes.ext.Extension):
             ) and hasattr(vm, "appvm_default_bootmode"):
                 bootmode_value = untrusted_feature_value
                 vm.features["boot-mode.appvm-default"] = bootmode_value
+
+        # allow VMs to switch on anon-timezone, but do not permit dropping it
+        if "anon-timezone" in untrusted_features:
+            untrusted_value = untrusted_features["anon-timezone"]
+            if untrusted_value == "1":
+                vm.features["anon-timezone"] = True
         del untrusted_features
 
         # default user for qvm-run etc
--- a/qubes/tests/ext.py
+++ b/qubes/tests/ext.py
@@ -1677,6 +1677,47 @@ class TC_00_CoreFeatures(qubes.tests.QubesTestCase):
             ],
         )
 
+    def test_060_anon_timezone_set(self):
+        del self.vm.template
+        self.loop.run_until_complete(
+            self.ext.qubes_features_request(
+                self.vm,
+                "features-request",
+                untrusted_features={
+                    "anon-timezone": "1",
+                },
+            )
+        )
+        self.assertListEqual(
+            self.vm.mock_calls,
+            [
+                ("features.items", (), {}),
+                ("features.__setitem__", ("anon-timezone", True), {}),
+                ("features.get", ("qrexec", False), {}),
+                ("features.get", ("qrexec", False), {}),
+            ],
+        )
+
+    def test_061_anon_timezone_prevent_unset(self):
+        del self.vm.template
+        self.loop.run_until_complete(
+            self.ext.qubes_features_request(
+                self.vm,
+                "features-request",
+                untrusted_features={
+                    "anon-timezone": "0",
+                },
+            )
+        )
+        self.assertListEqual(
+            self.vm.mock_calls,
+            [
+                ("features.items", (), {}),
+                ("features.get", ("qrexec", False), {}),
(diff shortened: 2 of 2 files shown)


qubes/ext/core_features.py
```diff
@@ -194,6 +194,12 @@ async def qubes_features_request(self, vm, event, untrusted_features):
             ) and hasattr(vm, "appvm_default_bootmode"):
                 bootmode_value = untrusted_feature_value
                 vm.features["boot-mode.appvm-default"] = bootmode_value
+
+        # allow VMs to switch on anon-timezone, but do not permit dropping it
+        if "anon-timezone" in untrusted_features:
+            untrusted_value = untrusted_features["anon-timezone"]
+            if untrusted_value == "1":
```
ben-grande: please don't compare against "1", but use truthy statement, as in `if untrusted_value:`.

ArrayBolt3: Done.

qubes/ext/core_features.py
```diff
@@ -194,6 +194,10 @@ async def qubes_features_request(self, vm, event, untrusted_features):
             ) and hasattr(vm, "appvm_default_bootmode"):
                 bootmode_value = untrusted_feature_value
                 vm.features["boot-mode.appvm-default"] = bootmode_value
+
+        # allow VMs to switch on anon-timezone, but do not permit dropping it
+        if "anon-timezone" in untrusted_features:
```
marmarek, line 199: But now it's too far in the other direction - requesting `anon-timezone=0` would still _enable_ the feature - IMO should be ignored.

marmarek (changes requested):

qubes/ext/core_features.py
ArrayBolt3: I mean, that would have been the case if I had done what Ben requested exactly, i.e.:

```
        # allow VMs to switch on anon-timezone, but do not permit dropping it
        if "anon-timezone" in untrusted_features:
            untrusted_value = untrusted_features["anon-timezone"]
            if untrusted_value:
```

That's because `"0"` is equivalent to `True`:

```
┌─╴aaron@kf-m2g5:~
└─╴$ python3
Python 3.12.3 (main, Mar  3 2026, 12:15:18) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> if "0":
...     print("OK")
... 
OK
```

Maybe something like this would be better?

```
        # allow VMs to switch on anon-timezone, but do not permit dropping it
        if "anon-timezone" in untrusted_features:
            untrusted_value = untrusted_features["anon-timezone"]
            try:
                if int(untrusted_value):
                    ...
```

marmarek: hm, maybe it should remain `== "1"` after all? we do have something like this elsewhere: https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/ext/core_features.py#L91-L101

ArrayBolt3: I'm happy to switch back to that if that's preferable.

ArrayBolt3: Switched back.

qubes/ext/core_features.py
ArrayBolt3: Undone: https://github.com/QubesOS/qubes-core-admin/pull/782#discussion_r3005572136

marmarek (approved):

PR #785, opened 2026-03-06
tests: run dispvm preload tests once instead of per-template

### LINKED ISSUE #10749: Disposable integration tests (system_tests_dispvm) have repeated tests, making the run slower than it should
### Qubes OS release

Qubes OS 4.3

### Brief summary

These tests are run for every installed template, Debian, Fedora and Whonix, but they are not affected by the template running them:

- test_011_preload_reject_max
- test_012_preload_low_mem
- test_015_preload_race_more \*
- test_016_preload_race_less
- test_017_preload_autostart
- test_018_preload_global
- test_019_preload_discard_outdated_volumes
- test_020_preload_discard_outdated_volume_size
- test_021_preload_discard_outdated_setting

The `test_015_preload_race_more` test is affected by the OS as a template that is heavier, such as Whonix-Workstation, [will sometimes timeout](https://openqa.qubes-os.org/tests/169307#step/TC_20_DispVM_whonix-workstation-18/6), thus showing that the OS is slow. But this was not the intention of this test, so I think it is okay to remove this test from being shared across templates and let `system_tests_dispvm_perf@hw*` decide.

### Steps to reproduce

1. Run integration tests for disposables
2. Notice the repeated tests that are not covering more ground

### Expected behavior

No repeated tests, faster run.

### Actual behavior

Repated tests, slower run.

### Additional information

_No response_

therohityadav:
This patch isolates template-agnostic DispVM preload tests so they run only once instead of once per template.

* **`DispVMHelpersMixin`**: Moves shared setup, teardown, and helper logic into a base mixin without template inference.
* **`TC_20_DispVMMixin`**: Keeps the OS/GUI-dependent tests that run for each template.
* **`TC_21_DispVM_Preload`**: Adds a standalone class to run template-independent preload tests once.

The following tests are moved to run only once:
* `test_011_preload_reject_max`
* `test_012_preload_low_mem`
* `test_015_preload_race_more`
* `test_016_preload_race_less`
* `test_017_preload_autostart`
* `test_018_preload_global`
* `test_019_preload_discard_outdated_volumes`
* `test_020_preload_discard_outdated_volume_size`
* `test_021_preload_discard_outdated_setting`

Fixes QubesOS/qubes-issues#10749

### DIFF
--- a/qubes/tests/__init__.py
+++ b/qubes/tests/__init__.py
@@ -1698,9 +1698,7 @@ def list_templates():
     return _templates
 
 
-def create_testcases_for_templates(
-    name, *bases, module, default_template_only=False, **kwds
-):
+def create_testcases_for_templates(name, *bases, module, **kwds):
     """Do-it-all helper for generating per-template tests via load_tests proto
 
     This does several things:
@@ -1733,13 +1731,7 @@ def create_testcases_for_templates(
     # possible to correctly guess frame from stack. Explicit is better than
     # implicit!
 
-    templates = list_templates()
-    if default_template_only and templates:
-        # Some tests are template-agnostic and should run only once
-        # to avoid slowing down the integration test suite.
-        templates = templates[:1]
-
-    for template in templates:
+    for template in list_templates():
         clsname = name + "_" + template
         if hasattr(module, clsname):
             continue
--- a/qubes/tests/__init__.py
+++ b/qubes/tests/__init__.py
@@ -1698,7 +1698,9 @@ def list_templates():
     return _templates
 
 
-def create_testcases_for_templates(name, *bases, module, **kwds):
+def create_testcases_for_templates(
+    name, *bases, module, default_template_only=False, **kwds
+):
     """Do-it-all helper for generating per-template tests via load_tests proto
 
     This does several things:
@@ -1731,7 +1733,13 @@ def create_testcases_for_templates(name, *bases, module, **kwds):
     # possible to correctly guess frame from stack. Explicit is better than
     # implicit!
 
-    for template in list_templates():
+    templates = list_templates()
+    if default_template_only and templates:
+        # Some tests are template-agnostic and should run only once
+        # to avoid slowing down the integration test suite.
+        templates = templates[:1]
+
+    for template in templates:
         clsname = name + "_" + template
         if hasattr(module, clsname):
             continue
--- a/qubes/tests/integ/dispvm.py
+++ b/qubes/tests/integ/dispvm.py
@@ -196,19 +196,13 @@ class TC_04_DispVM(qubes.tests.SystemTestCase):
 
 
 class DispVMHelpersMixin:
-    def setUp(self):  # pylint: disable=invalid-name
-        logger.info("start")
-        super().setUp()
-        if "whonix-g" in self.template:
-            self.skipTest(
-                "whonix gateway is not supported as DisposableVM Template"
-            )
+    def setup_dispvm_nodes(self):
+        """Initialize disp_base and related attributes. Called by child setUp"""
         self.app.add_handler("domain-add", self._on_domain_add)
         self.addCleanup(
             self.app.remove_handler, "domain-add", self._on_domain_add
         )
         self.adminvm = self.app.domains["dom0"]
-        self.init_default_template(self.template)
         self.disp_base = self.app.add_new_vm(
             qubes.vm.appvm.AppVM,
             name=self.make_vm_name("dvm"),
@@ -523,6 +517,17 @@ class DispVMHelpersMixin:
 
 
 class TC_20_DispVMMixin(DispVMHelpersMixin):
+    def setUp(self):  # pylint: disable=invalid-name
+        logger.info("start")
+        super().setUp()
+        if "whonix-g" in self.template:
+            self.skipTest(
+                "whonix gateway is not supported as DisposableVM Template"
+            )
+        self.init_default_template(self.template)
+        self.setup_dispvm_nodes()
+        logger.info("end")
+
     def test_010_dvm_run_simple(self):
         dispvm = self.loop.run_until_complete(
             qubes.vm.dispvm.DispVM.from_appvm(self.disp_base)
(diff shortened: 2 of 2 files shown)


qubes/tests/integ/dispvm.py
```diff
@@ -195,20 +195,14 @@ def test_011_failed_start_timeout(self):
...
-            )
+class DispVMHelpersMixin:
+    def setup_dispvm_nodes(self):
+        """Initialize disp_base and related attributes. Called by child setUp"""
         self.app.add_handler("domain-add", self._on_domain_add)
         self.addCleanup(
             self.app.remove_handler, "domain-add", self._on_domain_add
         )
         self.adminvm = self.app.domains["dom0"]
-        self.init_default_template(self.template)
```
ben-grande, line 211: Why? This is making more changes than what the commit should be, solely related to having a reusable class and a separate class for preloaded disposable tests that runs on all templates from the ones that don't need to run on all templates, just one. I think it makes sense to have it in the helper class also.

qubes/tests/integ/dispvm.py
```diff
@@ -248,11 +242,10 @@ def setUp(self):  # pylint: disable=invalid-name
             "--",
             "qubesdb-read /name | tr -d '\n'",
         ]
-        logger.info("end")
```
ben-grande, line 251: https://github.com/QubesOS/qubes-core-admin/pull/785/changes#r2909703570

qubes/tests/integ/dispvm.py
```diff
@@ -248,11 +242,10 @@ def setUp(self):  # pylint: disable=invalid-name
...
 
     def tearDown(self):  # pylint: disable=invalid-name
         logger.info("start")
-        if "gui" in self.disp_base.features:
```
ben-grande, line 255: https://github.com/QubesOS/qubes-core-admin/pull/785/changes#r2909703570

qubes/tests/integ/dispvm.py
```diff
@@ -274,7 +267,7 @@ def _run_cmd_and_log_output(self, qube, cmd, user="root", timeout=30):
             )
         except subprocess.CalledProcessError as e:
             stdout = getattr(e, "stdout", str(e))
-        logger.critical("{}: {}: {}".format(qube.name, cmd, stdout))
```
ben-grande, line 277: https://github.com/QubesOS/qubes-core-admin/pull/785/changes#r2909703570

qubes/tests/integ/dispvm.py
```diff
@@ -343,8 +336,8 @@ def cleanup_preload(self):
             if "preload-dispvm-max" not in qube.features or qube not in [
                 self.app.domains["dom0"],
                 default_dispvm,
-                self.disp_base,
-                self.disp_base_alt,
```
ben-grande, line 347: https://github.com/QubesOS/qubes-core-admin/pull/785/changes#r2909703570

qubes/tests/integ/dispvm.py
```diff
@@ -1354,6 +1378,7 @@ def create_testcases_for_templates():
 
 def load_tests(loader, tests, _pattern):
     tests.addTests(loader.loadTestsFromNames(create_testcases_for_templates()))
+    tests.addTests(loader.loadTestsFromTestCase(TC_21_DispVM_Preload))
```
ben-grande: I think this won't run on any template. I think you should add some option to `qubes.tests.create_testcases_for_template()` to run the test on the default template. You might need to call `init_default_template()` from there.

ben-grande (changes requested):
Just to let you know, you are in the right track.

For reviewers, split view seems better than unified view in this instance.

qubes/tests/__init__.py
```diff
@@ -1731,7 +1733,13 @@ def create_testcases_for_templates(name, *bases, module, **kwds):
     # possible to correctly guess frame from stack. Explicit is better than
     # implicit!
 
-    for template in list_templates():
+    templates = list_templates()
+    if default_template_only and templates:
+        # Some tests are template-agnostic and should run only once
+        # to avoid slowing down the integration test suite.
+        templates = templates[:1]
```
ben-grande: This doesn't get the default template. But in case someone set `QUBES_TEST_TEMPLATES` for openqa, it may not match the `app.default_template`. Get the first template in that list if the variable is set, else, use `self.app.default_template` (self is not passed to that function, so it needs to be passed another way).

qubes/tests/integ/dispvm.py
```diff
@@ -1030,22 +1026,17 @@ def test_100_open_in_dispvm(self):
...
     default template.
-    """
 
-    def setUp(self):  # pylint: disable=invalid-name
-        logger.info("start")
-        self.init_default_template()
-        super().setUp()
-        self.setup_dispvm_nodes()
-        logger.info("end")
+    See QubesOS/qubes-issues#10749
```
ben-grande: I don't think there is need to have this comment. If you git blame the PR, you will see this issue in the commit message and in the PR message.

ben-grande (changes requested):

qubes/tests/__init__.py
```diff
@@ -1731,7 +1733,26 @@ def create_testcases_for_templates(name, *bases, module, **kwds):
...
+    if default_template_only and templates:
+        # Run template-agnostic tests only once. If OpenQA overrides the
+        # template list via QUBES_TEST_TEMPLATES, use the first template
+        # from that list. Otherwise resolve the system default template.
+        if "QUBES_TEST_TEMPLATES" in os.environ:
+            templates = templates[:1]
+        else:
+            try:
+                app = qubes.Qubes()
+                if app.default_template:
```
ben-grande: Marek, do you have a recommendation to avoid instantiating `Qubes()` here?

ben-grande (changes requested):

qubes/tests/__init__.py
marmarek: While I don't have an idea for avoiding creating `Qubes()` instance to list templates (in absence of `QUBES_TEST_TEMPLATES` variable), I don't think you need to do that at all. Tests that are not template-specific (that would run on all (selected) templates) simply use default template, and have no template name in the test instance name. There is absolutely no need to handle the case of default template not being set.
I'd argue even the above choosing first template from `QUBES_TEST_TEMPLATES` (instead of the default one) is not necessary.

ben-grande: @therohityadav I am sorry, it seems that I created this confusion (https://github.com/QubesOS/qubes-core-admin/pull/785#discussion_r2909720977). The old code seems correct.

qubes/tests/integ/dispvm.py
ben-grande: This and other things haven't been addressed. Please check all comments.

qubes/tests/integ/dispvm.py
```diff
@@ -1352,9 +1365,27 @@ def create_testcases_for_templates():
     )
 
 
+def create_preload_testcases_for_default_template():
+    return qubes.tests.create_testcases_for_templates(
```
marmarek: Tests that are supposed to be running just on one template don't need the mixin approach at all. Simply define class inheriting from `qubes.tests.SystemTestCase` directly and call `self.init_default_template()` in `setUp()`. To make the changes simpler, you can also add `self.template = self.app.default_template`, so you still have the `self.template` attribute like in the other tests.

And then there is no need calling `create_testcases_for_templates()` for those tests at all, as unittest/pytest/nose will already correctly discover them automatically.

This is an example of test class running on a single (default) template: https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/tests/integ/basic.py#L52

marmarek (changes requested):

marmarek (approved):

PR #786, opened 2026-03-08
Add label to system_info

notmuchtohide:
This is necessary to display qube colors in qrexec-policy-graph (https://github.com/QubesOS/qubes-core-qrexec/pull/222).

### DIFF
--- a/qubes/api/internal.py
+++ b/qubes/api/internal.py
@@ -53,6 +53,8 @@ class SystemInfoCache:
         "property-reset:default_dispvm",
         "property-set:icon",
         "property-reset:icon",
+        "property-set:label",
+        "property-reset:label",
         "property-set:guivm",
         "property-reset:guivm",
         "property-set:relayvm",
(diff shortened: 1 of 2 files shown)


qubes/api/internal.py
```diff
@@ -131,6 +131,7 @@ def get_system_info(cls, app):
                         else None
                     ),
                     "icon": str(domain.label.icon),
+                    "label": str(domain.label.color),
```
ben-grande, line 136: See `vm_events` at the top of the file. See how `icon` was handled.

ben-grande:
Will schedule this for openqa. Although tests are failing, it is because the test was not adapted to consider label. The code is fine. I still expect the test to be fixed.

qubes/api/internal.py
notmuchtohide: Done! Thanks for the tip.

marmarek (approved):
it would be better to have this in one commit indeed, but it isn't too bad here

PR #789, opened 2026-03-15
Increase dom0 scheduling weight

### LINKED ISSUE #10778: dom0 doesn’t get extra scheduler credit
### Qubes OS release

Qubes OS 4.3

### Brief summary

Dom0 has always been supposed to have extra scheduler credit, and the code for this goes all the way back to the first public commit to the core-admin repository.  However, it broke once Xen was upgraded to a version that defaults to the credit2 scheduler.

This significantly reduces dom0’s responsiveness when starting many VMs.  I reported this in #10773 and this was found to be at least one reason.  I increased dom0’s credit in my `~/.profile` and it makes a very noticeable difference.

### Steps to reproduce

1. Run `xl sched-credit2` in dom0.
2. Look at the weight of dom0.

### Expected behavior

Dom0 has an elevated weight.

### Actual behavior

Dom0 has the same weight as all other VMs.

### Additional information

Increasing dom0 scheduler credit was recommended by @marmarek in <https://github.com/QubesOS/qubes-issues/issues/10773#issuecomment-4062049241>.  In <https://github.com/QubesOS/qubes-issues/issues/10773#issuecomment-4062896294>, he pointed out that dom0 was always supposed to have extra scheduling credit.

DemiMarie:
Dom0 is responsible for the UI and for critical background activities. In particular, it is on the critical path for VM start.  Therefore, it should be prioritized over all other VMs to avoid priority inversion.

Dom0 has been intended to have additional credit since a17989470afc ("Initial public commit."), but nowadays Xen uses the credit2 scheduler, so changing the parameters of the credit scheduler has no effect.

Fixes: QubesOS/qubes-issues#10778

### DIFF
--- a/linux/aux-tools/startup-misc.sh
+++ b/linux/aux-tools/startup-misc.sh
@@ -7,6 +7,7 @@ DOM0_MAXMEM=$(/usr/sbin/xl list 0 | tail -1 | awk '{ print $3 }')
 xenstore-write /local/domain/0/memory/static-max $[ $DOM0_MAXMEM * 1024 ]
 
 xl sched-credit -d 0 -w 2000
+xl sched-credit2 -d 0 -w 2000
 cp /var/lib/qubes/qubes.xml /var/lib/qubes/backup/qubes-$(date +%F-%T).xml
 
 /usr/lib/qubes/cleanup-dispvms


marmarek:
> Fixes: [a179894](https://github.com/QubesOS/qubes-core-admin/commit/a17989470afc0613847cd7597ea4f74827a2739e) ("Initial public commit.")

This is wrong, it was correct at that time.

marmarek (approved):

PR #790, opened 2026-03-16
admin: validate feature name in vm_feature_remove

### LINKED ISSUE #7186: qubesd: prohibit nonsensical feature names

qubesd doesn’t do much validation of feature names.  When they are set from a qube, they are limited by qrexec to a reasonable list of safe characters.  However, one can set invalid feature names from dom0 easily.  The results are rather amusing: `qvm-service someqube ../../../../../root/gotcha on` causes `someqube` to create an empty `/root/gotcha` file when it starts next, and `qvm-service someqube ../../../../../root/gotcha off` causes `someqube` to delete that file.

This is not a security issue because dom0 is ultimately trusted anyway.

### The solution you'd like

qubesd should disallow features from being set by dom0 that could not also be set via the Admin API.  This helps avoid mistakes and increases the likelihood that tools that currently run in dom0 will keep working via the Admin API.

### The value to a user, and who that user might be

Users who set features manually will get better errors.

Tab-Tarek3600:
Feature name validation was missing in vm_feature_remove, while vm_feature_set already validated it. This inconsistency allowed path traversal characters in feature names when removing features directly via dom0.

Also use ProtocolError instead of QubesValueError for invalid feature names in vm_feature_set, consistent with PR #751.

Fixes: QubesOS/qubes-issues#7186
Related: QubesOS/qubes-core-admin#751

### DIFF
--- a/qubes/api/__init__.py
+++ b/qubes/api/__init__.py
@@ -28,6 +28,7 @@ import re
 import shutil
 import socket
 import struct
+import string
 import traceback
 from typing import Union, Any
 import uuid
@@ -165,8 +166,19 @@ class AbstractQubesAPI:
         self.dest = decode_vm(dest, app.domains)
 
         #: argument
-        self.arg = arg.decode("ascii")
-
+        untrusted_arg = arg.decode("ascii")
+        # Validate arg against qrexec allowed characters
+        # Consistent with sanitize_name() in qrexec-daemon.c
+        _allowed_arg_chars = string.ascii_letters + string.digits + "-+_."
+        if untrusted_arg and not all(
+            c in _allowed_arg_chars for c in untrusted_arg
+        ):
+            raise ProtocolError(
+                "Argument contains characters outside safe set: "
+                + _allowed_arg_chars
+            )
+        self.arg = untrusted_arg
+        del untrusted_arg
         #: name of the method
         self.method = method_name.decode("ascii")
 
@@ -232,7 +244,7 @@ class AbstractQubesAPI:
             pre_event=True,
             dest=self.dest,
             arg=self.arg,
-            **kwargs
+            **kwargs,
         )
 
     def fire_event_for_filter(self, iterable, **kwargs):
--- a/qubes/api/admin.py
+++ b/qubes/api/admin.py
@@ -100,6 +100,18 @@ class QubesMgmtEventsDispatcher:
         vm.remove_handler("*", self.vm_handler)
 
 
+def _validate_feature_name(name):
+    """Validate feature name against allowed character set.
+
+    Allowed characters: alphanumeric, dot, underscore, hyphen.
+    Raises ProtocolError if name is invalid.
+    """
+    if not re.match(r"\A[a-zA-Z0-9_.+-]+\Z", name):
+        raise qubes.exc.ProtocolError(
+            "feature name contains illegal characters"
+        )
+
+
 class QubesAdminAPI(qubes.api.AbstractQubesAPI):
     """Implementation of Qubes Management API calls
 
@@ -1162,8 +1174,7 @@ class QubesAdminAPI(qubes.api.AbstractQubesAPI):
         "admin.vm.feature.Remove", no_payload=True, scope="local", write=True
     )
     async def vm_feature_remove(self):
-        # validation of self.arg done by qrexec-policy is enough
-
+        _validate_feature_name(self.arg)
         self.fire_event_for_permission()
         try:
             del self.dest.features[self.arg]
@@ -1175,10 +1186,7 @@ class QubesAdminAPI(qubes.api.AbstractQubesAPI):
     async def vm_feature_set(self, untrusted_payload):
         untrusted_value = untrusted_payload.decode("ascii", errors="strict")
         del untrusted_payload
-        if re.match(r"\A[a-zA-Z0-9_.-]+\Z", self.arg) is None:
-            raise qubes.exc.QubesValueError(
-                "feature name contains illegal characters"
-            )
+        _validate_feature_name(self.arg)
         if re.match(r"\A[\x20-\x7E]*\Z", untrusted_value) is None:
             raise qubes.exc.QubesValueError(
                 f"{self.arg} value contains illegal characters"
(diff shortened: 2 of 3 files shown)


qubes/api/__init__.py
```diff
@@ -166,6 +166,10 @@ def __init__(
 
         #: argument
         self.arg = arg.decode("ascii")
+        # Validate arg: must only contain characters allowed by qrexec
+        # See sanitize_name() in qrexec-daemon.c
+        if self.arg and not re.match(r"\A[\w.-]+\Z", self.arg):
```
ben-grande: Tests are failing and on this thread you can find a link pointing to all the characters that are allowed in qrexec.

Please don't [\w this](https://docs.python.org/3/library/re.html) (unless you use the ASCII flag) or be explicit like in qrexec-daemon.c.

> \w
> For Unicode (str) patterns:
> Matches Unicode word characters; this includes all Unicode alphanumeric characters (as defined by str.isalnum()), as well as the underscore (_).
> 
> Matches [a-zA-Z0-9_] if the ASCII flag is used.
> 
> For 8-bit (bytes) patterns:
> Matches characters considered alphanumeric in the ASCII character set; this is equivalent to [a-zA-Z0-9_]. If the LOCALE flag is used, matches characters considered alphanumeric in the current locale and the underscore.

ben-grande: I don't think exceptions from __init__ are returned to caller and logged correctly. Check other mentions of `ProtocolError` on this file.

qubes/api/admin.py
```diff
@@ -100,6 +100,18 @@ def on_domain_delete(self, subject, event, vm):
         vm.remove_handler("*", self.vm_handler)
 
 
+def _validate_feature_name(name):
+    """Validate feature name against allowed character set.
+
+    Allowed characters: alphanumeric, dot, underscore, hyphen.
+    Raises ProtocolError if name is invalid.
+    """
+    if not re.match(r"\A[a-zA-Z0-9_.\-]+\Z", name):
```
ben-grande: I don't think that escaping `-` is necessary if it is the last item. Did you test without it?

It's still missing a character. Check the link provided by Marek here: https://github.com/QubesOS/qubes-core-admin/pull/790#issuecomment-4067773086. The regex should also be in the argument validation.

qubes/tests/api_admin.py
```diff
@@ -1694,6 +1694,27 @@ def test_301_feature_remove_none(self):
...
+            )
+        self.assertFalse(self.app.save.called)
+
+    def test_302b_feature_set_invalid_name(self):
+        # Invalid feature names must raise ProtocolError in vm_feature_set
+        with self.assertRaises(qubes.exc.ProtocolError):
+            self.call_mgmt_func(
+                b"admin.vm.feature.Set",
+                b"test-vm1",
+                b"../../../root/gotcha",
```
ben-grande: This is not testing the issue https://github.com/qubesos/qubes-issues/issues/7186. The services feature starts with the `service.` prefix.

---

Technically, `/` slash is not dangerous for all feature, but only to service features. As qrexec doesn't allow that character, I think it should remain blocked.

ben-grande:
About tripping on device API with invalid characters, I presume discovered with unittests. This means that it is currently broken when using a GUI domain. Marek, will Qrexec daemon extend the allowed arguments to include special shell characters or is there a planned solution for it?

ben-grande:
> I fixed an issue like this before: [QubesOS/qubes-issues#10040](https://github.com/QubesOS/qubes-issues/issues/10040), is there some more places where not allowed characters are used for devices?

Taleb, please elucidate the reason for the regex change.

qubes/api/__init__.py
```diff
@@ -166,6 +166,10 @@ def __init__(
 
         #: argument
         self.arg = arg.decode("ascii")
+        # Validate arg against qrexec allowed characters
+        # Consistent with sanitize_name() in qrexec-daemon.c
+        if self.arg and not re.match(r"\A[a-zA-Z0-9_.+-]+\Z", self.arg):
+            raise ProtocolError(f"arg not allowed in qrexec: {self.arg!r}")
```
ben-grande: - I don't see f-strings being used on this file. Please use another formatting method that is already present on this file
- Please don't store invalid `self.arg`
- Please adjust the message to explain why the request is invalid. "not allowed" feels like PermissionDenied. A better message would say that "Argument contains characters outside safe set: ALLOWED_CHARS". I used set of characters, because in this case, regex is not necessary. Example (used similarly on This is the same method I used on api/admin in https://github.com/QubesOS/qubes-core-admin/pull/751):
    ```python3
    allowed_chars = string.ascii_letters + string.digits + "-+_."
    if self.arg and not all(c in allowed_chars for c in self.arg):
        raise qubes.exc.ProtocolError(
            "Argument must be in safe set: " + allowed_chars
        )
    ```

Is regex matching faster?

marmarek:
PipelineRetry

ben-grande:
> @ben-grande this will most likely conflict with your #751 , are you okay with merging this one first?

Yes.

qubes/api/__init__.py
```diff
@@ -165,8 +166,18 @@ def __init__(
...
+        # Consistent with sanitize_name() in qrexec-daemon.c
+        _allowed_arg_chars = string.ascii_letters + string.digits + "-+_."
+        if untrusted_arg and not all(
+            c in _allowed_arg_chars for c in untrusted_arg
+        ):
+            raise ProtocolError(
+                "Argument contains characters outside safe set: "
+                + _allowed_arg_chars
+            )
+        self.arg = untrusted_arg
```
ben-grande, line 180: End with `del untrusted_arg`.

ben-grande (changes requested):
Please squash the commits.

ben-grande:
I'd like to manually test this PR before it is merged, so please wait before accepting it.

ben-grande:
> I don't think exceptions from **init** are returned to caller and logged correctly. Check other mentions of `ProtocolError` on this file.

Tested. It seems that it is correctly returned to the client even being on `__init__`. Of course it requires this change: https://github.com/QubesOS/qubes-core-admin/pull/751/changes/63a406ca5e4b109b5c53284ff5d283b7250ad26d.

PR is decent.

ben-grande:
Can you please solve the merge conflicts, since https://github.com/QubesOS/qubes-core-admin/pull/751 was merged prior to this one.

PR #791, opened 2026-03-19
Propagate vm-config.net-* features through NetVM chain

### LINKED ISSUE #10758: Make Qubes MTU configurable / inherit from sys-net uplink MTU

Some networks might need a different MTU than 1500. (in my case I need to setup1492). There is no clean way of configuring this MTU up to the qubes.
My PPPoE uplink has a MTU of 1492. Having 1500 in the rest of the LAN was breaking partially IPv6 connectivity.

My local DHCP server is sending the MTU option `dhcp-option=26,1492` to force clients setting the correct MTU. This MTU is also set on the eth0 interface of sys-usb. 
But everything behind that keeps the default MTU of 1500.

I found this thread on the forum https://forum.qubes-os.org/t/changing-default-mtu-for-networking-vms/16902 which doesn't feel like a proper solution.

### The solution you'd like

I guess, propagating the MTU from eth0 on sys-net to the rest of the qubes connected to it, would be a good way for working out of the box. 

Avoiding situations where the MTU do not match, bringing IPv6 to drop packets on the GW, because it cannot fragment it to send over the next link with smaller MTU

(This section is for developer use only. Please do not modify it.)

Codingisinmyblud:
This patch fixes the Qubes OS bug #10758, propagating vm-config.net-* features to the descendant qubes in the NetVM chain, thus making the vm-config.net-* features visible in the Qubes DB of the descendant qubes.

In the current implementation, if a NetVM has a vm-config.net-mtu, the descendant qubes will not inherit the vm-config.net-mtu automatically. However, the in-qube networking scripts of the descendant qubes read the configuration from the Qubes DB.

This patch introduces the propagation of the vm-config.net-* features from the NetVM chain to the descendant qubes. The propagation code has been introduced in the file qubes/vm/mix/net.py. The propagation code performs the following actions:

on domain-qdb-create: the qube inherits the effective vm-config.net-* features from the NetVM chain

on domain-feature-set:*, the vm-config.net-* features of the NetVM are propagated to the running descendant qubes, except when overridden

on domain-feature-delete:*, the descendants of the NetVM inherit the next upstream vm-config.net-* feature or delete the entry if there are no more upstream features

Fixes: https://github.com/QubesOS/qubes-issues/issues/10758

### DIFF
--- a/qubes/vm/mix/net.py
+++ b/qubes/vm/mix/net.py
@@ -147,6 +147,13 @@ class NetVMMixin(qubes.events.Emitter):
         doc="IPv6 address of this domain.",
     )
 
+    mtu = qubes.property(
+        "mtu",
+        type=int,
+        default=None,
+        doc="Maximum Transmission Unit (MTU) for the virtual interface.",
+    )
+
     # CORE2: swallowed uses_default_netvm
     netvm = qubes.VMProperty(
         "netvm",
@@ -271,6 +278,17 @@ class NetVMMixin(qubes.events.Emitter):
     #
     # used in both
     #
+
+    @qubes.stateless_property
+    def effective_mtu(self):
+        """Effective MTU for this domain. Inherited from netvm if not
+        set explicitly."""
+        netvm = self
+        while netvm is not None:
+            if hasattr(netvm, "mtu") and netvm.mtu is not None:
+                return netvm.mtu
+            netvm = getattr(netvm, "netvm", None)
+        return None
 
     @qubes.stateless_property
     def dns(self):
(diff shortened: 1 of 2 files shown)


qubes/vm/mix/net.py
```diff
@@ -272,6 +279,17 @@ def connected_vms(self):
     # used in both
     #
 
+    @qubes.stateless_property
+    def effective_mtu(self):
+        """Effective MTU for this domain. Inherited from netvm if not
+        set explicitly."""
+        netvm = self
```
ben-grande, line 286: Please don't refer to a qube that may not be a `netvm` as such, it makes reading a bit confusing.

qubes/vm/mix/net.py
```diff
@@ -785,3 +803,46 @@ def firewall(self):
...
         """Return `True` if there are some vm specific firewall rules set"""
         return os.path.exists(os.path.join(self.dir_path, self.firewall_conf))
+
+    @qubes.events.handler("domain-qdb-create")
+    def on_domain_qdb_create_mtu(self, event):
+        """Copies effective MTU to this VM's QubesDB."""
+        # pylint: disable=unused-argument
+        eff_mtu = self.effective_mtu
+        if eff_mtu is not None:
+            self.untrusted_qdb.write("/vm-config/net-mtu", str(eff_mtu))
```
ben-grande, line 813: I don't think it needs to be in the `vm-config` namespace anymore.

ben-grande (changes requested):

qubes/vm/mix/net.py
```diff
@@ -272,6 +279,17 @@ def connected_vms(self):
...
+    @qubes.stateless_property
+    def effective_mtu(self):
+        """Effective MTU for this domain. Inherited from netvm if not
+        set explicitly."""
+        netvm = self
+        while netvm is not None:
+            if hasattr(netvm, "mtu") and netvm.mtu is not None:
+                return netvm.mtu
+            netvm = getattr(netvm, "netvm", None)
+        return None
```
marmarek, line 291: Alternative would be to build it into `mtu` property default value itself, I think it would be easier this way. Something like:

```python
def default_mtu(self):
    if self.netvm:
        return self.netvm.mtu
    return None
```

and then use that function as `default=` in the property definition.

qubes/vm/mix/net.py
```diff
@@ -785,3 +803,46 @@ def firewall(self):
...
+        "property-set:netvm",
+        "property-reset:netvm",
+    )
+    def on_property_set_mtu(
+        self, event, name, newvalue=None, oldvalue=None, **kwargs
+    ):
+        """Propagate mtu changes to connected VMs."""
+
+        # pylint: disable=unused-argument
+        def _update_descendants(netvm):
```
marmarek, line 827: Following the single mtu property pattern proposed above, this may fire `property-reset:mtu` event on VMs with `vm.property_is_default("mtu") (instead of checking for `not None`), and the rest be moved outside of the loop (to do it only for vm that got the event, not all its descendants - this would be handled by firing event there).

ben-grande (changes requested):
Minor review.

ben-grande:
Hi. Do you plan to continue on this?

Codingisinmyblud:
Yeah, sorry just got a little busy. Should send a fix in a bit hopefully

PR #757, opened 2025-11-26
Wait for user session for preloaded disposables

ben-grande:
With the GUI agent patch, it can start before the GUI daemon connects, allowing the user session to complete. Wait both services to guarantee no enabled user or system service tries to start after the preload is used.

Requires: https://github.com/QubesOS/qubes-gui-agent-linux/pull/251
Requires: https://github.com/QubesOS/qubes-gui-agent-linux/pull/255
Requires: https://github.com/QubesOS/qubes-core-agent-linux/pull/647
Requires: https://github.com/QubesOS/qubes-core-qrexec/pull/229
Fixes: https://github.com/QubesOS/qubes-issues/issues/9940
For: https://github.com/QubesOS/qubes-issues/issues/1512

PR #796, opened 2026-04-06
Close qmemman client with correct handler

ben-grande:

---

https://github.com/QubesOS/qubes-core-admin/pull/757#issuecomment-4195354520

### DIFF
--- a/qubes/vm/dispvm.py
+++ b/qubes/vm/dispvm.py
@@ -611,32 +611,45 @@ class DispVM(qubes.vm.qubesvm.QubesVM):
             return
         if self.preload_requested:
             return
-        qmemman_client = None
         break_task = asyncio.create_task(self.preload_requested_event.wait())
-        qmemman_task = asyncio.get_event_loop().run_in_executor(
-            None, self.set_mem
+        qmemman_client = qubes.qmemman.client.QMemmanClient()
+        qmemman_task = asyncio.get_running_loop().run_in_executor(
+            None,
+            qmemman_client.set_mem,  # type: ignore[arg-type]
+            {self.xid: 0},
         )
-        tasks = [break_task, qmemman_task]
+        tasks: list = [break_task, qmemman_task]
+        result = None
+        cancelled = False
+        self.log.info("Setting qube memory to pref mem")
         try:
             # CI uses Python 3.12 and asynchronous iterator requires >=3.13
             # pylint: disable=not-an-iterable
             async for earliest_task in asyncio.as_completed(tasks):
                 await earliest_task
                 if earliest_task == break_task:
+                    self.log.info(
+                        "Canceling ballooning task, server might continue"
+                    )
+                    cancelled = True
                     qmemman_task.cancel()
                 else:
+                    result = qmemman_task.result()
                     break_task.cancel()
         except asyncio.CancelledError:
-            if qmemman_client:
-                qmemman_client.close()
+            pass
+        except IOError as e:
+            raise IOError("Failed to connect to qmemman: {!s}".format(e))
         except Exception as exc:
             self.log.warning(
                 "Preload memory request before pause failed: %s", str(exc)
             )
-            if qmemman_client:
-                qmemman_client.close()
             raise
         finally:
+            if qmemman_client.sock:
+                qmemman_client.close()
+            if not result or cancelled:
+                self.log.warning("Failed to set memory")
             if self.preload_requested:
                 raise qubes.exc.QubesVMCancelledPauseError(
                     self,
--- a/qubes/vm/qubesvm.py
+++ b/qubes/vm/qubesvm.py
@@ -2097,27 +2097,6 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.LocalVM):
 
         return qmemman_client
 
-    def set_mem(self):
-        """
-        Balloon qube to preferred memory, which is just enough for it to be
-        running without problems.
-        """
-        if not qmemman_present or self.maxmem == 0:
-            return None
-
-        self.log.info("Setting qube memory to pref mem")
-        qmemman_client = qubes.qmemman.client.QMemmanClient()
-        try:
-            result = qmemman_client.set_mem({self.xid: 0})
-        except IOError as e:
-            raise IOError("Failed to connect to qmemman: {!s}".format(e))
-
-        if not result:
-            qmemman_client.close()
-            self.log.warning("Failed to set memory")
-
-        return qmemman_client
-
     @staticmethod
     async def start_daemon(*command, input=None, **kwargs):
         """Start a daemon for the VM


qubes/vm/dispvm.py
```diff
@@ -627,14 +626,14 @@ async def on_domain_pre_paused(self, event, **kwargs) -> None:
                 else:
                     break_task.cancel()
         except asyncio.CancelledError:
-            if qmemman_client:
-                qmemman_client.close()
+            if qmemman_task:
+                qmemman_task.close()
```
marmarek: You sure about this? This is just a task that would return qmemman client when completed. Maybe `qmemman_task.result().close()` would work, but only if the task was completed already. If it was cancelled, I'm not sure how to get the client reference in this code shape...

ben-grande: ~In the upstream docs, it uses `create_task()` and then [task.cancel()](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel).~ I was reading the `break_task`...

About `result()`, I think it works for `asyncio.CancelledError`, but for other exceptions that didn't cancel the task, it [will raise an exception](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.result).

 I am going to test these possibilities to be sure, by delaying the qmemman response with a ten seconds sleep and making a call to use the preloaded disposable, which would cancel the call. Then on another run, I can make the call fail by reporting that it did fail to free memory from qube.

ben-grande: > Maybe `qmemman_task.result().close()` would work, but only if the task was completed already. If it was cancelled, I'm not sure how to get the client reference in this code shape...

`qmeman_task.cancel()` is already being use if `preload_request_event` completes/is set, so I don't think I need to use `.result()`.

About cancelling a task twice,, as `asyncio.CancelledError` was already raised, yes, that seems weird.

ben-grande: Hm. Maybe I left the `qmemman_client.close()` inside `asyncio.CancelledError` because the cancelled task might be the `breask_task`, so in every failed scenario, I want the `qmemman_task` to be cancelled.

ben-grande:
```
  File "/usr/lib/python3.13/site-packages/qubes/vm/dispvm.py", line 623, in on_domain_pre_paused
    await earliest_task
  File "/usr/lib64/python3.13/asyncio/locks.py", line 213, in wait
    await fut
asyncio.exceptions.CancelledError
During handling of the above exception, another exception occurred:
  File "/usr/lib/python3.13/site-packages/qubes/events.py", line 243, in fire_event_async
    effect = task.result()
  File "/usr/lib/python3.13/site-packages/qubes/vm/mix/dvmtemplate.py", line 535, in on_domain_preload_dispvm_used
    await asyncio.gather(
    ...<4 lines>...
    )
  File "/usr/lib/python3.13/site-packages/qubes/vm/dispvm.py", line 777, in from_appvm
    dispvm = await cls.gen_disposable(appvm, preload=preload, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/qubes/vm/dispvm.py", line 822, in gen_disposable
    await dispvm.start()
  File "/usr/lib/python3.13/site-packages/qubes/vm/dispvm.py", line 985, in start
    await super().start(**kwargs)
  File "/usr/lib/python3.13/site-packages/qubes/vm/qubesvm.py", line 1553, in start
    await self.fire_event_async(
        "domain-start", start_guid=start_guid
    )
  File "/usr/lib/python3.13/site-packages/qubes/events.py", line 243, in fire_event_async
    effect = task.result()
  File "/usr/lib/python3.13/site-packages/qubes/vm/dispvm.py", line 586, in on_domain_started_dispvm
    await self.pause()
  File "/usr/lib/python3.13/site-packages/qubes/vm/qubesvm.py", line 1754, in pause
    await self.fire_event_async("domain-pre-paused", pre_event=True)
  File "/usr/lib/python3.13/site-packages/qubes/events.py", line 243, in fire_event_async
    effect = task.result()
  File "/usr/lib/python3.13/site-packages/qubes/vm/dispvm.py", line 631, in on_domain_pre_paused
    qmemman_task.close()
    ^^^^^^^^^^^^^^^^^^
AttributeError: '_asyncio.Future' object has no attribute 'close'
```

This is inside `asyncio.CancelledError`.

marmarek: To be clear: I mean `QMemmanClient.close()` needs to be called at some point, not just the asyncio task (of calling `QMemmanClient.set_mem()`) cancelled.

ben-grande: > You sure about this? This is just a task that would return qmemman client when completed. Maybe `qmemman_task.result().close()` would work, but only if the task was completed already. If it was cancelled, I'm not sure how to get the client reference in this code shape...

Hm... `qmemman_task.result().close()` works inside `asyncio.CancelledError` when cancelling the `break_task` and when `qmemman_task` was already cancelled (by completing the `break_task`). I will add some debugging steps to qmemman to see if it is fine, but I think it is.

ben-grande: I ended up using the recommended `.result().done()`. I tried to break in some ways by not completing the task (any other exception), by having already cancelled the task and trying to cancel it, but it appears to work (stops the memory request and releases qmemman lock).

ben-grande: > To be clear: I mean `QMemmanClient.close()` needs to be called at some point, not just the asyncio task (of calling `QMemmanClient.set_mem()`) cancelled.

The disconnection also happens when length of received data is 0, this is why it wasn't hanging on most cases before this PR and with this PR.

qubes/vm/dispvm.py
```diff
@@ -627,14 +626,14 @@ async def on_domain_pre_paused(self, event, **kwargs) -> None:
                 else:
                     break_task.cancel()
         except asyncio.CancelledError:
-            if qmemman_client:
-                qmemman_client.close()
+            if qmemman_task:
+                qmemman_task.result().close()
```
marmarek, line 632: So, if `earliest_task == break_task`, this doesn't work, because qmemman_task is cancelled. With added try/except around this line I get:
```
Apr 12 16:26:13 dom0 qubesd[10498]: Traceback (most recent call last):
Apr 12 16:26:13 dom0 qubesd[10498]:   File "/usr/lib/python3.13/site-packages/qubes/vm/dispvm.py", line 634, in on_domain_pre_paused
Apr 12 16:26:13 dom0 qubesd[10498]:     await earliest_task
Apr 12 16:26:13 dom0 qubesd[10498]: asyncio.exceptions.CancelledError
Apr 12 16:26:13 dom0 qubesd[10498]: During handling of the above exception, another exception occurred:
Apr 12 16:26:13 dom0 qubesd[10498]: Traceback (most recent call last):
Apr 12 16:26:13 dom0 qubesd[10498]:   File "/usr/lib/python3.13/site-packages/qubes/vm/dispvm.py", line 643, in on_domain_pre_paused
Apr 12 16:26:13 dom0 qubesd[10498]:     qmemman_task.result().close()
Apr 12 16:26:13 dom0 qubesd[10498]:     ~~~~~~~~~~~~~~~~~~~^^
Apr 12 16:26:13 dom0 qubesd[10498]: asyncio.exceptions.CancelledError
```

marmarek (changes requested):

marmarek:
Looking at it closer, I think there is an issue with how this is called in the non-requested case. set_mem() will instruct qmemman to reduce memory of the qube. So far so good. But as soon as the client disconnects, qmemman will start balancing memory again, likely giving just reduced qube more memory again. IIUC that disconnect happens currently only implicitly, when garbage collector releases qmemman_client object, which is IMO quite fragile. Theoretically, the qmemman socket should be kept open until after the qube is paused. In practice, the qube is paused fast enough so it doesn't change much.

See how similar qmemman handling is done during qube startup: first qubesd request to free some memory for the new qube, then starts it and only then close the qmemman socket - otherwise qmemman would reuse that freed memory and there wouldn't be space for starting that new qube anymore.

But here, it isn't a single function, it would be two separate event handlers - `domain-pre-paused` and `domain-paused`. So, qmemman_client instance would need to be passed from one to the other somehow. Maybe on the VM object as some attribute? If done in `set_mem()` itself, it would allow fixing also the other issue, as you wouldn't need a result of a (cancelled) task anymore.

marmarek:
But the flip side of the "store qmemman_client reference in an attribute" is that, if something goes wrong and close() isn't called, garbage collector won't save you anymore. So, it wouldn't be "delayed for 20min" anymore, but qmemman would stop doing anything forever (which includes breaking starting further qubes). Or at least until responsible qube is removed (or maybe until qubesd or qmemman restart).

qubes/vm/dispvm.py
ben-grande: > So, if `earliest_task == break_task`, this doesn't work, because qmemman_task is cancelled. With added try/except around this line I get:

On this comment: https://github.com/QubesOS/qubes-core-admin/pull/796#discussion_r3042549558, I expected it to fail also, but it succeeded.

So what do you mean with added try/except? Is this:

```python3
        except asyncio.CancelledeError:
            if qmemman_task:
                try:
                    qmemman_task.result().close()
                except:
                    raise
```

ben-grande:
> But here, it isn't a single function, it would be two separate event handlers - `domain-pre-paused` and `domain-paused`. So, qmemman_client instance would need to be passed from one to the other somehow. Maybe on the VM object as some attribute? If done in `set_mem()` itself, it would allow fixing also the other issue, as you wouldn't need a result of a (cancelled) task anymore.

I remember you voicing your concern about qmemman trying to give memory back to the domain before it being paused, when developing this feature last year, so, uh, I was expecting this topic to be revived at some point. I do agree that we are relying on implicit (as it just works) behavior rather than something guaranteed, which would be nicer. I think attribute works well enough for this case.

> But the flip side of the "store qmemman_client reference in an attribute" is that, if something goes wrong and close() isn't called, garbage collector won't save you anymore. So, it wouldn't be "delayed for 20min" anymore, but qmemman would stop doing anything forever (which includes breaking starting further qubes). Or at least until responsible qube is removed (or maybe until qubesd or qmemman restart).

The connection can seize when the client sends empty reply: https://github.com/QubesOS/qubes-core-admin/blob/2488f2b19385b4c30dd1f18aecf0252a82732828/qubes/tools/qmemmand.py#L197, which is what happens when the task is cancelled / connection broken. So I think attribute can work and I also think that taking qmemman "hostage" till the qube is paused is okay as paused usually takes ten ms and I don't remember a pause failure at all, so qmemman is not indefinitely locked.

qubes/vm/dispvm.py
ben-grande: Testing this...

ben-grande:
Set qmemman log level to info.

#### Letting the preload be be paused (not requesting it before pause)

```
INFO: vm.disp3373: Preload startup completed 'qubes.WaitForRunningSystem'
INFO: vm.disp3373: Setting qube memory to pref mem
do_balloon_dom(dom_memset={'18': 0})
mem requested for dom '18' is 0, using its pref '398119321' while actual mem is '4194304000' (10.54x)
mem-set domain 18 to 398119321
...
mem-set domain 18 to 398119321
WARNING: vm.disp3373: Cancelling wait for preload
WARNING: vm.disp3373: Cancelled qmemman on asyncio.CancelledError
client disconnected, resuming membalance
balance_when_enough_mem(xen_free_mem=36188021536, total_mem_pref=3760358680, total_available_mem=45537320456)
stat: dom '0' act=4294967296 pref=2101949721 last_target=4294967296
stat: dom '5' act=4194304000 pref=475978547 last_target=4194304000
stat: dom '6' act=4194304000 pref=784311091 last_target=4194304000
stat: dom '18' act=426082304 pref=398119321 last_target=398119321
stat: xenfree=36240450336 memset_reqs=[('0', 4294967296), ('5', 4194304000), ('6', 4194304000), ('18', 4194304000)]
mem-set domain 0 to 4294967296
mem-set domain 5 to 4194304000
mem-set domain 6 to 4194304000
mem-set domain 18 to 4194304000
INFO: vm.disp3373: Preloading completed
INFO: vm.disp3373: Paused preloaded qube
```

- We can notice that it does try rebalance domain 18 up to 4194304000 before pause event notifies it has finished. Yes, this should be guarded against...
- Cancelling the break task does cancel qmemman

#### Requesting the preload mid domain-pre-pause, mid qmemman pref mem

```
INFO: vm.disp4946: Preload startup completed 'qubes.WaitForRunningSystem'
INFO: vm.disp4946: Setting qube memory to pref mem
do_balloon_dom(dom_memset={'19': 0})
mem requested for dom '19' is 0, using its pref '399929753' while actual mem is '4194304000' (10.49x)
mem-set domain 19 to 399929753
...
mem-set domain 19 to 399929753
INFO: vm.disp4946: Requesting preloaded qube
INFO: vm.default-dvm: Removing qube(s) from preloaded list because qube was requested: 'disp4946'
INFO: vm.disp4946: Waiting preload completion with '72' seconds timeout
WARNING: vm.disp4946: Cancelling qmemman
INFO: vm.disp4946: Preloading completed
INFO: vm.disp4946: Using preloaded qube
INFO: Updating appmenus for 'disp4946' in 'dom0'
INFO: vm.default-dvm: Received preload event 'used' for dispvm 'disp4946' with a delay of 3.0 second(s)
mem-set domain 19 to 399929753
...
mem-set domain 19 to 399929753
client disconnected, resuming membalance
```

It just has `WARNING: vm.disp4946: Cancelling qmemman`, and it doesn't seem to cancel qmeman at that point, just later on due to lack of data received.

qubes/vm/dispvm.py
marmarek: Well, not just `raise` , log, exception here seems to get lost.

marmarek:
> It just has `WARNING: vm.disp4946: Cancelling qmemman`, and it doesn't seem to cancel qmeman at that point, just later on due to lack of data received.

I don't think the current implementation of qmemman checks if client disconnected mid-action. Not ideal, but also not a huge problem.

ben-grande:
> > It just has `WARNING: vm.disp4946: Cancelling qmemman`, and it doesn't seem to cancel qmeman at that point, just later on due to lack of data received.
> 
> I don't think the current implementation of qmemman checks if client disconnected mid-action. Not ideal, but also not a huge problem.

I moved the instantiation of `QMemmanClient` to `qubes.vm.dispvm` to avoid it hanging from `QubesVM.set_mem`, and `self.sock.close()` does not forcefully close. I tried `self.sock.shutdown(...)` with `socket.SHUT_*` options but same result, doesn't forcefully close. Depending on the option, it may close the client early, but the server will still be calling mem-set.

I didn't find a nice solution, I saw that something can maybe be done with `select` or `settimeout`. `setblocking(0)` doesn't seem much useful alone, has to be used with other options.

> Looking at it closer, I think there is an issue with how this is called in the non-requested case. set_mem() will instruct qmemman to reduce memory of the qube. So far so good. But as soon as the client disconnects, qmemman will start balancing memory again, likely giving just reduced qube more memory again. IIUC that disconnect happens currently only implicitly, when garbage collector releases qmemman_client object, which is IMO quite fragile. Theoretically, the qmemman socket should be kept open until after the qube is paused. In practice, the qube is paused fast enough so it doesn't change much.
> 
> See how similar qmemman handling is done during qube startup: first qubesd request to free some memory for the new qube, then starts it and only then close the qmemman socket - otherwise qmemman would reuse that freed memory and there wouldn't be space for starting that new qube anymore.
> 
> But here, it isn't a single function, it would be two separate event handlers - `domain-pre-paused` and `domain-paused`. So, qmemman_client instance would need to be passed from one to the other somehow. Maybe on the VM object as some attribute? If done in `set_mem()` itself, it would allow fixing also the other issue, as you wouldn't need a result of a (cancelled) task anymore.

So it seems that the client cannot easily forcefully close the connection nor hold it before ballooning again.

I am testing using this script:

```python3

import asyncio
import qubes
import qubes.qmemman.client

async def main():
    app = qubes.Qubes()
    domains = app.domains
    disp = domains["disp3668"]
    break_task = asyncio.create_task(asyncio.sleep(0.0003))
    qmemman_client = qubes.qmemman.client.QMemmanClient()
    #qmemman_task = asyncio.get_running_loop().run_in_executor(None, disp.set_mem)
    qmemman_task = asyncio.get_running_loop().run_in_executor(None, qmemman_client.set_mem, {disp.xid: 0})
    tasks = [break_task, qmemman_task]
    try:
        async for earliest_task in asyncio.as_completed(tasks):
            await earliest_task
            if earliest_task == break_task:
                print("Finished break, cancelling qmemman")
                qmemman_task.cancel()
            else:
                print("Finished qmemman, cancelling break")
                break_task.cancel()
    except asyncio.CancelledError:
        pass
    finally:
        print("Final")
        if qmemman_client.sock:
            qmemman_client.shutdown()

    result = None
    cancelled = False
    try:
        #result = await qmemman_task
        result = qmemman_task.result()
    except asyncio.CancelledError:
        cancelled = True
    finally:
        #qmemman_client.close()
        if not result or cancelled:
            disp.log.warning("Failed to set memory")

if __name__ == "__main__":
    asyncio.run(main())
```

Where QMemmanClient.shutdown is shutdown+close().

marmarek:
> does not forcefully close

Well, closing the socket does _not_ interrupt the function on the other end of it. qmemman will notice only when it will interact with the socket next time. For interrupting the call, qmemman would need some monitoring of the socket state during `do_balloon_dom` call. It's not going to be pretty with the current shape of qmemman (might be easier if qmemman would use asyncio, but that's a big change beyond the scope of this PR). But I think you can ignore this problem and accept that interrupting `set_mem` isn't possible (maybe come back to this issue later). The critical part is to not leave the socket open after the operation completes, regardless of the exit path (so, either after dispvm gets requested, or after pausing it).

ben-grande:
> maybe come back to this issue later

Right, it ended up being bigger changes that what I expected and I got lost trying to fix it all at once. Will focus just on closing the socket properly. I cannot use `task.result()` because if the task was cancelled, it returns `CancelledError` instead of `QMemmanClient`. This is why I had to instantiate from the DispVM so I could get the class instance to close the socket.

ben-grande:
> For interrupting the call, qmemman would need some monitoring of the socket state during `do_balloon_dom` call. It's not going to be pretty with the current shape of qmemman (might be easier if qmemman would use asyncio, but that's a big change beyond the scope of this PR).

What about writing to a file or modify DomainState somehow, so that `do_balloon_dom` stops looping if a condition is met. This way, we can also prohibit balloon between domain-pre-paused and domain-paused, and allow balloon at the end of domain-paused.

ben-grande:
Haven't tested latest commit. For another day.

marmarek:
> What about writing to a file or modify DomainState somehow, so that `do_balloon_dom` stops looping if a condition is met. This way, we can also prohibit balloon between domain-pre-paused and domain-paused, and allow balloon at the end of domain-paused.

We used to do file-based trigger, it's messy. For example if two DispVMs would preload in parallel.

ben-grande:
> We used to do file-based trigger, it's messy. For example if two DispVMs would preload in parallel.

What about a file that contains the domid, to remove it from balancing/ballooning? Tested now and it's working. Currently using 2 files for different purposes:

- one for cancelling the current balloon in case the preload was requested
- one for preventing balance/balloon until the preload is requested, which means it will wait for pause

marmarek:
> What about a file

I just gave one example out of many what could go wrong... No, I don't want file based interface for controlling stuff. If you want more examples:
- you'd need some locking/synchronization to ensure data is not interpreted mid-write
- restarting (or crashing) a service at the wrong time will leave inconsistent state (possibly preventing further ballooning)

We've been there. I don't want to keep finding new failure modes for the next several months again...

ben-grande:
Dropped the file based condition. Will leave for another future future PR. Opened separate issues:

- https://github.com/QubesOS/qubes-issues/issues/10857
- https://github.com/QubesOS/qubes-issues/issues/10858

marmarek (approved):

PR #797, opened 2026-04-07
Add Qubes and 3mdeb master keys

### LINKED ISSUE #2544: Add Qubes Master Signing Key to dom0
As [John David R. Smith suggested](https://groups.google.com/d/msgid/qubes-users/a04c000f-b0c1-55e4-535f-50cc2e44b2ed%40openmailbox.org), the Qubes Master Signing Key should be included in dom0, to mitigate the risk of an MITM attack replacing the key. This would also provide users with an additional means of confirming that their copy of the key is genuine. (As a corollary, it's yet another place that an attacker would have to attempt to replace the key with a forgery in order to deceive users.)

I recall someone suggesting this a long time ago, and I (think I) also recall @marmarek doing it, but I can't find the original thread or issue, and I don't see the key in `/etc/pki/rpm-gpg/`.

### LINKED ISSUE #10663: Include the Dasharo master key somewhere in the distribution

There is likely a fairly wide overlap between users of Qubes OS and users of Dasharo.  In particular, my understanding is that Dasharo is (or at least was) used by all currently certified hardware.  However, there is no trivial way to obtain a chain of trust to the 3mdeb key used to sign Dasharo release files.

This is particularly important for those who purchased Qubes OS-certified NovaCustom laptops with versions of Dasharo not supported by fwupd, but it is more general.

### The solution you'd like

Include the Dasharo public key in the Qubes distribution (a trivial change).

I’m willing to send a patch, but its verification by the Qubes team and inclusion in a release would be the real value.

Users of Dasharo firmware with Qubes OS will have an easier time securely obtaining the needed tools to work with this firmware.

_No response_

DemiMarie:
These allow verifying software from within dom0 if necessary.

(I did NOT verify the 3mdeb key, but presumably @marmarek can.)

Fixes: QubesOS/qubes-issues#10663
Fixes: QubesOS/qubes-issues#2544


marmarek:
1. Please don't add more loose files in root repo dir.
2. I don't think core-admin is the right place. It's supposed to have code / scripts, not infra / meta files. Maybe do that in meta-packages repo? Ideally a new sub-package (`qubes-keys`? `qubes-distro-keys`?)

ben-grande:
Please reference the issues you are trying to solve:

- https://github.com/qubesos/qubes-issues/issues/10663
- https://github.com/qubesos/qubes-issues/issues/2544

And just if you want to add more keys:

- https://github.com/qubesos/qubes-issues/issues/4292

marmarek:
> 2\. I don't think core-admin is the right place.

Let me be more direct: not this repo.

PR #800, opened 2026-04-09
replace `random` with `secrets`

tabs-are-better-than-spaces:
This PR replaces the `random` module that uses an algorithm that is not a suitable CSPRNG (https://docs.python.org/3/library/random.html) with pythons `secrets` module.

The function `random_string` is called by the qubes admin api that generates a token https://github.com/QubesOS/qubes-core-admin/blob/2488f2b19385b4c30dd1f18aecf0252a82732828/qubes/api/admin.py#L470 where i am not sure if it is supposed to be secret.

The change in zfs.py does not seem to have any security impact, but to prevent calling it in the future to generate values that are supposed to be secret i changed it along.

### DIFF
--- a/qubes/storage/zfs.py
+++ b/qubes/storage/zfs.py
@@ -7,7 +7,7 @@ import contextlib
 import dataclasses
 import logging
 import os
-import random
+import secrets
 import shlex
 import shutil
 import string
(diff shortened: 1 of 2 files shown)


ben-grande:
The usage in the example you showed is indeed not something that has to be secret, it is more intended to be matched later. Anyway, I don't think it would hurt to improve it a bit, since there is this warning on the "random" page you shared:

> Warning
> 
> The pseudo-random generators of this module should not be used for security purposes. For security or cryptographic uses, see the [secrets](https://docs.python.org/3/library/secrets.html#module-secrets) module.

And we don't know if future use wouldn't fall into using the `random_string()` for secrets.

marmarek (approved):

PR #803, opened 2026-04-16
reject invalid port 65536

tabs-are-better-than-spaces:
the highest possible port should be 65535


marmarek (approved):
spaces are better than tabs! :)

PR #804, opened 2026-04-17
tests: Add an environment sync test to TC_20_NonAudio

ArrayBolt3:
Multiple bugs in Qubes have been the result of the either the user session or the systemd user manager missing environment variables that were set in /etc/profile.d (see Qubes issues #10299, #10712, #10759). We now have code that syncs the environment from the systemd user manager into a user session, and vice versa. Add tests that ensure environment variables are being synced both ways without clobbering values that are supposed to be augmented.

Note that this test will fail without https://github.com/QubesOS/qubes-gui-agent-linux/pull/257. Also, this is completely untested and unlinted, as I don't have a machine for running integration tests locally, and pylint is broken for Qubes repos on Fedora 43 (or at least was last I tried it). I didn't immediately see any code that deletes VMs that are created during the test process, so I've made no effort to clean up the StandaloneVMs this code creates, assuming it will be automatically cleaned up.

### DIFF
--- a/qubes/tests/integ/vm_qrexec_gui.py
+++ b/qubes/tests/integ/vm_qrexec_gui.py
@@ -32,6 +32,7 @@ import qubes.devices
 import qubes.tests
 import qubes.vm.appvm
 import qubes.vm.templatevm
+import qubes.vm.standalonevm
 
 in_qemu = os.path.exists("/sys/firmware/qemu_fw_cfg")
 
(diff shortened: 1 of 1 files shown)


qubes/tests/integ/vm_qrexec_gui.py
```diff
@@ -930,6 +931,100 @@ async def _test_400_long_window_title(self, utf8=False):
             except ProcessLookupError:  # already dead
                 pass
 
+    def test_500_gui_agent_env_sync(self):
+        # Create a StandaloneVM so changes made to it will survive a reboot
+        self.env_sync_vm = self.app.add_new_vm(
```
marmarek, line 936: Creating standalone out of template is a bit weird on core-admin side. See https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/tests/integ/basic.py#L565-L575

qubes/tests/integ/vm_qrexec_gui.py
```diff
@@ -930,6 +931,100 @@ async def _test_400_long_window_title(self, utf8=False):
...
+        self.assertEqual(self.env_sync_vm.get_power_state(), "Running")
+        self.loop.run_until_complete(self.wait_for_session(self.env_sync_vm))
+
+        try:
+            self.loop.run_until_complete(
+                self.env_sync_vm.run_for_stdio(
+                    "printf '%s\\n' "
+                    + "'#!/bin/bash' "
+                    + "'QUBES_ENV_TEST=first' "
+                    + "'export QUBES_ENV_TEST' "
```
marmarek, line 958: Env generator is supposed to print to stdout, not export.

qubes/tests/integ/vm_qrexec_gui.py
```diff
@@ -930,6 +931,100 @@ async def _test_400_long_window_title(self, utf8=False):
...
+        # equal to 'first:second', indicating that the environment has been
+        # synced from systemd into the session
+        self.loop.run_until_complete(self.env_sync_vm.start())
+        self.assertEqual(self.env_sync_vm.get_power_state(), "Running")
+        self.loop.run_until_complete(self.wait_for_session(self.env_sync_vm))
+
+        try:
+            self.loop.run_until_complete(
+                self.env_sync_vm.run_for_stdio(
+                    "test \"$(printf '%s\\n' \"$QUBES_ENV_TEST\")\" "
```
marmarek: This won't produce useful stdout/stderr for debugging. Maybe better retrieve variable value, and compare on the python side?

qubes/tests/integ/vm_qrexec_gui.py
```diff
@@ -930,6 +931,100 @@ async def _test_400_long_window_title(self, utf8=False):
...
+            self.fail(
+                "environment was not synced to session properly, expected "
+                + "value 'first:second', got value: {}".format(e.stderr)
+            )
+
+        ## Make sure the systemd user manager has the variable set properly too
+        try:
+            self.loop.run_until_complete(
+                self.env_sync_vm.run_for_stdio(
+                    "test \"$(systemctl --user show-environment "
```
marmarek: And same here.

marmarek (changes requested):
Besides few technical comments, there is one more important factor (especially for the shell variable case): whether the command runs as part of user (GUI) session, or not. You can force either mode via:
- to force running inside existing session - add wait_for_session call
- to force running outside of it - specify different user (for example root).

ArrayBolt3:
> to force running inside existing session - add wait_for_session call

This is what I want in all situations here, but I think I'm doing this already? I have `self.loop.run_until_complete(self.wait_for_session(self.env_sync_vm))` in a couple spots.

qubes/tests/integ/vm_qrexec_gui.py
ArrayBolt3: Fixed, thanks!

qubes/tests/integ/vm_qrexec_gui.py
ArrayBolt3: Fixed.

qubes/tests/integ/vm_qrexec_gui.py
ArrayBolt3: Good idea, done.

qubes/tests/integ/vm_qrexec_gui.py
ArrayBolt3: Done.

ArrayBolt3:
Alright, so we get binary content with a newline at the end, now I know. Will fix.

ben-grande:
> with a newline at the end

I'd prefer not inserting the newline in the printf format than stripping it later.

ArrayBolt3:
@ben-grande Done.

ben-grande:
Black complains, after fixing that, I will trigger: `openQArun LABEL=openqa-pending TEST=system_tests_basic_vm_qrexec_gui`.

ArrayBolt3:
Formatting hopefully fixed.

marmarek:
openQArun LABEL=openqa-pending TEST=system_tests_basic_vm_qrexec_gui MACHINE=64bit

marmarek:
openQArun PR_LABEL=openqa-pending TEST=system_tests_basic_vm_qrexec_gui MACHINE=64bit

marmarek (approved):

PR #807, opened 2026-05-06
Return appropriate exception when libvirt shutdown times out

ben-grande:
For: https://github.com/QubesOS/qubes-issues/issues/10835

### DIFF
--- a/qubes/app.py
+++ b/qubes/app.py
@@ -181,6 +181,9 @@ class VMMConnection:
         self._libvirt_conn = None
         self._xs = None
         self._xc = None
+        self._libvirt_conn_type = None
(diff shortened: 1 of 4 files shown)


marmarek:
> qubes/vm/qubesvm.py:1:0: C0302: Too many lines in module (3015/3000) (too-many-lines)

really? ...

It is long, but I don't think splitting it would make it any easier, so maybe disable it for this file.

As for the test failure, they aren't running with real libvirt, you'll likely need to replace the function with a mockup...

marmarek:
Black complains

marmarek (approved):

PR #812, opened 2026-05-27
Cache device information from hwdata

ben-grande:
For: https://github.com/QubesOS/qubes-issues/issues/10648

### DIFF
--- a/qubes/device_protocol.py
+++ b/qubes/device_protocol.py
@@ -837,8 +837,8 @@ class DeviceInterface:
             return "Microphone"
         return repr(self)
 
-    @staticmethod
-    def _load_classes(bus: str):
+    @classmethod
+    def _load_classes(cls, bus: str):
         """
         List of known device classes, subclasses and programming interfaces.
         """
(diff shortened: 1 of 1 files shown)


qubes/device_protocol.py
```diff
@@ -802,7 +802,8 @@ def __str__(self):
             return "Block Device"
         if self.devclass in ("usb", "pci"):
             # try subclass first as in `lspci`
-            result = self._load_classes(self.devclass).get(
+            # pylint: disable=protected-access
+            result = self.__class__._load_classes(self.devclass).get(
```
marmarek: Same comment as in the other PR. And then you can also drop pylint disable comment.

qubes/device_protocol.py
```diff
@@ -813,7 +814,8 @@ def __str__(self):
                 "vendor specific subclass",
             ):
                 # if not, try interface
-                result = self._load_classes(self.devclass).get(
+                # pylint: disable=protected-access
+                result = self.__class__._load_classes(self.devclass).get(
```
marmarek: and here

qubes/device_protocol.py
```diff
@@ -822,7 +824,8 @@ def __str__(self):
                 "undefined",
             ):
                 # if not, try class
-                result = self._load_classes(self.devclass).get(
+                # pylint: disable=protected-access
+                result = self.__class__._load_classes(self.devclass).get(
```
marmarek: and here

marmarek (changes requested):

qubes/device_protocol.py
marmarek: > And then you can also drop pylint disable comment.

marmarek (approved):

PR #813, opened 2026-05-28
Add support for default standalone boot modes

ArrayBolt3:
When a standalone is cloned from a template, the usual default boot mode of the standalone is the same as the boot mode the template originally used. In some instances this is undesirable; Kicksecure for instance puts its templates into a "sysmaint" boot mode that impairs standalone functionality.

To fix this, add a new qvm-feature, "boot-mode.standalone-default", which states which boot mode a standalone should be switched to when its root volume is cloned from any other VM. (Note that cloning a standalone from another standalone will switch it to the boot mode specified by "boot-mode.standalone-default", cloning a template isn't the only way to trigger this.)

Allows fixing https://github.com/QubesOS/qubes-issues/issues/10514, but does not fix this bug by itself.

I've tested the main code of this on my R4.3 machine, and it seems to work. The test code is untested though, and I'd be surprised if the integration tests are actually correct as-is. Note also that I *only* implemented integration tests for the actual "change boot mode on root volume clone" part of this, since I couldn't figure out how to sanely implement normal tests. Thus coverage.py will probably be unhappy with this.

### DIFF
--- a/qubes/ext/core_features.py
+++ b/qubes/ext/core_features.py
@@ -194,6 +194,16 @@ class CoreFeatures(qubes.ext.Extension):
             ) and hasattr(vm, "appvm_default_bootmode"):
                 bootmode_value = untrusted_feature_value
                 vm.features["boot-mode.appvm-default"] = bootmode_value
+        if "boot-mode.standalone-default" in untrusted_features:
+            untrusted_feature_value = untrusted_features[
+                "boot-mode.standalone-default"
+            ]
+            if (
+                f"boot-mode.kernelopts.{untrusted_feature_value}" in vm.features
+                or untrusted_feature_value == "default"
+            ):
+                bootmode_value = untrusted_feature_value
+                vm.features["boot-mode.standalone-default"] = bootmode_value
 
         # allow VMs to switch on anon-timezone, but do not permit dropping it
         if "anon-timezone" in untrusted_features:
--- a/qubes/tests/integ/storage.py
+++ b/qubes/tests/integ/storage.py
@@ -31,6 +31,7 @@ import qubes.tests.storage_reflink
 import qubes.tests.storage_zfs
 import qubes.utils
 import qubes.vm.appvm
+import qubes.vm.standalonevm
 
 
 class StorageTestMixin(object):
@@ -46,6 +47,16 @@ class StorageTestMixin(object):
             qubes.vm.appvm.AppVM, name=self.make_vm_name("vm2"), label="red"
         )
         self.loop.run_until_complete(self.vm2.create_on_disk())
+        self.standalone_vm1 = self.app.add_new_vm(
+            qubes.vm.standalonevm.StandaloneVM,
+            name=self.make_vm_name("standalone-vm1"),
+            label="red",
+        )
+        self.standalone_vm1.clone_properties(self.app.default_template)
+        self.standalone_vm1.features.update(self.app.default_template.features)
+        self.loop.run_until_complete(
+            self.standalone_vm1.clone_disk_files(self.app.default_template)
+        )
         self.pool = None
         self.init_pool()
         self.app.save()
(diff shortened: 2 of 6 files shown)


marmarek:
Tests are missing lines like this in expected calls: `"features.__contains__", ('boot-mode.kernelopts.new-mode', ), {}`

ArrayBolt3:
Fixed. (I noticed the tests failed earlier today and was planning on getting to this, but then forgot, so thank you for the ping :) )

ArrayBolt3:
Looking at the codecov output, it makes me realize the integration tests I created aren't really comprehensive (they make sure that a normal boot mode transition works, but don't test what happens if the transition isn't possible or what happens when trying to transition to the "default" mode). It's probably worth testing those edge cases too. I can add those tests once I have confirmation that the existing tests look right.

qubes/ext/core_features.py
```diff
@@ -194,6 +194,16 @@ async def qubes_features_request(self, vm, event, untrusted_features):
...
                 bootmode_value = untrusted_feature_value
                 vm.features["boot-mode.appvm-default"] = bootmode_value
+        if "boot-mode.standalone-default" in untrusted_features:
+            untrusted_feature_value = untrusted_features[
+                "boot-mode.standalone-default"
+            ]
+            if (
+                f"boot-mode.kernelopts.{untrusted_feature_value}" in vm.features
+                or untrusted_feature_value == "default"
+            ) and hasattr(vm, "standalone_default_bootmode"):
```
marmarek: This attribute is not defined anywhere, but also, it's unused (except this line).

qubes/tests/integ/storage.py
```diff
@@ -485,6 +500,25 @@ async def _test_006_no_revisions(self):
...
+    async def _test_007_standalone_clone_boot_mode_transition(self):
+        standalone_vm1.features["boot-mode.kernelopts.orig-mode"] = "test1"
+        standalone_vm1.features["boot-mode.kernelopts.new-mode"] = "test2"
+        standalone_vm1.features["boot-mode.active"] = "orig-mode"
+        standalone_vm1.features["boot-mode.standalone-default"] = "new-mode"
+        template_rootvol = self.template.storage.get_volume("root")
+        standalone_vm1_rootvol = self.standalone_vm1.storage.get_volume("root")
+        await qubes.utils.coro_maybe(
+            standalone_vm1_rootvol.import_volume(template_rootvol)
+        )
```
marmarek, line 517: That should use `standalone_vm1.storage.import_volume()`, calling import_volume on the volume object itself would skip firing the event...

marmarek (changes requested):

qubes/ext/core_features.py
ArrayBolt3: That was a leftover of an earlier implementation attempt, will remove.

qubes/tests/integ/storage.py
ArrayBolt3: Good to know, will fix.

Edit: Looked at the code closer to see what I missed. I see it now, it should have been obvious, not sure why I missed this...

qubes/tests/integ/storage.py
```diff
@@ -46,6 +47,20 @@ def setUp(self):
             qubes.vm.appvm.AppVM, name=self.make_vm_name("vm2"), label="red"
         )
         self.loop.run_until_complete(self.vm2.create_on_disk())
+        self.standalone_vm1 = self.app.add_new_vm(
+            qubes.vm.standalonevm.StandaloneVM,
+            name=self.make_vm_name("standalone-vm1"),
+            label="red",
+        )
+        self.standalone_vm1.clone_properties(self.app.domains[self.template])
```
marmarek: This isn't test for each template, just use `self.app.default_template` (and in the few places below)

ArrayBolt3: Fixed.

marmarek:
> * StorageFile: [test_007_standalone_clone_boot_mode_transition](https://openqa.qubes-os.org/tests/185986#step/StorageFile/8) (error)
>       `NameError: name 'standalone_vm1' is not defined`

:(

ArrayBolt3:
Should be fixed now.

marmarek (approved):

PR #816, opened 2026-05-30
tests: add missing test cleanup

marmarek:
skip_kernel_validation_patch was not stopped, and keeps reference to
some objects.


marmarek:
PipelineRetry

PR #817, opened 2026-06-01
tests: do not fail dispvm+thunderbird test early if TB is not installed

marmarek:
Checking if Thunderbird is installed is done later. Do not fail on
preparing its configuration - make it work also if /etc/thunderbird/pref
isn't pre-existing.

If TB is not installed, the test will be recorded as skipped anyway (so
the file won't be used), but create the dir anyway. This will allow
working with distributions not pre-creating that directory (if any).


marmarek:
PipelineRetry

PR #818, opened 2026-06-01
Skip running check for untrusted_qdb access

ben-grande:
One access doesn't takes few nanoseconds, but when looping through domains multiple times, this blocking operation can impact the responsiveness of qubesd.

For: https://github.com/QubesOS/qubes-issues/issues/10648

### DIFF
--- a/qubes/app.py
+++ b/qubes/app.py
@@ -1233,6 +1233,15 @@ class Qubes(qubes.PropertyHolder):
 
         for vm in self.domains:
             vm.events_enabled = True
+            if (
+                not self.vmm.offline_mode
+                and isinstance(vm, qubes.vm.qubesvm.QubesVM)
+                and not vm.untrusted_qdb
+                and vm.is_running()
+            ):
+                import qubesdb  # pylint: disable=import-error
+
+                vm._qdb_connection = qubesdb.QubesDB(vm.name)
             vm.fire_event("domain-load")
 
         # get a file timestamp (before closing it - still holding the lock!),
--- a/qubes/ext/audio.py
+++ b/qubes/ext/audio.py
@@ -36,13 +36,13 @@ class AUDIO(qubes.ext.Extension):
     @staticmethod
     def set_qubesdb_audiovm(vm):
         # Ensure that qube is ready
-        if not vm.is_running():
+        if not getattr(vm, "untrusted_qdb"):
             return
 
         # Add AudioVM Xen ID for gui-agent
         audiovm = getattr(vm, "audiovm", None)
         if audiovm is not None:
-            if audiovm != vm and audiovm.is_running():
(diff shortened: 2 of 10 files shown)


qubes/app.py
```diff
@@ -1233,6 +1233,15 @@ def load(self, lock=False):
...
             vm.events_enabled = True
+            if (
+                not self.vmm.offline_mode
+                and isinstance(vm, qubes.vm.qubesvm.QubesVM)
+                and not vm.untrusted_qdb
+                and vm.is_running()
+            ):
+                import qubesdb  # pylint: disable=import-error
+
+                vm._qdb_connection = qubesdb.QubesDB(vm.name)
```
ben-grande, line 1244: The reason I had to move to `qubes.app` is because firing the event "domain-load" reaches extensions before it reaches `qubes.vm.qubesvm`...

marmarek:
This PR likely completely breaks qubesdb access for freshly started VMs (as in - started when qubesd is already running).
Why exactly you do that? Generally, qdb property shouldn't be accessed if VM isn't running, so the case of this hitting `not vm.is_running()` (so, not opening the conneciton) should be rather rare.
And the just-pushed force-push is even worse - checking for `untrusted_qdb` property as a proxy for checking if VM is running is a very obscure solution.

Is it all to avoid calling `vm.is_running()`? Why? Did you found it slow? If so, the proper (the only acceptable) solution would be to optimize it, not avoid it. Maybe cache that info in the VM object, and update it based on libvirt events?

ben-grande:
> This PR likely completely breaks qubesdb access for freshly started VMs (as in - started when qubesd is already running). Why exactly you do that? Generally, qdb property shouldn't be accessed if VM isn't running, so the case of this hitting `not vm.is_running()` (so, not opening the conneciton) should be rather rare. And the just-pushed force-push is even worse - checking for `untrusted_qdb` property as a proxy for checking if VM is running is a very obscure solution.
> 
> Is it all to avoid calling `vm.is_running()`? Why? Did you found it slow?

Yes. It is slow. One call is fine, but many blocking calls takes a considerable time considering it can be called thousand of times (3k-5k just from the two recorded API calls) to check that a halted domain, that has no untrusted_qdb, is not running.

I find it horrible to call `libvirt_domain.isActive()` multiple times for domains that is not running, and that we would need to write to their qubesdb, which is only available when the domain is running anyway.

> If so, the proper (the only acceptable) solution would be to optimize it, not avoid it. Maybe cache that info in the VM object, and update it based on libvirt events?

I guess that is possible... I though it would be dangerous to cache such state, but I can try and see how it goes.

ben-grande:
Superseded by https://github.com/QubesOS/qubes-core-admin/pull/819

PR #819, opened 2026-06-01
Cache libvirt domain state

ben-grande:
The state changes rarely, but querying it can take a considerable time, that is blocking, when looping through the state of multiple domains.

---

Didn't run openqa locally, only experiment some things:

- start
- shutdown
- kill
- pause
- unpause
- checking Domains Widget
- restart qubesd
- `qubed-query -e dom0 admin.vm.CurrentState QUBE`

---

TODO:

- [x] test with https://github.com/QubesOS/qubes-core-admin/pull/833


ben-grande:
PipelineRetryFailed

ben-grande:
Didn't run openqa locally, only experiment some things, start, shutdown, kill, checking Domains Widget.

ben-grande:
PipelineRetryFailed

ben-grande:
PipelineRetryFailed

ben-grande:
PipelineRetryFailed

ben-grande:
Marked as draft as there still is some caching issues for the `_power_state`.

marmarek:
If necessary (or simpler), it's IMO okay to use cache only for `is_running()` (that is used frequently), but still do active check for `get_power_state()`. It could be also a safety valve for cases when cache are not reliable (like during startup or shutdown).

ben-grande:
> If necessary (or simpler), it's IMO okay to use cache only for `is_running()` (that is used frequently), but still do active check for `get_power_state()`. It could be also a safety valve for cases when cache are not reliable (like during startup or shutdown).

Experimenting with cache a bit more and it's getting better, but if I don't get it to a hundred percent, will not cache the power state.

ben-grande:
From my tests, it seems to be working. My nemesis, OpenQA, should try me. Do you know a subset of tests that are interesting before joining this PR in the full run?

Looking at this [recent test](https://openqa.qubes-os.org/tests/overview?distri=qubesos&version=4.3&build=2026053102-devel&groupid=11):

- system_tests_suspend (soft-failed)
- system_tests_usbproxy
- system_tests_devices 	
- system_tests_gui_tools
- system_tests_guivm_gui_interactive

---

Line to use after this PR is not draft anymore:
```
openQArun TEST=system_tests_suspend,system_tests_usbproxy,system_tests_devices,system_tests_gui_tools,system_tests_guivm_gui_interactive
```

marmarek:
I'd include also at least one of the guivm tests

ben-grande:
PipelineRetryFailed

ben-grande:
Last CI failed on Fedora 43 to due to:

```
ERROR: qubes.tests.api_admin/TC_00_VMs/test_150_pool_info
  File "/usr/lib64/python3.14/contextlib.py", line 85, in inner
    return func(*args, **kwds)
  File "/home/gitlab-runner/builds/QubesOS/qubes-core-admin/qubes/tests/api_admin.py", line 76, in setUp
    super().setUp()
    ~~~~~~~~~~~~~^^
  File "/home/gitlab-runner/builds/QubesOS/qubes-core-admin/qubes/tests/__init__.py", line 516, in setUp
    self.loop = asyncio.get_event_loop()
                ~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib64/python3.14/asyncio/events.py", line 715, in get_event_loop
    raise RuntimeError('There is no current event loop in thread %r.'
                       % threading.current_thread().name)
RuntimeError: There is no current event loop in thread 'MainThread'.
```

So it finally reached this repo.

ben-grande:
PipelineRetryFailed

ben-grande:
openQArun TEST=system_tests_suspend,system_tests_usbproxy,system_tests_devices,system_tests_gui_tools,system_tests_guivm_gui_interactive

ben-grande:
> * system_tests_gui_tools
>   
>   * qui_widgets_update: [unnamed test](https://openqa.qubes-os.org/tests/182585#step/qui_widgets_update/18) (unknown)
>   * qui_widgets_update: [Failed](https://openqa.qubes-os.org/tests/182585#step/qui_widgets_update/19) (test died)
>     `# Test died: no candidate needle with tag(s) 'qubes-update-finish' ...`
>   * qui_widgets_update: [unnamed test](https://openqa.qubes-os.org/tests/182585#step/qui_widgets_update/21) (unknown)

Seeing "Next and previous results", shows that this happened 3 days ago also, so unrelated to this PR: https://openqa.qubes-os.org/tests/182011#step/qui_widgets_update/18.

> * system_tests_guivm_gui_interactive
>   
>   * update_guivm: [wait_serial](https://openqa.qubes-os.org/tests/182587#step/update_guivm/23) (unknown)
>     `# Command: (set -o pipefail; qubesctl --all --show-output state.hig...`
>   * update_guivm: [Failed](https://openqa.qubes-os.org/tests/182587#step/update_guivm/24) (test died + timed out)
>     `# Test died: command '(set -o pipefail; qubesctl --all --show-outpu...`
>   * update_guivm: [wait_serial](https://openqa.qubes-os.org/tests/182587#step/update_guivm/100) (unknown)
>     `# Command: curl --form upload=@/var/log/libvirt/libxl/libxl-driver....`

Failed recently also, and `qubesctl-sys-gui.log` was not uploaded. Video shows it froze at that stage.

> * system_tests_usbproxy
>   
>   * TC_20_USBProxy_core3_debian-13-xfce: [test_090_attach_stubdom](https://openqa.qubes-os.org/tests/182588#step/TC_20_USBProxy_core3_debian-13-xfce/14) (error)
>     `qubes.exc.QubesVMError: Cannot connect to qrexec agent for 120 seco...`

Doesn't seem related to this PR.

---

I will restart the failed tests.

marmarek:
PipelineRetry

qubes/storage/__init__.py
```diff
@@ -657,24 +657,28 @@ def attach(self, volume, rw=False):
...
     def _is_already_attached(self, volume):
         """Checks if the given volume is already attached"""
+        # TODO: ben: cache
         parsed_xml = lxml.etree.fromstring(self.vm.libvirt_domain.XMLDesc())
-        disk_sources = parsed_xml.xpath("//domain/devices/disk/source")
-        for source in disk_sources:
-            if source.get("dev") == "/dev/%s" % volume.vid:
-                return True
+        disks = parsed_xml.xpath(
+            "//domain/devices/disk/source[@dev='/dev/{}']".format(volume.vid)
```
DemiMarie: Instead of injecting the variable with `.format()`, use XPath variables.  Something like this (not tested):

```python
        disks = parsed_xml.xpath(
            "//domain/devices/disk/source[@dev = $dev]",
            dev = "/dev/" + volume.vid)
```

This avoids an XPath injection vulnerability if `volume.vid` winds up containing an apostraphe.  You can use a regular expression to validate `volume.vid`, but even then it is much better to use XPath variables.

`.find()` is even faster if it’s an option.

qubes/storage/__init__.py
```diff
@@ -657,24 +657,28 @@ def attach(self, volume, rw=False):
...
-        for disk in disks:
-            source = disk.xpath("source")[0]
-            if source.get("dev") == "/dev/%s" % volume.vid:
-                disk_xml = lxml.etree.tostring(disk, encoding="utf-8")
-                self.vm.libvirt_domain.detachDevice(disk_xml)
-                return
-        raise StoragePoolException("Volume {!r} is not attached".format(volume))
+        disks = parsed_xml.xpath(
+            "//domain/devices/disk/source[@dev='/dev/{}']".format(volume.vid)
+        )
```
DemiMarie: Use an XPath variable.

DemiMarie (commented):
XPath variables are super nice.  They are [the XPath equivalent of SQL parameterized queries](https://lxml.de/xpathxslt.html).

qubes/vm/qubesvm.py
```diff
@@ -1707,7 +1826,11 @@ async def shutdown(self, force=False, wait=False, timeout=None):
                     await asyncio.wait_for(waiter, timeout=timeout)
                 except asyncio.TimeoutError:
                     raise qubes.exc.QubesVMShutdownTimeoutError(self)
+
+            self._is_running = False
+            self._power_state = "Halted"
```
marmarek: At this point, it should be `Halted` already, set by libvirt event. So, maybe better add an assert about that and see if it actually works?

qubes/vm/qubesvm.py
```diff
@@ -1729,14 +1852,21 @@ async def kill(self):
...
+                self._is_running = False
+                self._power_state = "Halted"
                 raise qubes.exc.QubesVMNotStartedError(self)
+            self._is_running = None
+            self._power_state = None
             raise
 
         await waiter
+        self._is_running = False
+        self._power_state = "Halted"
```
marmarek: Similar here.

qubes/ext/pci.py
```diff
@@ -341,35 +354,39 @@ def _load_desc(self) -> Dict[str, str]:
...
+        return self._all_attached.get(self.port_id, None)
 
 
 class PCIDeviceExtension(qubes.ext.Extension):
     def __init__(self):
         super().__init__()
         # lazy load this
         self.pci_classes = {}
+        self._all_devices = []
+        self._attached = {}
```
marmarek: This cache will fail by not noticing assign/unassign when VM is not running. Specifically:
1. Assign a device to a VM (or for example take sys-usb with USB controller assigned)
2. Start it, and do `qvm-pci ls` to cache attached devices.
3. Shutdown it
4. Unassign device
5. Start it again

Now the cache has outdated info. The same probably works the other way (with assign in step 4).

qubes/ext/block.py
```diff
@@ -313,6 +305,11 @@ def _try_get_block_device_info(app, disk):
 
 class BlockDeviceExtension(qubes.ext.Extension):
 
+    def __init__(self):
+        super().__init__()
+        self._attachments = {}
```
marmarek: Similar comment to PCI - check if the cache is handled properly across restarting a VM with "required" assignments.

marmarek: And also check if non-required attachments are invalidated on shutdown.

qubes/vm/qubesvm.py
```diff
@@ -1544,6 +1573,10 @@ async def start(
                 )
                 self.skip_unpause_event = True
                 self.libvirt_domain.resume()
+                if self.is_fully_usable():
+                    self._power_state = "Running"
+                else:
+                    self._power_state = "Transient"
                 await self.fire_event_async("domain-unpaused")
```
marmarek: The `domain-unpaused` event is problematic for the client: here it really means it's still "Transient" state (until `domain-start` event), but in other cases it means transition to `Running` state. Client has no way of knowing and currently at least domains widget misinterprets it. Maybe it should get an argument saying what is the new state?
I proposed removing this event yesterday, and might be also a good idea to try, but check if this doesn't break anything (like preloaded disposables).

marmarek: And just to note - the `Transient` state is cursed. In many cases there is no event when going in/out of that state (for example if qrexec-agent is started/stopped in the qube)...

marmarek (changes requested):
Partial review

ben-grande:
I had to move the devices handling to a separate PR to unblock this one. See linked PR above. I will address the review made by Demi and Marek later, I saved the links of the the review to get back to it.

qubes/vm/qubesvm.py
ben-grande: > I proposed removing this event yesterday, and might be also a good idea to try, but check if this doesn't break anything (like preloaded disposables).

The only thing I've found this removal to crash is "Qube Manager", because I replaced `Transient` with `Starting` and `Halting`, and "Qube Manager" tried to get an icon for "Starting", which doesn't exist. Do you prefer that we have a transition period, where the server sends `Transient`  instead of `Starting`? (`Halting` was already handled, even though it was never sent...).

I made a PR to the qubes-manager to avoid breaking on new states in the future.

marmarek: > Do you prefer that we have a transition period, where the server sends `Transient` instead of `Starting`?

Yes, especially because there may be some 3rd-party apps with similar assumptions.

ben-grande: What is a proper way to announce that it will be deprecated in a future release? Just release notes? Sending deprecating warning is too much noise?

marmarek: Yes, release notes. And add an issue to https://github.com/orgs/QubesOS/projects/32 for the list of breaking changes.

ben-grande:
PipelineRetryFailed

marmarek:
Some tests are failing here

qubes/vm/qubesvm.py
```diff
@@ -1757,9 +1757,8 @@ async def shutdown(self, force=False, wait=False, timeout=None):
             if self.is_paused() and not force and not is_preload:
                 raise qubes.exc.QubesVMNotRunningError(self)
 
-            if self.__waiter is None:
-                self.__waiter = asyncio.get_running_loop().create_future()
-            waiter = self.__waiter
+            if self._power_state_waiter is None:
+                self._power_state_waiter = asyncio.get_running_loop().create_future()
```
marmarek: The local variable (`waiter`) was important part. Using attribute directly does not guarantee you wait for the same operation you initiated here - it might be something else that happened later (like, somebody calling kill just after shutdown, or maybe start just after kill).

This is the commit that introduced the issue, but the same issue applies to later `change_libvirt_state` context manager.

qubes/vm/qubesvm.py
```diff
@@ -1663,6 +1657,35 @@ async def start(
...
+    ):
+        """
+        Asynchronously wait for libvirt method to complete.
+        """
+        self.init_lifecycle_waiter()
+        try:
+            yield
+        finally:
+            if wait:
+                await asyncio.wait_for(self._lifecycle_waiter, timeout=timeout)
```
marmarek: Using the same event object for different events feels risky. What happens for example if kill() is called while start() is running? It might return early for example when `VIR_DOMAIN_EVENT_DEFINED` or `VIR_DOMAIN_EVENT_RESUMED` is delivered.
This would be even more complicated in connection with #867 that relies more on waiting for an operation started by somebody else (like `cancel_start()` does). But `shutdown()` already has similar case.

qubes/vm/qubesvm.py
```diff
@@ -1633,19 +1626,14 @@ async def start(
...
                     "domain-start", start_guid=start_guid
                 )
 
             except Exception as exc:  # pylint: disable=bare-except
-                self.log.error("Start failed: %s", str(exc))
-                # let anyone receiving domain-pre-start know that startup failed
-                await self.fire_event_async(
-                    "domain-start-failed", reason=str(exc)
-                )
+                await self.notify_failed_startup(exc=exc)
```
marmarek, line 1655: This will set `self._power_state` too early, no?

marmarek (changes requested):

qubes/vm/qubesvm.py
ben-grande: > The local variable (`waiter`) was important part. Using attribute directly does not guarantee you wait for the same operation you initiated here - it might be something else that happened later (like, somebody calling kill just after shutdown, or maybe start just after kill).

Hm.. right. I think I will change for action specific waiters, so a waiter for shutdown, a wait for suspend etc.

A use case for attribute instead of local variable is https://github.com/QubesOS/qubes-core-admin/pull/876/changes#r3914042842, as I think an attribute might prevent the shutdown from being called twice when there is an awaitable before the libvirt shutdown.

I think the same thing that is being done for `shutdown` on the other PR, should be done to the other power state actions also, although I understand that the shutdown is more critical.

qubes/vm/qubesvm.py
ben-grande: Yes, it's wrong, will fix it.

qubes/vm/qubesvm.py
ben-grande: Yes, seems too early.

qubes/vm/qubesvm.py
```diff
@@ -1597,21 +1664,150 @@ async def start(
...
+        if self._lifecycle_waiter[event] is None:
+            self._lifecycle_waiter[event] = (
+                asyncio.get_running_loop().create_future()
+            )
+        try:
+            yield
+        finally:
+            if wait:
+                await asyncio.wait_for(
+                    self._lifecycle_waiter[event], timeout=timeout
```
marmarek: As said in https://github.com/QubesOS/qubes-core-admin/pull/819#discussion_r3874590758, this should use waiter saved in a local variable, before `yield`. Otherwise you may hit `None` here if the event is delivered early (or in a very unlucky situation, even wait for another instance of that event, make it wait too long or even forever).

marmarek (changes requested):

qubes/vm/qubesvm.py
ben-grande: Now I understand what you mean by that, thanks for the explanation.

marmarek:
openQArun TEST=system_tests_suspend,system_tests_usbproxy,system_tests_devices,system_tests_gui_tools,system_tests_guivm_gui_interactive,system_tests_basic_vm_qrexec_gui

marmarek:
openQArun TEST=system_tests_suspend,system_tests_usbproxy,system_tests_devices,system_tests_gui_tools,system_tests_guivm_gui_interactive,system_tests_basic_vm_qrexec_gui,system_tests_dispvm

marmarek (approved):

PR #820, opened 2026-06-02
Short loop.run_in_executor with asyncio.to_thread

ben-grande:
Doesn't require:

- finding the loop as it uses the default;
- passing "None" as executor as it uses the default;
- using partial to pass kwargs

---

~Didn't test at all~, but would be nice to use a shorter version, since we are not using the full capabilities of the executor, a wrapper is easier to read. If you look at `asyncio/threads.py`, it's just calling `loop.run_in_executor`.

Edit: tested manually the preloaded disposable early request before qmemman finishes and watching the logs. Didn't run integration tests, and that would be appropriate for the different files I changed.

### DIFF
--- a/qubes/storage/reflink.py
+++ b/qubes/storage/reflink.py
@@ -60,9 +60,7 @@ def _coroutinized(function):
 
     @functools.wraps(function)
     async def wrapper(*args, **kwargs):
-        return await asyncio.get_event_loop().run_in_executor(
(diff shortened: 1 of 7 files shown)


ben-grande:
PipelineRetryFailed

marmarek:
PipelineRetry

marmarek (approved):

PR #822, opened 2026-06-03
Remove StandardOutput setting from systemd units

ben-grande:
When validating the units, I see this message:

        Standard output type syslog is obsolete, automatically updating
        to journal. Please update your unit file, and consider removing
        the setting altogether

---

Didn't test. Just doing as the message says.


marmarek:
PipelineRetry

marmarek (approved):

PR #825, opened 2026-06-09
api/internal: invalidate SystemInfoCache on domain-add and domain-delete

### LINKED ISSUE #10911: SystemInfoCache.on_domain_add does not invalidate the cache -> freshly-created VMs are invisible to qrexec policy until the next invalidating event
### Qubes OS release

Qubes OS 4.3 (tested on 4.3.0; `qubes-core-dom0-4.3.45-1.fc41`, `qubes-core-qrexec-4.3.12-1.fc41`).

### Brief summary

`qubes/api/internal.py:SystemInfoCache` caches the `system_info` dict that qubes-qrexec-policy-daemon queries via `internal.GetSystemInfo`. The cache is invalidated on the per-VM events listed in `vm_events` (via `event_handler`, which does `cls.cache = None`), but `on_domain_add` does **not** invalidate it — it only registers per-VM handlers on the new VM. `vm_events` contains neither `domain-add` nor `domain-delete`, and `on_domain_delete` has the same omission.

As a result, a freshly-created VM can be absent from the cached `system_info` until some other event happens to invalidate the cache. While the cache is stale, qubes-qrexec-policy-daemon logs `target 'X' does not exist, using @default instead` for the new VM and falls back to whatever rule matches `@default`.

`on_domain_add` does register the per-VM invalidation handlers (including `domain-tag-add`) on the new VM, but it does so after the VM is constructed and added, and does not invalidate the cache for the new VM's own current state. So the new VM stays absent from the cached snapshot until some later event invalidates it. The bug is most easily observed for `qvm-create --class DispVM` disposables whose tags are inherited from the DVMT at creation (no post-create tag-add fires — re-adding an already-present tag is a no-op), but it is architectural: any freshly-created VM that does not immediately trigger an invalidating event is affected for the duration of that window.

### Steps to reproduce

The bug is directly visible in the source and needs no run to confirm: `on_domain_add` (see Additional information) does not invalidate the cache. The steps below make it observable at runtime. They reproduce most cleanly **on a quiet system right after a qubesd restart** — background VM/tag activity can invalidate the cache between create and query and mask the symptom (this is why it is intermittent without instrumentation).

1. (Optional — for live confirmation only; the code in Additional information is sufficient on its own.) Add four logging-only `logging.warning` calls to `SystemInfoCache` in `/usr/lib/python3.*/site-packages/qubes/api/internal.py`:
   - in `event_handler`, before `cls.cache = None` → `INVALIDATE event=<event>`;
   - in `on_domain_add`, before `cls.register_events_vm(vm)` → `DOMAIN-ADD vm=<vm.name> cache_state=<None or count=len(cls.cache['domains'])>`;
   - in `get_system_info` at the cache-hit return (`if cls.cache is not None: return cls.cache`) → `CACHE-HIT count=<len(cls.cache['domains'])>`;
   - in `get_system_info` just before `cls.cache = system_info` → `CACHE-REBUILD count=<len(system_info['domains'])>`.
   Then `sudo systemctl restart qubesd qubes-qrexec-policy-daemon`.
2. Create a DVMT carrying a tag: `qvm-create --class AppVM --template <template> --label red dvmt-probe`; `qvm-prefs dvmt-probe template_for_dispvms True`; `qvm-tags dvmt-probe add repro-tag`.
3. Install a probe policy in `/etc/qubes/policy.d/29-repro.policy`: `admin.vm.property.Get * <source-vm> @tag:repro-tag allow target=@adminvm`, then `sudo systemctl restart qubes-qrexec-policy-daemon`.
4. Loop 5 times. Each iteration: `qvm-create --class DispVM --template dvmt-probe --label red repro-N`; `qvm-tags repro-N add repro-tag` (no-op, already inherited); `sleep 1`; restart `qubes-qrexec-policy-daemon`; from `<source-vm>` run `qrexec-client-vm repro-N admin.vm.property.Get+name`; then inspect the qubesd cache log and the policy-daemon journal. (The per-iteration policy-daemon restart is not the trigger — the staleness is a qubesd-side cache property; the restart only forces a fresh `internal.GetSystemInfo` query.)

### Expected behavior

After a VM is created, the next `internal.GetSystemInfo` query (e.g. from qubes-qrexec-policy-daemon) should include the new VM, so a qrexec call naming it resolves to that VM and is matched against the intended policy rule.

### Actual behavior

Iteration 1 masks the bug (the cache is `None` at qubesd start, so the first query rebuilds and includes the new VM). Iterations 2-5 fail: the call is refused (`rc=126`), and the qubesd cache log shows the cache already populated when `on_domain_add` runs, with no subsequent invalidation before the next cache hit returns the stale snapshot. Verbatim from an instrumented run (qubes-core-dom0-4.3.45; the `Q5DBG` prefix is the added logging; the new VM is `...-2`):

```
qubesd: Q5DBG CACHE-REBUILD count=39 names=[...'...-1']
qubesd: Q5DBG DOMAIN-ADD vm='...-2' cache_state=count=39
qubesd: Q5DBG CACHE-HIT count=39
qrexec-policy-daemon: qrexec: target '...-2' does not exist, using @default instead
qrexec-policy-daemon: qrexec: admin.vm.property.Get+name: <source> -> ...-2: denied: denied by policy ...
```

The load-bearing observables are: `DOMAIN-ADD … cache_state=count=N` (the cache is already populated when `on_domain_add` runs) → **no** `INVALIDATE` line → `CACHE-HIT count=N` (the same stale snapshot is served, still missing the new VM) → the policy daemon's `target '…' does not exist, using @default instead`. That last `does not exist, using @default` line is the universal symptom; the `denied by policy …` line that follows it depends on local policy (on this box it was the project's own deny rule, hence the path shown above). Throughout, the new VM exists in qubesd's view — `qvm-check` succeeds and its `klass`/tags read correctly via the Admin API; only the cached `system_info` snapshot is stale. With the logging in place, the loop reproduced this 4/5 times consecutively.

### Additional information

**Root cause** — `qubes/api/internal.py` (confirmed on `release4.3`). `on_domain_add` registers handlers but never invalidates the cache:

```python
@classmethod
def on_domain_add(cls, subject, event, vm):
    cls.register_events_vm(vm)
```

compare `event_handler`, which does invalidate for the `vm_events`:

```python
@classmethod
def event_handler(cls, subject, event, **kwargs):
    """Invalidate cache on specific events"""
    cls.cache = None
```

`vm_events` contains neither `domain-add` nor `domain-delete`, and `on_domain_delete` likewise only calls `unregister_events_vm` without invalidating.

**Proposed fix** — one line (plus the symmetric `on_domain_delete`):

```python
@classmethod
def on_domain_add(cls, subject, event, vm):
    cls.cache = None
    cls.register_events_vm(vm)
```

`domain-add`/`domain-delete` are app-level events; invalidating there is the symmetric counterpart to how every `vm_events` entry already invalidates. Performance impact is negligible — these events are infrequent relative to the per-VM events that already clear the cache.

**Reproduction rate** — 4/5 with the logging patch; intermittent without it, because background events can invalidate the cache between create and query. The mechanism is deterministic; the patch only makes the existing behaviour observable.

**Security note (low impact)** — while the cache is stale, qrexec target resolution falls back to `@default`. Stock policy ships `qubes.Notifications * @anyvm @default allow target=dom0`, so a `qubes.Notifications` call naming a not-yet-cached VM is routed to dom0 instead of the intended target. I found no admin/sensitive `@default`-routed service that would enable privilege escalation; flagging it as fail-permissive behaviour, not as an exploit.

**Origin** — surfaced while building tag-scoped Admin API policies for an MCP-server isolation project (qubes-mcp), and discussed on qubes-devel (thread 4NuSqL64DVE). `SystemInfoCache` was introduced in qubes-core-admin#651.

alex-schose:
`SystemInfoCache` invalidates `cls.cache` for every event in `vm_events`
(via `event_handler`), but `on_domain_add` and `on_domain_delete` only
register / unregister the per-VM handlers - they never invalidate the cache
itself, and `vm_events` contains neither `domain-add` nor `domain-delete`.

As a result a freshly-created (or freshly-deleted) VM is missing from / stale
in the cached `system_info` until some other event happens to invalidate it.
While stale, qubes-qrexec-policy-daemon's `internal.GetSystemInfo` query
returns the old snapshot: a call naming a new VM logs
`target '...' does not exist, using @default instead`, and a deleted VM can
still be offered by a `@default` `ask` handler (e.g. qvm-copy).

Fix: invalidate the cache in both handlers, after (un)registering the per-VM
handlers. `domain-add` / `domain-delete` are app-level events; this is the
symmetric counterpart to how every `vm_events` entry already invalidates via
`event_handler`. Performance impact is negligible - these events are
infrequent relative to the per-VM events that already clear the cache.

Fixes QubesOS/qubes-issues#10911.

### DIFF
--- a/qubes/api/internal.py
+++ b/qubes/api/internal.py
@@ -76,11 +76,13 @@ class SystemInfoCache:
     def on_domain_add(cls, subject, event, vm):
         # pylint: disable=unused-argument
         cls.register_events_vm(vm)
+        cls.cache = None
 
     @classmethod
     def on_domain_delete(cls, subject, event, vm):
         # pylint: disable=unused-argument
         cls.unregister_events_vm(vm)
+        cls.cache = None
 
     @classmethod
     def register_events_vm(cls, vm):


marmarek:
PipelineRefresh

marmarek:
PipelineRetry

andrewdavidwong:
> Fixes #10911.

You must add `QubesOS/qubes-issues` in front of `#10911` because it's a different repo. The whole thing should look like this: `Fixes QubesOS/qubes-issues#10911`.

alex-schose:
Done — updated to `Fixes QubesOS/qubes-issues#10911`.

ben-grande:
To avoid rebuilding the cache on very frequent disposable creation and deletion, can you try just appending and removing the relevant domain to the cache? You will have to split the cache building.

marmarek:
> To avoid rebuilding the cache on very frequent disposable creation and deletion, can you try just appending and removing the relevant domain to the cache? You will have to split the cache building.

I don't think it's worth it at this point. _IF_ rebuilding cache will become too expensive at some point, then maybe, but lets keep it simple for now.

alex-schose:
Agreed, keeping it simple. For the record — measured on a 37-qube R4.3 box (dom0 4.3.45): a full `get_system_info` rebuild is ≈21 ms / ~107 libvirt round-trips; on top of #819 it's ≈4.5 ms / 2 calls. Against ≈0.5 cache-invalidating events/s under serial dispvm churn that's not a hotspot, and #819 removes most of the cost anyway — worth revisiting only if a preload-saturated workload pushes the rate much higher.

marmarek (approved):

ben-grande:
> a full `get_system_info` rebuild is ≈21 ms / ~107 libvirt round-trips; on top of #819 it's ≈4.5 ms / 2 calls

Cool. Can you share how many qubes you have on your system so I can get an idea of how much time it would take on a system with hundreds of qubes?

alex-schose:
37 here, ~10 running (R4.3, dom0 4.3.45). Rather than guess the scaling I measured it — spun up running disposables and watched `get_system_info`. The rebuild grows **+8 libvirt round-trips per running qube** (same counter as the 107 above), dead linear from R=9 to R=19; a halted qube adds ~1, a paused one 2. So it tracks the *running* count, not the total: a running qube does isActive + state + the usability/qrexec checks, a halted one returns at isActive.

Extrapolating that slope: 300 qubes with 50 running ≈ 650 round-trips (~130 ms/rebuild); 100 running ≈ 1000 (~200 ms) — all blocking the qubesd loop. (Counting the ID()/is_qrexec_running() calls too it's +10/running.)

Which is the term #819 takes out: with power state served from the cache the per-domain round-trips go to zero, so the rebuild stops scaling with running-count. Happy to share the probe, or run a #819-applied flat-curve if it'd help the PR.

ben-grande:
> Happy to share the probe, or run a #819-applied flat-curve if it'd help the PR.

And I'd be happy to see the results and more of the methodology.

alex-schose:
Ran the #819-applied flat-curve on the same box — same probe, same counter, same R sweep as the numbers above; the only thing that changes is the imported `qubes` tree (deployed `main` vs a PYTHONPATH overlay of #819 at `fd6ef64`, against the live libvirt in dom0 — nothing installed, qubesd untouched).

Steady-state (`_power_state` warm, as a running qubesd keeps it) the per-domain round-trips go to **zero** — the rebuild stops scaling with running-count:

| R running (incl dom0) | N | main rt_v1 | main rt_all | #819 rt_v1 | #819 rt_all |
|---|---|---|---|---|---|
| 9  | 37 | 93  | 109 | 1 | 1 |
| 12 | 40 | 117 | 139 | 1 | 1 |
| 15 | 43 | 141 | 169 | 1 | 1 |
| 19 | 47 | 173 | 209 | 1 | 1 |

main slope **+8 rt_v1 / +10 rt_all per running qube**; #819 is **flat — a constant 1** (a single amortized `lookupByUUID`; zero `isActive`/`state`/`is_fully_usable`/`ID`/`is_qrexec_running`). Round-trip counts were identical across all 6 iterations at every plateau; wall-clock I'd ignore (too noisy on a live box). The short-circuit is in `set_power_state` (`qubesvm.py:2564` at `fd6ef64`): once `_power_state` is populated it returns before any of those calls fire, and `get_system_info` calls `get_power_state()` with no args.

One caveat so the picture's complete: on a **cold** `_power_state` — the first rebuild after a qubesd restart, before any lifecycle hook has run — #819 populates the cache once. That populate scales with running-count, but it's **one-time and already cheaper than a single `main` rebuild**: 34 calls at R9, 74 at R19 — versus main's 109 / 209 on *every* rebuild. After it, flat. So #819 turns the per-rebuild cost from O(running) into O(1) libvirt IPC, and the win grows with how many qubes are running — the few-hundred-qube case you asked about.

Methodology: each point is the median of 6 cold-`SystemInfoCache` rebuilds, with `_power_state` pre-warmed once per domain first (what a live qubesd looks like — lifecycle hooks keep it hot). Round-trips are counted by wrapping the entry points `get_system_info` touches; `rt_v1` is the four-call subset comparable to the 107 I cited earlier, `rt_all` the full count. The measurement core is just:

```python
import gc, time, statistics, libvirt, qubes, qubes.vm.qubesvm
from qubes.api.internal import SystemInfoCache

counts = {}
def wrap(obj, name, key):                       # count every call to `key`
    orig = getattr(obj, name); counts[key] = 0
    def w(self, *a, **k):
        counts[key] += 1
        return orig(self, *a, **k)
    setattr(obj, name, w)

wrap(libvirt.virConnect, "lookupByUUID", "lookupByUUID")
wrap(libvirt.virDomain,  "isActive",     "isActive")
wrap(libvirt.virDomain,  "state",        "state")
wrap(libvirt.virDomain,  "ID",           "ID")
wrap(qubes.vm.qubesvm.QubesVM, "is_fully_usable",   "is_fully_usable")
wrap(qubes.vm.qubesvm.QubesVM, "is_qrexec_running", "is_qrexec_running")
V1 = ("lookupByUUID", "isActive", "state", "is_fully_usable")  # the subset == the 107

app = qubes.Qubes()
for d in app.domains:                           # pre-warm _power_state (steady-state qubesd)
    try: d.get_power_state()
    except Exception: pass

rows = []
for _ in range(6):                              # median of 6 cold-cache rebuilds
    SystemInfoCache.cache = SystemInfoCache.cache_for_app = None
    gc.collect()
    before = dict(counts)
    t0 = time.perf_counter_ns()
    SystemInfoCache.get_system_info(app)
    ms = (time.perf_counter_ns() - t0) / 1e6
    delta = {k: counts[k] - before[k] for k in counts}
    rows.append((ms, sum(delta.values()), sum(delta[k] for k in V1)))

ms, rt_all, rt_v1 = (statistics.median(x) for x in zip(*rows))
print(f"N={sum(1 for _ in app.domains)}  rt_v1={rt_v1}  rt_all={rt_all}  ms={ms:.1f}")
```

The #819 measurement is the same script run with `PYTHONPATH` pointed at a `pull/819/head` checkout, against the live libvirt in dom0. Sweep was just starting minimal disposables in plateaus and re-running it; happy to share the lifecycle/teardown harness if useful.

PR #826, opened 2026-06-11
Inform clients about suspend

ben-grande:
Since there is "resume", and there is a moment when resuming that domain might still be suspended, or event suspending a domain on purposed, allow clients to know the domain state.

For: https://github.com/QubesOS/qubes-core-admin-client/pull/333


ben-grande:
Superseded by https://github.com/QubesOS/qubes-core-admin/pull/827

PR #830, opened 2026-06-16
Disable notification for denied property.GetAll call

marmarek:
It's expected to be denied when some properties are allowed but not all.

QubesOS/qubes-issues#10534

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,6 +195,8 @@ endif
 	mkdir -p $(DESTDIR)/usr/libexec/qubes
 	install -m 0644 qubes-rpc-policy/90-default.policy \
 		$(DESTDIR)/etc/qubes/policy.d/90-default.policy
+	install -m 0644 qubes-rpc-policy/91-admin-default-deny.policy \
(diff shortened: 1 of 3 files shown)


adromanov:
I can imagine scenario when rouge VM tries to call `GetAll` to abuse misconfigured policy. Currently there would be a notification, but after this change there wouldn't.
@marmarek could you please explain is this a dangerous situation? Are properties sensible information?

marmarek:
It disables notification for a _denied_ call, it doesn't allow anything new.
As for sensitive info - not really, besides seeing what you have configured. Generally, audiovm gets access to only info it really needs, no need to share more with it.

adromanov:
Yeah, I understand that nothing is being allowed by this change, just notifications being made silent.
What I meant is that if currently some VM tries to call `GetAll` hoping that user accidentally configured policies wrong - this suspicious activity would be immediately noticed. But after the change it would not, VMs could freely query `GetAll` because why not, maybe they will be lucky.
But if properties are not that sensitive then that does not really matter. Thanks for the answer!

PR #835, opened 2026-07-03
Allow all storage drivers to shrink volumes, like file-reflink

rustybird:
Similar in purpose to #477, #478, #479 by @a-barinov in 2022, but with the shrinking check happening on the frontend side as [suggested](https://github.com/QubesOS/qubes-core-admin/pull/477#issuecomment-1138114094) by @marmarek. The frontend already had such a check but only in the `qvm-volume` tool; QubesOS/qubes-core-admin-client#475 centralizes this check to `qubesadmin.storage.Volume.resize()` so that any other callers are protected as well. (Qube Manager's qube Settings dialog / `qubes-vm-settings` also has its own logic that avoids calling `resize()` if it would shrink.)

Closes #479

### DIFF
--- a/qubes/storage/file.py
+++ b/qubes/storage/file.py
@@ -362,14 +362,6 @@ class FileVolume(qubes.storage.Volume):
             )
             raise qubes.exc.StoragePoolException(msg)
 
-        if size < self.size:
-            raise qubes.exc.StoragePoolException(
-                "For your own safety, shrinking of %s is"
-                " disabled. If you really know what you"
-                " are doing, use `truncate` on %s manually."
-                % (self.name, self.vid)
-            )
-
         with open(self.path, "a+b") as fd:
             fd.truncate(size)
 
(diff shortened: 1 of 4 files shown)


rustybird:
@Rudd-O: Are you sure that OpenZFS's `zfs set volsize=...` - which the 'zfs' driver uses to resize volumes, right? - does not support shrinking? Here's someone in the OpenZFS forum (upvoted by two people) wishing that it *didn't* shrink without prompting for confirmation: https://github.com/openzfs/zfs/discussions/14820. And some people in the FreeNAS forum also claiming that it works: https://www.truenas.com/community/threads/changing-size-shrink-on-active-zvol.42142/#post-444130

Maybe it fails if snapshots of the zvol exist? If OpenZFS does support shrinking even then, I propose to add this commit to the PR: https://github.com/rustybird/qubes-core-admin/commit/shrink-volumes-zfs

rustybird:
After setting up ZFS, I tested shrinking a 'zfs' driver volume and it worked fine. So I've added that commit to the PR.

marmarek (approved):

PR #840, opened 2026-07-14
device busy attribute to prevent attaching devices whose children are in use

### LINKED ISSUE #9566: The hidden block devices due to attachment are not restored after detachment

### Qubes OS release

R4.2.3

### Brief summary

When block device e.g. /dev/sda has multiple partitions and you attach a single partition e.g. /dev/sda1 to another qube, then block device /dev/sda is removed from the list of block devices available for attachment.
And the same is when whole block device e.g. /dev/sda is being attached to another qube, then all partitions of this block device e.g. /dev/sda1 are removed from the list of block devices available for attachment.
I guess this is to prevent the attachment of the same device to multiple qubes.
The problems that after you detach the attached block device, the block devices removed to prevent the attachment of the same device to multiple qubes won't be available for attachment anymore.
If you run `partprobe` in the source qube to refresh the block device info after detachment, then it'll restore the missing block devices in the list of block devices available for attachment.

Maybe related:
https://github.com/QubesOS/qubes-issues/issues/4692

### Steps to reproduce

In disposable qube disp1234 run these commands:
```
dd if=/dev/zero of=/tmp/tst.img bs=1M count=128
dev=$(sudo losetup -f --show /tmp/tst.img)
sudo sgdisk -n 0:0:+64MiB $dev
sudo sgdisk -n 0:0: $dev
sudo partprobe $dev
```
Check the available block devices in dom0:
```
[user@dom0 ~]$ qvm-block
BACKEND:DEVID     DESCRIPTION                 USED BY
disp1234:loop0    /tmp/tst.img                
disp1234:loop0p1   ()                         
disp1234:loop0p2   ()                         
```
Attach a single partition of the loop device as block device in dom0 to another disposable qube:
```
qvm-block attach disp4321 disp1234:loop0p1
```
Check the available block devices in dom0:
```
[user@dom0 ~]$ qvm-block
BACKEND:DEVID     DESCRIPTION                 USED BY
disp1234:loop0p1   ()                         disp4321 (read-only=no, frontend-dev=xvdi)
disp1234:loop0p2   ()                         
```
Detach the block device:
```
qvm-block detach disp4321 disp1234:loop0p1
```
Check the available block devices in dom0:

### Expected behavior

```
[user@dom0 ~]$ qvm-block
BACKEND:DEVID     DESCRIPTION                 USED BY
disp1234:loop0    /tmp/tst.img                
disp1234:loop0p1   ()                         
disp1234:loop0p2   ()                         
```

### Actual behavior

```
[user@dom0 ~]$ qvm-block
BACKEND:DEVID     DESCRIPTION                 USED BY
disp1234:loop0p1   ()                         
disp1234:loop0p2   ()                         
```

### LINKED ISSUE #10180: Strange behavior of Qubes Devices widget when attaching and detaching block devices
### Qubes OS release

Qubes OS 4.3

### Brief summary

When attaching and detaching a block device to/from a qube repeatedly, strange behavior occurs. The first time a device is attached it will succeed. And the first time it is detached, it will succeed again. But the second time this device is to be attached to a qube, it fails with an error: "Error Attaching device Block_Storage: sub-device of sys-usb:sda RESCUE1201 to personal failed. Error: QubesDaemonAccessError - Got empty response from qubesd. See journalctl in dom0 for details." Checking journalctl in dom0 shows some warning lines like "Aug 24 03:38:39 dom0 qubesd[2390]: WARNING: protocol error for call b'admin.vm.device.block.Attach'+b'sys-usb+sda1:0930:6545:50E549C20268BFC0198E69DF:u080650:1.0:1' (b'dom0' → b'personal') with payload of 158 bytes". In addition to this, attaching and detaching a device like /dev/sda1 will cause /dev/sda to vanish. The error after reattaching /dev/sda1 only seems to appear in Qubes 4.3-rc1, but doesn't appear in Qubes 4.2. The bug about vanishing devices is present in 4.2 as well.

### Steps to reproduce

1. Boot to Q4.3-RC1 desktop
2. Attach a formatted (has a partition and a filesystem on it) usb flash drive to a free usb port
3. Make sure the "personal" qube is up and running
4. Click on the Qubes Devices widget tray icon to bring up the menu with a list of devices
5. You will see the new device listed i.e. as /dev/sda and partitions as /dev/sda1, /dev/sda2, etc
6. Attach the device /dev/sda1 to qube personal - attach will succeed
7. Detach the device from qube personal - detach will succeed
8. Reattach the same device again to qube personal - attach will fail any subsequent time

### Expected behavior

It should be possible to attach and detach a block device to a qube unlimited times.

### Actual behavior

A block device like /dev/sda1 can be only attached and detached once, then an error occurs upon subsequent attempts to attach the same device.

### Additional information

There are seemingly two bugs with the device widget - the first one has to do with the failed reattachment of block devices (described in this bug report), and the other is with the disappearing devices. The latter is also present in 4.2, and there's already an open bug report filed for it; might want to update it that it also affects 4.3.

https://github.com/QubesOS/qubes-issues/issues/9566

I'm also unsure whether the first bug is limited only to block devices or it also affects other devices since I've only tested it with block devices.

### LINKED ISSUE #10828: Re-attaching USB partition (block device) shows error in dom0's journal
### Qubes OS release

4.3.0

### Brief summary

Detaching and attaching again the same block storage (of USB device) shows errors in dom0 journal.

### Steps to reproduce

Connect a USB device that has a partition with a file system.

```
user@dom0:~ > qvm-device block list
BACKEND:DEVID  DESCRIPTION                                               USED BY
dom0:dm-5      Block_Storage: hosted by dom0 qubes dom0-swap             
sys-usb:sda    Block_Storage: sub-device of sys-usb:3-2 USB Flash Drive  
sys-usb:sda1   Block_Storage: sub-device of sys-usb:sda LEXAR
```

In another console:

```
root@dom0:~ # journalctl -o short-monotonic -f
```

```
user@dom0:~ > qvm-device block detach ddd4385 sys-usb:sda1
user@dom0:~ > qvm-device block attach ddd4385 sys-usb:sda1
user@dom0:~ > echo $?
0
```

### Expected behavior

No errors.
Or, in case of error, have correct error exit code.

### Actual behavior

Error in journal shows on second `attach`:

```
[ 6282.631637] dom0 widget-wrapper[7700]: Failed to handle event: ddd4385, device-attach:block, {'device': sys-usb+sda1:0000:0000::?******:1, 'options': "{'read-only': 'no', 'frontend-dev': 'xvdi'}"}
[ 6282.631637] dom0 widget-wrapper[7700]: Traceback (most recent call last):
[ 6282.631637] dom0 widget-wrapper[7700]:   File "/usr/lib/python3.13/site-packages/qubesadmin/events/__init__.py", line 307, in handle
[ 6282.631637] dom0 widget-wrapper[7700]:     handler(subject, event, **kwargs)
[ 6282.631637] dom0 widget-wrapper[7700]:     ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
[ 6282.631637] dom0 widget-wrapper[7700]:   File "/usr/lib/python3.13/site-packages/qui/devices/device_widget.py", line 606, in device_attached
[ 6282.631637] dom0 widget-wrapper[7700]:     self.devices[dev_id] = backend.Device(device, self)
[ 6282.631637] dom0 widget-wrapper[7700]:                            ~~~~~~~~~~~~~~^^^^^^^^^^^^^^
[ 6282.631637] dom0 widget-wrapper[7700]:   File "/usr/lib/python3.13/site-packages/qui/devices/backend.py", line 160, in __init__
[ 6282.631637] dom0 widget-wrapper[7700]:     self._interfaces = dev.interfaces
[ 6282.631637] dom0 widget-wrapper[7700]:                        ^^^^^^^^^^^^^^
[ 6282.631637] dom0 widget-wrapper[7700]: AttributeError: 'VirtualDevice' object has no attribute 'interfaces'
```

and the exit code is 0, as if no error occurred.

### Additional information

The attached device works as expected in both cases (1st and 2nd `attach`).

piotrbartman:
Add a `busy` concept to `DeviceInfo`, `BlockDevice`: `True` when the device itself or any of its children (partition, sub-device) is in use (attached, or used locally inside the backend, e.g. mounted, dm slave, swap etc.). Busy devices get refused at attach time and skipped by auto-attach.

- busy bool serialized/deserialized in the admin protocol as `busy=True (absent = free); 
- `BlockDevice.busy` lazily reads/caches /qubes-block-devices/<id>/busy from the backend's QDB
- on detach: the device's own busy marker is cleared synchronously; ancestors are refreshed calling `qubes.RefreshBlockDevices`, (see QubesOS/qubes-linux-utils/pull/144) rather than blindly clearing as dom0 can't see local mounts inside the backend, so a naive clear could free a USB parent while a sibling partition is still mounted there. This change is in the second commit and might be opted out after discussion.

Open items:

- `_refresh_busy_after_detach` does `await asyncio.sleep(1) ` before calling the backend (I'm not happy here, >.<), waiting for vbd teardown to actually disappear from sysfs.
- If the backend vm doesn't have `qubes.RefreshBlockDevices` the code falls back to an attachment-based unmark which ignore internal usage. IMHO these are limited consequences.
- The mechanism is TOCTOU-prone, but it narrows the window for accidental attach conflicts, it's not a security boundary against a compromised backend, which can misreport any QDB value anyway.
- local mount/umount inside the backend generates no block uevent, so both setting and clearing the busy marker for locally-used devices is "opportunistic", i.e.,  happens whenever the next unrelated udev event fires, not immediately.
- `_mark_parents_busy`, `_unmark_parents_busy` swallow all exceptions silently (`# best effort`)

fix: https://github.com/QubesOS/qubes-issues/issues/9566
fix: https://github.com/QubesOS/qubes-issues/issues/10180
should fix: https://github.com/QubesOS/qubes-issues/issues/10828

### DIFF
--- a/qubes/ext/block.py
+++ b/qubes/ext/block.py
@@ -49,6 +49,10 @@ SYSTEM_DISKS = ("xvda", "xvdb", "xvdc")
 SYSTEM_DISKS_DOM0_KERNEL = SYSTEM_DISKS + ("xvdd",)
 
 
+def get_busy_qdb_key(port_id):
+    return f"/qubes-block-devices/{port_id}/busy"
+
+
 class BlockDevice(qubes.device_protocol.DeviceInfo):
 
     def __init__(self, port: qubes.device_protocol.Port):
@@ -195,7 +199,10 @@ class BlockDevice(qubes.device_protocol.DeviceInfo):
             parent_ident, sep, interface_num = self._sanitize(
                 untrusted_parent_info
             ).partition(":")
-            devclass = "usb" if sep == ":" else "block"
+            if sep == ":" or re.match(r"^\d+-", parent_ident):
+                devclass = "usb"
+            else:
+                devclass = "block"
             if not parent_ident:
                 return None
             try:
@@ -242,6 +249,24 @@ class BlockDevice(qubes.device_protocol.DeviceInfo):
                 return True
         return False
 
+    @property
+    def busy(self) -> bool:
+        """
+        Is this device busy?  A device is considered busy if any of its children
+        are in use (e.g., a partition of a disk is attached).
+        """
+        if self._busy is None:
+            if not self.backend_domain or not self.backend_domain.is_running():
+                # don't cache this value
+                return False
+            untrusted_busy = self.backend_domain.untrusted_qdb.read(
+                get_busy_qdb_key(self.port_id)
+            )
+            self._busy = qubes.device_protocol.qbool_untrusted_busy(
+                untrusted_busy, self.backend_domain.log
+            )
+        return self._busy
+
     @property  # type: ignore[misc]
     def device_id(self) -> str:
         """
@@ -284,6 +309,18 @@ class BlockDevice(qubes.device_protocol.DeviceInfo):
         return "".join(
             c if c in set(safe_chars) else "_" for c in untrusted_device_desc
         )
+
+    def mark_busy(self, busy: bool) -> None:
+        """Set or clear the busy marker in the backend VM's QDB."""
+        key = get_busy_qdb_key(self.port_id)
+        if not busy:
+            try:
+                self.backend_domain.untrusted_qdb.rm(key)
+            except Exception:  # pylint: disable=broad-except
+                pass
+        else:
+            self.backend_domain.untrusted_qdb.write(key, b"True")
+        self._busy = busy
 
 
 def _try_get_block_device_info(app, disk):
@@ -330,6 +367,19 @@ class BlockDeviceExtension(qubes.ext.Extension):
                 for dev in self.on_device_list_block(vm, None)
             )
             self.devices_cache[vm.name] = current_devices
+            # Re-establish busy markers for partitions already attached
+            # before this qubesd instance started (e.g. after qubesd restart).
+            for port_id, front_vm in device_attachments.items():
+                if front_vm is not None:
+                    self._mark_parents_busy(
+                        BlockDevice(
+                            Port(
+                                backend_domain=vm,
+                                port_id=port_id,
+                                devclass="block",
+                            )
+                        )
+                    )
         else:
             self.devices_cache[vm.name] = {}
 
@@ -383,7 +433,7 @@ class BlockDeviceExtension(qubes.ext.Extension):
         if not untrusted_qubes_device_attrs:
             return None
         return BlockDevice(
-            Port(backend_domain=vm, port_id=port_id, devclass="block")
+            Port(backend_domain=vm, port_id=port_id, devclass="block"),
         )
 
     @qubes.ext.handler("device-list:block")
@@ -520,6 +570,12 @@ class BlockDeviceExtension(qubes.ext.Extension):
             )
             raise qubes.exc.UnrecognizedDevice()
 
+        if device.busy:
+            raise qubes.exc.QubesValueError(
+                f"Device {device} is busy: "
+                "a child partition is currently passed through to a VM"
+            )
+
         # validate options
         for option, value in options.items():
             if option == "frontend-dev":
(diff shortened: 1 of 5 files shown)


marmarek:
> * local mount/umount inside the backend generates no block uevent

Systemd's mount unit appears for mounted filesystems, so maybe there is a way to get events about it via dbus? Or maybe there is some systemd hook for this?

But generally, there should be clear role split who is responsible for setting/clearing the busy flag. It should be either the backend (in which case qubesd should not change it without coordination), or qubesd (in which case, the backend shouldn't have influence over it - so, store outside of qubesdb). The mixed model is asking for troubles, when both sides have different idea about the current state.

marmarek:
> system_tests_devices
> 
>     * TC_00_List_debian-13-xfce: [test_021_list_loop_partition_mounted](https://openqa.qubes-os.org/tests/189005#step/TC_00_List_debian-13-xfce/8) (failure)
>       `AssertionError: Device test-inst-vm:loop0::0 (/tmp/test.img) should...`

This tests needs an update, to look at the busy attribute now.

piotrbartman:
> But generally, there should be clear role split who is responsible for setting/clearing the busy flag. It should be either the backend (in which case qubesd should not change it without coordination), or qubesd (in which case, the backend shouldn't have influence over it - so, store outside of qubesdb). The mixed model is asking for troubles, when both sides have different idea about the current state.

I need to think about it

marmarek:
openQArun PR_LABEL=openqa-group-4

qubes/devices.py
```diff
@@ -455,11 +561,127 @@ class DeviceManager(dict):
...
         self._vm = vm
+        # guards the recursion in `busy_ports()`
+        self._computing_busy_ports = False
 
     def __missing__(self, key):
         self[key] = DeviceCollection(self._vm, key)
         return self[key]
 
+    @property
+    def usage_tracking(self) -> bool:
```
marmarek, line 572: Would it make sense to make this property per device class, not per VM? That may complicate some calculations, OTOH I can imagine separate devclasses may support it or not (while currently both usb and block are advertised by qubes-utils package, some other dev classes may be updated separately). Or not even about updating packages, webcam device (video companion) don't have usage reporting at all (and not sure if it even makes sense there).
Or maybe what you really need here is information if VM reports usage for _any_ dev class? Not sure if safe...

qubes/ext/block.py
```diff
@@ -65,6 +67,10 @@ def __init__(self, port: qubes.device_protocol.Port):
         self._mode: Optional[str] = None
         self._size: Optional[int] = None
         self._interface_num: Optional[str] = None
+        # optimization: when listing, we calculate it once and pass it here to
+        #  avoid lazy evaluation later for each device.
+        #  bus = exported OR used
```
marmarek, line 72:
```suggestion
        #  busy = exported OR used
```

qubes/ext/block.py
```diff
@@ -510,17 +566,121 @@ def on_device_pre_attached_block(self, vm, event, device, options):
...
+        `check_local_usage=False` ignores the last one for backward
+        compatibility.
+        """
+        if current_attachment is None:
+            sub_attach = attachments.attached_subdevice(device)
+            if sub_attach is not None:
+                subdevice, frontend = sub_attach
+                raise qubes.exc.DeviceUsed(
+                    f"Device {device} cannot be attached: it's subdevice "
+                    f"{subdevice} is attached to {frontend}."
```
marmarek, line 591: This potentially leaks qube name to a management vm that may not be allowed to see full system state. Better skip frontend name. The client may still (try to) check itself where that device is attached to (if it has access).

qubes/ext/block.py
```diff
@@ -510,17 +566,121 @@ def on_device_pre_attached_block(self, vm, event, device, options):
...
+        if not current_attachment:
+            return
+
+        if current_attachment == vm:
+            raise qubes.exc.DeviceAlreadyAttached(
+                f"Device {device} is already attached to this VM ({vm})."
+            )
+
+        raise qubes.exc.DeviceAlreadyAttached(
+            f"Device {device} already attached to {current_attachment}."
```
marmarek, line 636: Similar to the other comment, this may leak info the client shouldn't have access to.

qubes/tests/devices_block.py
```diff
@@ -1349,3 +1374,328 @@ def test_091_on_domain_shutdown_backend(self):
...
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(self.ext.attach_and_notify(front, assignment))
+        self.assertTrue(front.libvirt_domain.attachDevice.called)
+        front.fire_event_async.assert_called_once_with(
+            "device-attach:block", device=exp_dev, options=assignment.options
+        )
+
+    def test_110_busy_marker_missing_while_tracking(self):
+        # no marker if tracking => busy
+        # garbage marker if tracking => busy
```
marmarek, line 1488: Is this a todo comment? This isn't what this test does...

qubes/ext/block.py
```diff
@@ -243,6 +252,46 @@ def _is_attached_to(self, vm):
                 return True
         return False
 
+    @property
+    def busy(self) -> bool:
+        """
+        Is this device busy?  A device is considered busy if it or any of
+        its children is in use: attached to a VM or used locally in the
+        backend VM (mounted or part of a device-mapper).
```
marmarek, line 260: Shouldn't this cover also ancestors? For example, if the whole disk is attached, partition shouldn't be possible to attach. Currently it's handled by `refuse_unavailable()`, but wouldn't it be more consistent by handling it via `busy` property too? Then, the frontend would see that property too.
In practice it probably would be about changing `busy_ports()`.

marmarek (changes requested):
There are at least two significant refactors of this feature within the PR, I'm not sure if keeping this commit history helps. Maybe better squash at least some of the commits together, possibly with combining their commit messages to explain why it's done this way?

PR #845, opened 2026-07-17
Allow shutdown to kill preloads without force

### LINKED ISSUE #10993: `QubesVM.shutdown` of preloaded disposables requires `force=True`
### Qubes OS release

Qubes OS 4.3

### Brief summary

It is not possible to shutdown a paused domain. You have to kill it. `QubesVM.shutdown` allows killing paused domains if `force=True`: https://github.com/QubesOS/qubes-core-admin/blob/f5e7f8b7f4f4c1cbc20561eecf1caeaf6f1633b3/qubes/vm/qubesvm.py#L1677.

Destroying/Killing a domains might be harmful, as per `help(qube.libvirt_domain.destroy)`: _which may produce undesirable results, for example unflushed disk cache in the guest_.

### Steps to reproduce

1. `qvm-shutdown --all --wait --exclude sys-net`

### Expected behavior

Preloaded disposable are killed.

### Actual behavior

The client doesn't send force despite `--all` because there is `--exclude`, so VMM sends shutdown signal to preloaded disposables, which obviously fail. The conditional force enabling on the client is fine, let's allow the server to decide what to do with preloaded disposables.

### Additional information

Preloaded disposables are paused domains that holds no important data. There is no problem in killing them.

nopg186:
Fixes: https://github.com/QubesOS/qubes-issues/issues/10993

### DIFF
--- a/qubes/vm/qubesvm.py
+++ b/qubes/vm/qubesvm.py
@@ -1667,7 +1667,8 @@ class QubesVM(qubes.vm.mix.net.NetVMMixin, qubes.vm.LocalVM):
                 "domain-pre-shutdown", pre_event=True, force=force
             )
 
-            if self.is_paused() and not force:
+            is_preload = getattr(self, "is_preload", False)
+            if self.is_paused() and not force and not is_preload:
                 raise qubes.exc.QubesVMNotRunningError(self)
 
             if self.__waiter is None:


qubes/vm/qubesvm.py
```diff
@@ -1667,7 +1667,9 @@ async def shutdown(self, force=False, wait=False, timeout=None):
                 "domain-pre-shutdown", pre_event=True, force=force
             )
 
-            if self.is_paused() and not force:
+            is_preload = getattr(self, "is_preload", False)
+
```
ben-grande: Please remove this blank line.

ben-grande:
Title of PR and commit: `Allow shutdown to kill preloads without force`

Body of PR and commit: `Fixes: https://github.com/QubesOS/qubes-issues/issues/10993`.

marmarek (approved):

PR #851, opened 2026-07-22
Call `stop_encrypted()` even for nominally unencrypted volumes

rustybird:
See commit messages

### DIFF
--- a/qubes/storage/__init__.py
+++ b/qubes/storage/__init__.py
@@ -857,16 +857,12 @@ class Storage:
                 pass
 
     async def stop(self):
-        """Execute the stop method on each volume"""
+        """Stop each volume"""
         await qubes.utils.void_coros_maybe(
-            # pylint: disable=line-too-long
-            (
-                vol.stop_encrypted(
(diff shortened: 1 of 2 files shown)


qubes/storage/__init__.py
```diff
@@ -857,19 +857,14 @@ async def start(self):
...
-        await qubes.utils.void_coros_maybe(
-            # pylint: disable=line-too-long
-            (
-                vol.stop_encrypted(
-                    vol.encrypted_volume_path(self.vm.name, name)
-                )
-                if vol.ephemeral
-                else vol.stop()
+        """Stop each volume."""
+        for name, vol in self.vm.volumes.items():
```
marmarek: This changed structure stops volumes sequentially now, instead of concurrently as before.

marmarek (changes requested):

qubes/storage/__init__.py
rustybird: Oops, fixed.

rustybird:
Force-pushed again to remove a stray dot from the docstring diff

marmarek (approved):

PR #854, opened 2026-07-24
Format files with black 26.5.1

ben-grande:
Cause it is asking to reformat on my recent pushes: https://gitlab.com/QubesOS/qubes-core-admin/-/jobs/15513318441.


marmarek:
Hm, is there an option to force specific format version? Or that requires freezing black version itself?

ben-grande:
Only found two options from the manual page:

```
  -S, --skip-string-normalization
                                  Don't normalize string quotes or prefixes.
  -C, --skip-magic-trailing-comma
                                  Don't use trailing commas as a reason to
                                  split lines.
```

And that wouldn't help broadly in the case of this PR.

marmarek (approved):

PR #856, opened 2026-07-28
Test that packagekit works via proxy

notmuchtohide:
Here is the [requested](https://github.com/QubesOS/qubes-core-agent-linux/pull/659#issuecomment-4972251476) test.

Packagekit still needs to be installed in the template before the test. Should it be installed in the test function or somewhere else?

I'm also not sure about the right location to place the test, or test number.

### DIFF
--- a/qubes/tests/integ/vm_update.py
+++ b/qubes/tests/integ/vm_update.py
@@ -755,6 +755,13 @@ SHA256:
                 self.testvm1, "pkcon install -y test-pkg", self.ret_code_ok
             )
 
+            # verify if it was really installed
+            self.assertRunCommandReturnCode(
+                self.testvm1,
+                self.install_test_cmd.format("test-pkg"),
+                self.ret_code_ok,
+            )
+
     def upgrade_status_notify(self):
         """
         Run upgrades-status-notify at test vm.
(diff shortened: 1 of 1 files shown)


ben-grande:
> Should it be installed in the test function or somewhere else?

https://github.com/QubesOS/openqa-tests-qubesos/blob/main/extra-files/system-tests/vm.sls

qubes/tests/integ/vm_update.py
```diff
@@ -729,6 +729,32 @@ def test_010_update_via_proxy(self):
...
+
+            # update repository metadata
+            self.assertRunCommandReturnCode(
+                self.testvm1, "pkcon refresh", self.ret_code_ok
+            )
+
+            # install test package
+            self.assertRunCommandReturnCode(
+                self.testvm1, "pkcon install -y test-pkg", self.ret_code_ok
+            )
```
marmarek, line 756: I think this wants a final check if the package really got installed, similar to the other test:
```
            # verify if it was really installed
            self.assertRunCommandReturnCode(
                self.testvm1,
                self.install_test_cmd.format("test-pkg"),
                self.ret_code_ok,
            )
```

marmarek (approved):

PR #857, opened 2026-07-29
q-dev: expose device assignment modes

piotrbartman:
Until now, the available assignment modes were hidden in code without an endpoint in the API.

related: QubesOS/qubes-core-admin-client/pull/487
related: QubesOS/qubes-app-linux-usb-proxy/pull/61

### DIFF
--- a/qubes/device_protocol.py
+++ b/qubes/device_protocol.py
@@ -899,8 +899,24 @@ class DeviceInterface:
         return True
 
 
+class AssignmentMode(Enum):
+    """
+    Device assignment modes
+    """
+
+    MANUAL = "manual"
+    ASK = "ask-to-attach"
+    AUTO = "auto-attach"
+    REQUIRED = "required"
+
+
 class DeviceInfo(VirtualDevice):
     """Holds all information about a device"""
+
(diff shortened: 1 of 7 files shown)


qubes/device_protocol.py
```diff
@@ -899,9 +899,25 @@ def matches(self, other: "DeviceInterface") -> bool:
...
+    ASK = "ask-to-attach"
+    AUTO = "auto-attach"
+    REQUIRED = "required"
+
+
 class DeviceInfo(VirtualDevice):
     """Holds all information about a device"""
 
+    # Subclasses should override this.
+    SUPPORTED_ASSIGNMENT_MODES: "frozenset[AssignmentMode]" = frozenset(
```
marmarek, line 917: Would it make sense to include `manual` here too? Then, missing `manual` would mean device can only be assigned (likely only with `required` mode), but manual attach is not permitted.

piotrbartman: Yup, I like this

marmarek:
Test needs an update to include `manual` too

marmarek (approved):

PR #827, opened 2026-06-11
Report more statistics

ben-grande:
- online vcpus
- time of each vcpu
- share all vm_info captured with clients (I presume there is no intent to capture private information internally and not broadcast to clients, as the only place where `get_vm_stats()` is called is from `admin.vm.Stats`.

---

Requires:

- https://github.com/QubesOS/qubes-linux-utils/pull/140

PR #859, opened 2026-07-31
Cache VM stats

ben-grande:
TTL cache to avoid multiple clients or requests from the same client of blocking CPU cycles from qubesd.


ben-grande:
Marek said: https://github.com/QubesOS/qubes-core-admin/pull/827#discussion_r3690517835

> The change to use `self.stats_cache_time` instead of the explicit
> value from the parameter is not great. If nobody is updating
> (requesting) stats for some time, next one will return average of
> a long time pretending it's the "current" value.

Let's assume the previous moment the stats were collected was a long
time ago.

- Empty `previous`, stats never called:
  - Without this PR: 0%
  - With PR: 0%
  - Resolution: Unaffected

- Empty `previous`, but stats generated at least once:
  - Without this PR: 0%
  - With PR: 0%, `calculate_cpu_usage_raw()` is not called as
    `domid` is not in `previous`, as there is no `previous`
  - Resolution: Unaffected

- Non-empty `previous`:
  - Without PR: avg considering the previous time
  - With PR: avg considering the previous time
  - Resolution: It already happened if the client is not using the
    API loop and calling `get_vm_stats()` directly

> - Empty `previous`, but stats generated at least once:
> ...
>   - With PR: 0%, `calculate_cpu_usage_raw()` is not called as
>     `domid` is not in `previous`, as there is no `previous`

But let's assume that I did an average of a long time by assigning `previous` to the last cache if `previous=None`:

```python3
if not previous and self.stats_cache_data:
    previous = self.stats_cache_data
```

I don't think it would be wrong to have an average of a long time compared to reporting 0%.

ben-grande:
If you don't like the `_get_vm_stats` checking for the cache. I could make it default to ignore so the behavior doesn't change (even though it is a protected version), and then, if calling via `admin.vm.Stats`, always use the cache.

PR #860, opened 2026-07-31
A vm should not be named 'Dom0'

marmarta:
That's extremely confusing.

Thanks to @tokideveloper for finding this!

### DIFF
--- a/qubes/vm/__init__.py
+++ b/qubes/vm/__init__.py
@@ -84,8 +84,8 @@ def validate_name(holder, prop, value):
                 "{} value contains illegal characters".format(prop.__name__),
             )
         raise qubes.exc.QubesValueError("VM name contains illegal characters")
-    if value == "Domain-0":
-        raise qubes.exc.QubesValueError("VM name cannot be 'Domain-0'")
+    if value in ("Domain-0", "Dom0"):
+        raise qubes.exc.QubesValueError("VM name cannot be 'Domain-0' or 'Dom0'")
     if value in ("none", "default"):
         raise qubes.exc.QubesValueError(
             "VM name cannot be 'none' nor 'default'"
--- a/qubes/vm/__init__.py
+++ b/qubes/vm/__init__.py
@@ -84,8 +84,10 @@ def validate_name(holder, prop, value):
                 "{} value contains illegal characters".format(prop.__name__),
             )
         raise qubes.exc.QubesValueError("VM name contains illegal characters")
-    if value == "Domain-0":
-        raise qubes.exc.QubesValueError("VM name cannot be 'Domain-0'")
+    if value in ("Domain-0", "Dom0"):
+        raise qubes.exc.QubesValueError(
+            "VM name cannot be 'Domain-0' or 'Dom0'"
+        )
     if value in ("none", "default"):
         raise qubes.exc.QubesValueError(
             "VM name cannot be 'none' nor 'default'"


rustybird:
Or a case-insensitive comparison to also catch DOM0 etc.?

ben-grande:
And the upper case letter `o` than? I don't think we can get rid of look a likes.

```
dom0
dom0.
domO
domO.
Dom0
Dom0.
DomO
DomO.
D0M0
D0M0.
D0MO
D0MO.
DOM0
DOM0.
DOMO
DOMO.

dom0- # but this is spottable
D0m0 # but this is spottable
```

And this is not a complete list, but mixing upper-case and lower-case is more easy to distinguish as a bad name.

- If you notice, upper case letters allows the same name to be represented in upper and lower case, and those are different names. IMHO, upper case should never have been allowed.
- Names can end with a dot, that may be difficult to see.
- Names can end with a hyphen, not so difficult to see

If we restrict the name to not allow upper case and ending with a symbol, there is no look-a-like anymore. I understand this is breaking change, I don't think name normalization can safely work on backup restore of a major upgrade with a stricter validation.

marmarta:
This one case is particularly evil, because in XFCE for some godforsaken reason dom0 in window title is capitalized.

marmarek:
While banning such name is probably a good idea, this can't be shipped to R4.3, otherwise it would basically brick systems that already have such qube (and we know there is at least one such person). So, it's okay to merge it to main, and then add special handling to the upgrade tool (likely to detect before upgrade, and ask user to remove/rename such qubes).

EDIT: This is about this PR in general, not specifically the later list in Ben's comment.

marmarek:
pylint complains, and black proposes a fix

marmarek:
Now all tests exploded, because it refuses also `dom0`, so it fails access to the actual dom0... Anyway, I don't think it's the right approach. Refusing literal `Dom0` is desirable only because that's what Xfce have (had) in window titles. But otherwise treating just dom0 variants as special is quite confusing, and potentially never ends (should `don0` be also banned? or `domO` ?). In practice, confusing `vault` with `vau1t` would likely be more impactful, yet I think we both agree both are valid names.

marmarta:
PipelineRetry

marmarek (approved):

ben-grande:
Sorry for the indentation, it is logged like this from integration tests:

> 2026-09-02 22:12:24,351 INFO appmenus[30784]: Updating appmenus for 'disp1229' in 'dom0'
> 2026-09-02 22:12:24,868 ERROR appmenus[30784]: Failed to update appmenus for 'disp1229' in 'dom0': Traceback (most recent call last):_  File "/usr/bin/qvm-appmenus", line 33, in <module>_    sys.exit(load_entry_point('qubesdesktop==4.4.0', 'console_scripts', 'qvm-appmenus')())_             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^_  File "/usr/lib/python3.13/site-packages/qubesappmenus/__init__.py", line 881, in main_    appmenus.appmenus_update(vm, force=args.force)_    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^_  File "/usr/lib/python3.13/site-packages/qubesappmenus/__init__.py", line 726, in appmenus_update_    for child_vm in vm.appvms:_                    ^^^^^^^^^_  File "/usr/lib/python3.13/site-packages/qubesadmin/vm/__init__.py", line 427, in appvms_    if vm.template == self:_       ^^^^^^^^^^^_  File "/usr/lib/python3.13/site-packages/qubesadmin/base.py", line 263, in __getattr___    property_str = self.qubesd_call(_        self._method_dest,_        self._method_prefix + 'Get',_        item,_        None)_  File "/usr/lib/python3.13/site-packages/qubesadmin/base.py", line 107, in qubesd_call_    return self.app.qubesd_call(dest, method, arg, payload,_           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^_        payload_stream)_        ^^^^^^^^^^^^^^^_  File "/usr/lib/python3.13/site-packages/qubesadmin/app.py", line 922, in qubesd_call_    return self._parse_qubesd_response(return_data)_           ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^_  File "/usr/lib/python3.13/site-packages/qubesadmin/base.py", line 141, in _parse_qubesd_response_    raise exc_class(format_string, *args)_qubesadmin.exc.QubesValueError: VM name cannot be 'Domain-0' or 'Dom0'_
>

ben-grande:
So the exception handling works, everything is fine. Ignore previous message, I have a wrongly named qube...

PR #870, opened 2026-08-13
Deferred template

### LINKED ISSUE #8070: Allow deferred template changes

Currently, the Qube Manager will not allow you to change a running qube's template. (The field is grayed out.) The qube must be shut down first.

In addition, if you try to change the template on which a disposable template is based while there are running disposables based on that disposable template, you will receive an error message stating that the action cannot be performed due to the running disposables.

### The solution you'd like

As @marmarek suggested in https://github.com/QubesOS/qubes-doc/pull/1297#issuecomment-1454343087, it should be possible for the user to change the template in the GUI even when qubes based on that template are running, but defer the actual change until after those running qubes shut down or reboot.

### The value to a user, and who that user might be

Users will be able to tell the system what to do at any time, and the system will more intelligently handle those directives at the appropriate time, relieving users of some micromanagement.

ben-grande:
Add ability for template based qubes (AppVM and DispVM) to be able to change the template while running. When the qube is running and you modify the `template` property, it is set to the `newvalue`, while the `active_template` changes to the `oldvalue` if it is still the default.

The use of separate properties doesn't break new clients. For example, Qube Manager reacts to `property-set:template` and adapts the field to the current value, despite it not being the active one. This is what happens for other properties also, such as `virt_mode` and `kernel`, you may change it live, but it only applies on the next boot.

Fixes: https://github.com/qubesos/qubes-issues/issues/8070

Clients should deal with the new property this way:

- react to `property-(re)?set:(active_)?template)`
- if `active_template` property is default, show only it, else, show it alongside the `template` indicating that it is deferred, this is for a minimal view with less clutter. On clients that would be better to show two columns containing both values, that is also fine.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -59,10 +59,10 @@ const-rgx=(([A-Za-z_][A-Za-z0-9_]*)|(__.*__))$
 class-rgx=([A-Z_][a-zA-Z0-9]+|TC_\d\d_[a-zA-Z0-9_]+)$
 
 # Regular expression which should only match correct function names
-function-rgx=(test_[0-9]{3}_[a-z0-9_]{2,50}|[a-z_][a-z0-9_]{2,50})$
+function-rgx=(test_[0-9]{3}_[a-z0-9_]{2,60}|[a-z_][a-z0-9_]{2,60})$
 
 # Regular expression which should only match correct method names
-method-rgx=(setUp|setUpClass|tearDown|tearDownClass|test_[0-9]{3}_[a-z0-9_]{2,50}|[a-z_][a-z0-9_]{2,50})$
+method-rgx=(setUp|setUpClass|tearDown|tearDownClass|test_[0-9]{3}_[a-z0-9_]{2,60}|[a-z_][a-z0-9_]{2,60})$
 
 # Regular expression which should only match correct instance attribute names
 attr-rgx=(maxDiff|[a-z_][a-z0-9_]{2,40})$
--- a/qubes/api/admin.py
+++ b/qubes/api/admin.py
@@ -1401,6 +1401,16 @@ class QubesAdminAPI(qubes.api.AbstractQubesAPI):
             raise qubes.exc.QubesFeatureNotFoundError(self.dest, self.arg)
         return value
 
+    def _get_template_feature_options(self):
+        active = True
+        if "+" in self.arg:
+            untrusted_feature, untrusted_options = self.arg.split("+", 1)
+            if untrusted_options == "deferred":
+                active = False
+        else:
+            untrusted_feature = self.arg
+        return untrusted_feature, active
+
     @qubes.api.method(
         "admin.vm.feature.CheckWithTemplate",
         wants_arg=True,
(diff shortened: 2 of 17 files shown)


ben-grande:
openQArun PR_LABEL=openqa-group-2

ben-grande:
openQArun PR_LABEL=openqa-group-2

ben-grande:
https://openqa.qubes-os.org/tests/192952#step/TC_00_Basic/19

`system_tests_basic_vm_qrexec_gui_btrfs`
```python3

# test_300_deferred_template
# failure: 

# timestamp 2026-08-19T15:29:38.593611
    return func(*args, **kwds)
  File "/usr/lib/python3.13/site-packages/qubes/tests/integ/basic.py", line 590, in test_300_deferred_template
    self.loop.run_until_complete(self._test_300_deferred_template())
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/asyncio/base_events.py", line 725, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/qubes/tests/integ/basic.py", line 600, in _test_300_deferred_template
    await self.defer_tpl(appvm)
  File "/usr/lib/python3.13/site-packages/qubes/tests/__init__.py", line 1790, in defer_tpl
    check_deferred(qube, active=template, deferred=template)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/qubes/tests/__init__.py", line 1770, in check_deferred
    self.assertEqual(get_tpl_vol_vid(qube), get_tpl_vol_path(active))
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: 'fedora-44-xfce-pool/root' != 'vm-fedora-44-xfce-pool-root'
- fedora-44-xfce-pool/root
?                    ^
+ vm-fedora-44-xfce-pool-root
? +++                   ^
```

Different fs are having different names here... I need to compare the volume vid and the path differently...

ben-grande:
> ```python
>     self.assertEqual(get_tpl_vol_vid(qube), get_tpl_vol_path(active))
>     ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> AssertionError: 'fedora-44-xfce-pool/root' != 'vm-fedora-44-xfce-pool-root'
> - fedora-44-xfce-pool/root
> ?                    ^
> + vm-fedora-44-xfce-pool-root
> ? +++                   ^
> ```

I didn't find an easy way to compare. Technically, each pool type should have the same method to query it's format, by providing the qube name and volume name, and then it returns the volume string. I also noticed the pools are switched by OpenQA, so there isn't a test doing that that I can run locally, well, I could copy the whole pool switch code... I ended up checking each pool volume format, and just ThinPool for LVM uses a certain format with `-` while the others use `/`.

ben-grande:
openQArun PR_LABEL=openqa-group-2

marmarek:
Can't you simply take properties of relevant volume objects instead of building the id manually?

ben-grande:
openQArun PR_LABEL=openqa-group-2

ben-grande:
openQArun PR_LABEL=openqa-group-2

ben-grande:
I don't see the pipeline in neither:

- https://gitlab.com/QubesOS/qubes-core-admin/-/pipelines
- https://openqa.qubes-os.org/group_overview/11

ben-grande:
> * https://gitlab.com/QubesOS/qubes-core-admin/-/pipelines
> 
> * [](https://openqa.qubes-os.org/group_overview/11)

Nevermind, it is in https://gitlab.com/QubesOS/qubes-continuous-integration/-/pipelines/2806087686.

ben-grande:
Just solving merge conflict.

PR #871, opened 2026-08-15
storage: persistent per-volume LUKS2 encryption (#1293)

rishi-jat:
### Summary

First slice of QubesOS/qubes-issues#1293: persistent per-volume encryption on the existing storage stack (same approach as ephemeral volatile encryption in #396).

### Behavior

- `Volume.encrypted` on `save_on_stop` + `rw` + dom0 volumes (private on AppVM/Standalone, root on Template/Standalone). Mutually exclusive with `ephemeral`.
- LUKS2 (`luksFormat` / `open --type=luks2`). Passphrase stays in qubesd memory only (`admin.vm.volume.SetPassphrase`); not XML, not `volume.Info`.
- `encrypted=` is visible in `admin.vm.volume.Info` so start tools know to prompt first.
- Enable: empty volume → `luksFormat`; volume with data → grow 32 MiB + `reencrypt --encrypt`. Dirty volumes and volumes with revisions are refused.
- Start unlocks, stop closes. Failed open/resize after `vol.start()` rolls back `vol.stop()` and keeps the passphrase for retry.
- Clone copies the LUKS container raw (same passphrase).
- `admin.vm.volume.ChangePassphrase` payload is `old\nnew`.
- Disabling encryption is not implemented (explicit error).
- Snapshot consumers (AppVM from encrypted template root, DispVM from encrypted AppVM private) are refused.

### Out of scope

- Decrypt / disable
- Encrypting a source that already has snap-on-start children
- Automatic revision cleanup before encryption
- Live cryptsetup integration tests on CI
### Test plan

```bash
./run-tests qubes.tests.storage_encrypt qubes.tests.api_admin \
  qubes.tests.storage qubes.tests.storage_file qubes.tests.storage_lvm
```
On a real system: enable on a new private volume, start/stop, wrong passphrase then retry, clone, refuse encrypt on a template that already has AppVMs.

### DIFF
--- a/qubes/api/admin.py
+++ b/qubes/api/admin.py
@@ -481,6 +481,7 @@ class QubesAdminAPI(qubes.api.AbstractQubesAPI):
             "snap_on_start",
             "revisions_to_keep",
             "ephemeral",
+            "encrypted",
         ]
 
         def _serialize(value):
@@ -768,6 +769,121 @@ class QubesAdminAPI(qubes.api.AbstractQubesAPI):
 
         self.dest.volumes[self.arg].ephemeral = newvalue
         self.app.save()
+
+    @qubes.api.method(
+        "admin.vm.volume.Set.encrypted",
+        wants_arg=True,
+        wants_payload=True,
+        dest_adminvm=None,
+        scope="local",
+        write=True,
+    )
+    async def vm_volume_set_encrypted(self, untrusted_payload):
+        self.enforce_arg(
+            wants=self.dest.volumes.keys(),
+            short_reason=self.EXC_ARG_NOT_IN_DEST_VOLUMES,
+        )
+        try:
+            newvalue = qubes.property.bool(
+                None, None, untrusted_payload.decode("ascii")
+            )
+        except (UnicodeDecodeError, ValueError):
+            raise qubes.exc.ProtocolError("Payload is not boolean ASCII")
+        del untrusted_payload
+
+        self.fire_event_for_permission(newvalue=newvalue)
+
+        if not self.dest.is_halted():
+            raise qubes.exc.QubesVMNotHaltedError(self.dest)
+
+        volume = self.dest.volumes[self.arg]
+        if not newvalue:
+            if volume.encrypted:
+                raise qubes.exc.QubesValueError(
+                    "Disabling persistent encryption is not implemented"
+                )
+            return
+
+        # Snapshot consumers would see a locked LUKS container.
+        # Compare by pool+vid, not identity.
+        consumers = list(qubes.storage.snapshot_consumers(self.app, volume))
+        if consumers:
+            raise qubes.exc.QubesValueError(
+                "Cannot encrypt a volume that is a snapshot source "
+                "for other qubes"
+            )
+        if not volume.has_passphrase():
+            raise qubes.exc.QubesException(
+                "Passphrase must be set before enabling encryption"
+            )
+        volume.encrypted = True
+        try:
+            await volume.setup_luks()
+        except Exception:
+            # Once the backing device has been mutated, keep the flag
+            # so start will not attach ciphertext as a normal disk.
+            if not getattr(volume, "_luks_device_mutated", False):
+                volume._encrypted = False  # pylint: disable=protected-access
+            else:
+                self.app.save()
+            raise
+        self.app.save()
+
+    @qubes.api.method(
+        "admin.vm.volume.SetPassphrase",
+        wants_arg=True,
+        wants_payload=True,
+        dest_adminvm=None,
+        scope="local",
+        write=True,
+    )
+    async def vm_volume_set_passphrase(self, untrusted_payload):
+        """Set the in-memory LUKS passphrase.  Not serialized to XML."""
+        self.enforce_arg(
+            wants=self.dest.volumes.keys(),
+            short_reason=self.EXC_ARG_NOT_IN_DEST_VOLUMES,
+        )
+        # Do not include the passphrase in the permission event.
+        self.fire_event_for_permission()
+        volume = self.dest.volumes[self.arg]
+        if not volume.encrypted and not volume.is_encryptable():
+            raise qubes.exc.QubesValueError(
+                "Volume does not support persistent encryption"
+            )
+        volume.set_passphrase(untrusted_payload)
+        del untrusted_payload
+        # Intentionally not saved: the passphrase must never hit qubes.xml.
+
+    @qubes.api.method(
+        "admin.vm.volume.ChangePassphrase",
+        wants_arg=True,
+        wants_payload=True,
+        dest_adminvm=None,
+        scope="local",
+        write=True,
+    )
+    async def vm_volume_change_passphrase(self, untrusted_payload):
+        """Replace the LUKS passphrase.  Payload is ``old\\nnew``."""
+        self.enforce_arg(
+            wants=self.dest.volumes.keys(),
+            short_reason=self.EXC_ARG_NOT_IN_DEST_VOLUMES,
+        )
+        self.fire_event_for_permission()
+        if b"\n" not in untrusted_payload:
+            raise qubes.exc.ProtocolError(
+                "Payload must be old passphrase, newline, new passphrase"
+            )
+        old, new = untrusted_payload.split(b"\n", 1)
+        del untrusted_payload
+        if b"\n" in new:
+            raise qubes.exc.QubesValueError(
+                "Passphrase must not contain newline character"
+            )
+        volume = self.dest.volumes[self.arg]
+        if not volume.encrypted:
+            raise qubes.exc.QubesValueError("Volume is not encrypted")
+        await volume.change_passphrase(old, new)
+        # Intentionally not saved.
 
     @qubes.api.method(
         "admin.vm.tag.List",
--- a/qubes/api/admin.py
+++ b/qubes/api/admin.py
@@ -482,6 +482,7 @@ class QubesAdminAPI(qubes.api.AbstractQubesAPI):
             "revisions_to_keep",
             "ephemeral",
             "encrypted",
+            "has_passphrase",
         ]
 
         def _serialize(value):
@@ -816,17 +817,20 @@ class QubesAdminAPI(qubes.api.AbstractQubesAPI):
             raise qubes.exc.QubesException(
                 "Passphrase must be set before enabling encryption"
             )
-        volume.encrypted = True
-        try:
-            await volume.setup_luks()
-        except Exception:
-            # Once the backing device has been mutated, keep the flag
-            # so start will not attach ciphertext as a normal disk.
-            if not getattr(volume, "_luks_device_mutated", False):
-                volume._encrypted = False  # pylint: disable=protected-access
-            else:
-                self.app.save()
-            raise
+        # Persist the flag *before* mutating the device so a crash
+        # cannot leave qubes.xml saying the volume is plaintext.
+        # Refuse a guest-written LUKS header (including cipher=none).
+        if not volume.encrypted:
+            if await volume.is_luks():
+                raise qubes.exc.QubesValueError(
+                    "Volume already has a LUKS header; refuse to mark it "
+                    "encrypted without formatting it with the requested "
+                    "passphrase"
+                )
+            volume._assert_safe_to_encrypt()  # pylint: disable=protected-access
+            volume.encrypted = True
+            self.app.save()
+        await volume.setup_luks()
         self.app.save()
 
     @qubes.api.method(
--- a/qubes/storage/__init__.py
+++ b/qubes/storage/__init__.py
@@ -24,6 +24,7 @@
 
 import functools
 import inspect
+import logging
 import os
 import os.path
 import string
@@ -291,8 +292,10 @@ class Volume:
         """Persistent LUKS2 encryption of this volume.
 
         Allowed only on writable, save_on_stop, non-snapshot, dom0-backed
-        volumes (private on AppVM/StandaloneVM, root on TemplateVM /
-        StandaloneVM).  Mutually exclusive with :py:attr:`ephemeral`.
+        volumes.  The practical target is ``private`` on AppVM/StandaloneVM
+        (and ``root`` on StandaloneVM).  TemplateVM ``root`` meets the
+        mechanical checks but is not useful: AppVMs cannot snapshot an
+        encrypted source.  Mutually exclusive with :py:attr:`ephemeral`.
         The passphrase is never stored on disk; see
         :py:meth:`set_passphrase`.
         """
@@ -307,16 +310,7 @@ class Volume:
                 )
             self._encrypted = False
             return
-        if self.ephemeral:
-            raise qubes.exc.QubesValueError(
-                "Cannot enable encryption on ephemeral volume"
-            )
-        if (
-            self.snap_on_start
-            or not self.save_on_stop
-            or self.domain is not None
-            or not self.rw
-        ):
+        if not self.is_encryptable():
             raise qubes.exc.QubesValueError(
                 "Cannot enable encryption on snap_on_start or "
                 "non-save_on_stop or non-dom0 or not writable volume"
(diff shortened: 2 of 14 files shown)


rishi-jat:
Note: I used agent for the rebase.

Rebased onto current `main` (`e231e518`) to clear the merge conflicts. The three logical commits are unchanged in purpose:

1. `storage: add persistent LUKS2 volume encryption`
2. `admin: add volume encryption and passphrase APIs`
3. `tests: cover per-volume LUKS2 encryption`

Conflict resolutions:
- `qubes/storage/__init__.py`: kept LUKS start/stop (`start_luks`/`stop_luks`) and always call `stop_encrypted()` for non-encrypted volumes (so flipping `ephemeral` while running still cleans up). `import_volume` keeps the encryption eligibility checks, then fires `domain-import-volume`.
- `doc/qubes-storage.rst`: kept current `main` list style (`-`) and the `Volume.encrypted` paragraph.

The dummy-VM import test now stubs `fire_event_async` so it matches the new import event.

rishi-jat:
Follow-up after the first post-rebase pipeline:

- `lint`: Black wrapping in `qubes/storage/__init__.py` and `qubes/tests/storage_encrypt.py`.
- `r4.3:build:host-fc41`: package `qubes/tests/storage_encrypt.py` in `rpm_spec/core-dom0.spec.in` (unpackaged file).
- `checks:pylint` was 10.00/10 and then failed with a runner system error (`Custom executor exit code 16`).

Still three commits. New tip: `e5f6abc7`.

rishi-jat:
CC @marmarek

qubes/storage/__init__.py
```diff
@@ -48,6 +48,55 @@
 BYTES_TO_ZERO = 1 << 16
 _big_buffer = b"\0" * BYTES_TO_ZERO
 
+# LUKS2 header / data-offset reservation used when encrypting an existing
+# volume in place.  cryptsetup reencrypt --reduce-device-size takes this
+# much from the end of the device so the guest-visible size stays the same
+# after the header is written.  32 MiB matches cryptsetup's recommended
+# default for LUKS2.
+LUKS2_HEADER_SIZE = 32 << 20
```
marmarek, line 57: There should be a test to verify this constant.

qubes/storage/__init__.py
```diff
@@ -214,6 +286,342 @@ def ephemeral(self, value):
             )
         self._ephemeral = bool(value)
 
+    @property
+    def encrypted(self):
+        """Persistent LUKS2 encryption of this volume.
+
+        Allowed only on writable, save_on_stop, non-snapshot, dom0-backed
+        volumes (private on AppVM/StandaloneVM, root on TemplateVM /
```
marmarek: While technically TemplateVMs's root is allowed here, it isn't practical, since you block creating AppVMs based on it (then "Cannot create a snapshot volume from an encrypted source" message above).

qubes/storage/__init__.py
```diff
@@ -214,6 +286,342 @@ def ephemeral(self, value):
...
+            return
+        if self.ephemeral:
+            raise qubes.exc.QubesValueError(
+                "Cannot enable encryption on ephemeral volume"
+            )
+        if (
+            self.snap_on_start
+            or not self.save_on_stop
+            or self.domain is not None
+            or not self.rw
```
marmarek: Use `is_encryptable` here to reduce duplication?

qubes/storage/__init__.py
```diff
@@ -214,6 +286,342 @@ def ephemeral(self, value):
...
+                    self.vid
+                )
+            )
+        if device is None:
+            device = self._luks_backend_path()
+        if not device or not os.path.exists(device):
+            raise StoragePoolException(
+                "Cannot set up LUKS: device {!r} does not exist".format(device)
+            )
+        if await self.is_luks(device):
```
marmarek, line 479: There is a corner case here: if the qube created LUKS header itself on its own, then the data won't get re-encrypted (with requested passphrase), yet the volume will be marked as encrypted. Usually it will "just" result in a failure later on when starting it (due to a different passphrase), but it could also create a header with `none` cipher, to avoid encryption at all.

I'd propose to fix it this way:
1. Before enabling encryption, check if there is already existing LUKS header - if yes, refuse.
2. If no, continue, save volume.encrypted=True (and really save it by calling `app.save()`, in case the process gets interrupted).
3. Only then do the LUKS setup.

qubes/storage/__init__.py
```diff
@@ -214,6 +286,342 @@ def ephemeral(self, value):
...
+        mapper_name = name[12:]
+        assert "/" not in mapper_name, "Invalid name passed to cryptsetup"
+        if not self.has_passphrase():
+            raise StoragePoolException(
+                "Passphrase required to unlock encrypted volume {!s}".format(
+                    self.vid
+                )
+            )
+        if os.path.exists(name):
+            await qubes.utils.cryptsetup("--", "close", mapper_name)
```
marmarek, line 554: When it can happen? Looks like an unexpected situation, worthy at least a warning log entry, if not an exception.

qubes/api/admin.py
```diff
@@ -769,6 +770,121 @@ async def vm_volume_set_ephemeral(self, untrusted_payload):
...
+        if consumers:
+            raise qubes.exc.QubesValueError(
+                "Cannot encrypt a volume that is a snapshot source "
+                "for other qubes"
+            )
+        if not volume.has_passphrase():
+            raise qubes.exc.QubesException(
+                "Passphrase must be set before enabling encryption"
+            )
+        volume.encrypted = True
```
marmarek: setup_luks already sets this flag, no? See also the comment about pre-existing LUKS

qubes/storage/__init__.py
```diff
@@ -214,6 +286,342 @@ def ephemeral(self, value):
...
+                    "Failed to unlock encrypted volume {!s}".format(self.vid)
+                ) from exc
+            # Apply any offline backing-store grow to the LUKS payload.
+            try:
+                await qubes.utils.cryptsetup("--", "resize", mapper_name)
+            except subprocess.CalledProcessError as exc:
+                raise StoragePoolException(
+                    "Failed to resize unlocked volume {!s}".format(self.vid)
+                ) from exc
+            self.clear_passphrase()
```
marmarek, line 586: This is missing in the failed case, intentional? Maybe there should be some flag visible (expose `has_passphrase` to be visible in `qvm-volume info`?) when the passphrase is set, so some GUI can warn the user if they are left with cached passphrase after failed startup.

But also, clearing the passphrase on successful start will kinda conflict with restarting (https://github.com/QubesOS/qubes-issues/issues/8196). But at this point it isn't a request to cache it longer.

marmarek (changes requested):
Nice work :)

Besides the comments below, this could also use an integration test, see qubes/tests/integ/storage.py. It doesn't need to be too detailed, but should cover basic cases:
 - enabling encryption
 - starting / stopping volume (and checking if data properly survives)
 - maybe also encrypting volume with existing data (and again, checking if data is intact)

`test_003_snapshot` might be a good inspiration.

qubes/storage/__init__.py
rishi-jat: It can happen after a failed start. If the open/resize fails, we try to roll back, but the mapper can still be left behind if that cleanup doesn’t complete or the process is interrupted.

I will keep the close-if-exists handling and add a warning when it happens.

qubes/api/admin.py
rishi-jat: Yes. The setter currently sets it before calling setup_luks(), which sets it again. will remove that duplication and let the encryption setup handle the flag after the existing-LUKS check.

qubes/storage/__init__.py
rishi-jat: Yes, intentional. The passphrase is kept after a failed open/resize so the user can retry without entering it again. It’s cleared after a successful start.

Exposing has_passphrase in admin.vm.volume.Info makes sense, I’ll add that so the GUI can detect when a passphrase is still cached after a failed start.

I’m leaving the successful-start clearing behavior unchanged for this PR.

rishi-jat:
CC @marmarek

marmarek:
Black complains about one line in tests

marmarek:
openQArun MACHINE=64bit TEST=system_tests_gui_interactive,system_tests_basic_vm_qrexec_gui,system_tests_pvgrub_salt_storage,system_tests_gui_interactive,system_tests_gui_tools,system_tests_backup

marmarek:
> system_tests_pvgrub_salt_storage
> 
>     * StorageFile: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/195034#step/StorageFile/9) (error)
>       `qubes.exc.StoragePoolException: Cannot encrypt a volume that has re...`
> 
>     * StorageLVM: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/195034#step/StorageLVM/9) (error)
>       `qubes.exc.StoragePoolException: Cannot encrypt a volume that has re...`
> 
>     * StorageReflinkOnBtrfs: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/195034#step/StorageReflinkOnBtrfs/9) (error)
>       `qubes.exc.StoragePoolException: Cannot set up LUKS: device '/var/tm...`
> 
>     * StorageReflinkOnExt4: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/195034#step/StorageReflinkOnExt4/9) (error)
>       `qubes.exc.StoragePoolException: Cannot set up LUKS: device '/var/tm...`

This is definitely related to this PR. I'm not 100% sure if this is more about issue in tests, or implementation, but needs some fixes....

qubes/storage/__init__.py
```diff
@@ -214,6 +287,355 @@ def ephemeral(self, value):
...
+        Drivers that expose a real container (file, LVM) set :py:attr:`path`.
+        Reflink keeps the committed image in ``_path_clean`` and points
+        :py:attr:`path` at the dirty file, which is gone after stop.
+        Prefer a path that actually exists.  Others (ZFS) fall back to
+        the block device.
+        """
+        candidates = []
+        if self.path:
+            candidates.append(self.path)
+        clean = getattr(self, "_path_clean", None)
```
marmarek: Accessing driver-specific field here is a layering violation. I'm not sure why reflink driver behaves this way, might be a bug (@rustybird  can you help?). But if that's intentional, better override the luks_backend_path function in the reflink driver with driver-specific logic (and not mark it as private anymore).

rishi-jat: Yeah, that’s intentional in Reflink. Fixed by moving the clean/dirty path handling into the public luks_backend_path() override.

marmarek:
openQArun MACHINE=64bit TEST=system_tests_gui_interactive,system_tests_basic_vm_qrexec_gui,system_tests_pvgrub_salt_storage,system_tests_gui_interactive,system_tests_gui_tools,system_tests_backup

marmarek:
> * system_tests_pvgrub_salt_storage
>       
>       * StorageFile: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/195973#step/StorageFile/9) (error)
>         `qubes.exc.StoragePoolException: Failed to resize unlocked volume ap...`
>       * StorageLVM: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/195973#step/StorageLVM/9) (error)
>         `qubes.exc.StoragePoolException: Failed to resize unlocked volume qu...`
>       * StorageReflinkOnBtrfs: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/195973#step/StorageReflinkOnBtrfs/9) (error)
>         `qubes.exc.StoragePoolException: Failed to resize unlocked volume ap...`
>       * StorageReflinkOnExt4: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/195973#step/StorageReflinkOnExt4/9) (error)
>         `qubes.exc.StoragePoolException: Failed to resize unlocked volume ap...`
> 

Now it fails this way

rishi-jat:
> > * system_tests_pvgrub_salt_storage
> >   ```
> >   * StorageFile: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/195973#step/StorageFile/9) (error)
> >     `qubes.exc.StoragePoolException: Failed to resize unlocked volume ap...`
> >   * StorageLVM: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/195973#step/StorageLVM/9) (error)
> >     `qubes.exc.StoragePoolException: Failed to resize unlocked volume qu...`
> >   * StorageReflinkOnBtrfs: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/195973#step/StorageReflinkOnBtrfs/9) (error)
> >     `qubes.exc.StoragePoolException: Failed to resize unlocked volume ap...`
> >   * StorageReflinkOnExt4: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/195973#step/StorageReflinkOnExt4/9) (error)
> >     `qubes.exc.StoragePoolException: Failed to resize unlocked volume ap...`
> >   ```
> 
> Now it fails this way

@marmarek Could you please rerun system_tests_pvgrub_salt_storage and check test_010_persistent_luks with the latest fix in 11b7064?

marmarek:
openQArun MACHINE=64bit TEST=system_tests_basic_vm_qrexec_gui,system_tests_pvgrub_salt_storage,system_tests_backup

marmarek:
> * StorageFile: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/196037#step/StorageFile/9) (error)
>   `subprocess.CalledProcessError: Command 'echo test123 | cat - /dev/z...`
> 
> * StorageLVM: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/196037#step/StorageLVM/9) (error)
>   `subprocess.CalledProcessError: Command 'echo test123 | cat - /dev/z...`
> 
> * StorageReflinkOnBtrfs: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/196037#step/StorageReflinkOnBtrfs/9) (error)
>   `subprocess.CalledProcessError: Command 'echo test123 | cat - /dev/z...`
> 
> * StorageReflinkOnExt4: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/196037#step/StorageReflinkOnExt4/9) (error)
>   `subprocess.CalledProcessError: Command 'echo test123 | cat - /dev/z...`

Now the test says data doesn't survive encrypting. I have not checked what actually happens (is it empty? scrambled? something else?)

rishi-jat:
> > * StorageFile: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/196037#step/StorageFile/9) (error)
> >   `subprocess.CalledProcessError: Command 'echo test123 | cat - /dev/z...`
> > * StorageLVM: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/196037#step/StorageLVM/9) (error)
> >   `subprocess.CalledProcessError: Command 'echo test123 | cat - /dev/z...`
> > * StorageReflinkOnBtrfs: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/196037#step/StorageReflinkOnBtrfs/9) (error)
> >   `subprocess.CalledProcessError: Command 'echo test123 | cat - /dev/z...`
> > * StorageReflinkOnExt4: [test_010_persistent_luks](https://openqa.qubes-os.org/tests/196037#step/StorageReflinkOnExt4/9) (error)
> >   `subprocess.CalledProcessError: Command 'echo test123 | cat - /dev/z...`
> 
> Now the test says data doesn't survive encrypting. I have not checked what actually happens (is it empty? scrambled? something else?)
I investigated the latest test_010_persistent_luks failure.

@marmarek , I want to confirm the expected device size after in-place LUKS2 re-encryption.

If the original VM volume is 128 MiB and we temporarily grow it by 32 MiB for --reduce-device-size 32M, after re-encryption completes, what size should cryptsetup open expose to the guest?

Should it be:
128 MiB, preserving the original VM disk size, or
144 MiB, because the LUKS2 data segment starts 16 MiB after the beginning of the backing device?

If the expected guest size is 128 MiB, should we shrink the unused tail after re-encryption, or is there a preferred way to make cryptsetup open expose exactly the original 128 MiB?

I’m asking because OpenQA compares the whole /dev/xvde against the original 128 MiB image, so exposing 144 MiB causes the persistence test to fail even though the original data itself is intact.

marmarek:
I'd say encryption should be transparent for the VM, meaning the size should not change. So, 128M in this example.
IIUC extending by 32M (instead of 16M) is to have extra margin in case the header will need more space in the future. This should be fine. But then, IMO data should start at 32M offset. Looks like there is `--offset` option for that.

marmarek:
openQArun MACHINE=64bit TEST=system_tests_basic_vm_qrexec_gui,system_tests_pvgrub_salt_storage,system_tests_backup

rishi-jat:
@marmarek  Latest OpenQA run is green. test_010_persistent_luks passes on all four storage backends, and the other requested tests pass as well.

marmarek:
Some further test results:
1. Encrypting a volume of a qube that was never started does luks format on its private volume - so far so good. But then, the qube sees random content of its "fresh" private volume, and [refuses to create a filesystem there](https://github.com/QubesOS/qubes-core-agent-linux/blob/main/init/setup-rwdev.sh#L15-L20). This basically means startup failure. Maybe it should at least zero initial space after luks format?
2. Encrypting a volume of a qube that was started before is a bit complicated procedure, because it requires no revisions stored:
    1. Set revisions_to_keep to 0
    2. Start and stop the qube to actually remove past revisions
    3. Set passphrase and then encrypted=true
    4. Set revisions_to_keep back to the desired value
    
    Can this be improved? While I see why requiring no past revisions is there, maybe there is some better way to deal with this? For example, if `revisions_to_keep` is already 0, then remove past revisions before encrypting instead of refusing the operation? Another approach could be removing revisions when encrypting in any case, but that may be an unexpected side effect, so probably not a good idea... Or maybe encrypt past revisions too?

3. I noticed SetPassphrase survives qube restart when encrypted=false - is that intentional?
4. There is no `allow_discards` option set. It is fine for ephemeral volumes that are discarded on qube shutdown anyway, but for a long term storage this may be a problem - now, if you remove data within a qube, it won't release space in the underlying pool anymore. On the other hand, `allow_discards` may leak some information about the data structure inside (at least how much data you have - which is kinda intentional). FWIW, LUKS setup in dom0 has the flag enabled by default (not only in Qubes OS, it's a default in many distributions).

PR #872, opened 2026-08-18
Filter backup info

### LINKED ISSUE #11062: `admin.backup.Info` leaks all qube names without filtering for permission
### Qubes OS release

R4.3

### Brief summary

When attempting to call `qvm-backup --profile PROF` from an interface qube, it didn't work and there was a notification that `admin.backup.Info` was denied. I allowed the call, but the returned data shows in the unselected section, all qube names from the system that are not selected. An interface qube should not be able to know about them. It might be fine for them to know about the ones in the `include` section, for usability purposes, which includes "names, if is running, size".

Discussed with Marek that omitting information from the `include` section would not be wise, if the call is allowed, the caller wants to know about it, and lying about qube running state and size might cause issues.

Another layer of protection would be to deal with a denied call to `admin.backup.Info` as something that can happen, but still proceed to the update. This would allow a privileged qube to execute the backup without knowing it's contents or passphrase (it is in the profile in dom0), and still be able to track the `backup-progress`, and maybe, or maybe not, `admin.backup.Cancel` if the policy allows. One idea would be to treat the exception of "Request refused" as some specific exception, instead of `QubesDaemonAccessError`, by checking if the returncode is 126 in `qubesadmin.app.QubesRemote.qubesd_call`.

The backup summary contains `global` information, the scope of `admin.backup.Info` needs to change accordingly.

### Steps to reproduce

1. Create `sys-gui-vnc` following the [guide](https://doc.qubes-os.org/en/r4.3/user/advanced-topics/gui-domain.html#vnc-gui-domain-sys-gui-vnc)
2. Create a minimal backup profile in `dom0`
3. Attempt to backup from `sys-gui-vnc`
4. Notice no backup can be made until allowing the call `admin.backup.Info`

### Expected behavior

1. Backup works even if `admin.backup.Info` is denied
2. When `admin.backup.Info` is allowed, still don't include unknown qubes (unknown as in not allowed to be seen by the caller) from the unselected list
3. API registers `admin.backup.Info` as `scope="global"`

### Actual behavior

1. Backup can't be made until `admin.backup.Info` is allowed
2. Backup summary leaks all qube names in the unselection section
3. API registers `admin.backup.Info` as `scope="local"`

### Additional information

_No response_

ben-grande:
Non-dom0 should not know more information than necessary. If the call to `admin.backup.Info` is allowed, it is okay to show qubes in the `include` section, but it is not necessary to show `exclude`d qubes that caller is not aware of.

For: https://github.com/qubesos/qubes-issues/issues/11062

### DIFF
--- a/qubes/api/admin.py
+++ b/qubes/api/admin.py
@@ -2488,11 +2488,10 @@ class QubesAdminAPI(qubes.api.AbstractQubesAPI):
             "/" not in self.arg,
             reason="Backup profile ID must not contain forward slash character",
         )
-
-        self.fire_event_for_permission()
-
+        domains = self.fire_event_for_filter(self.app.domains)
         backup = await self._load_backup_profile(self.arg, skip_passphrase=True)
-        return backup.get_backup_summary()
+        summary = backup.get_backup_summary(filtered_domains=domains)
+        return summary
 
     def _send_stats_single(self, info_time, info, only_vm, filters):
         """A single iteration of sending VM stats
--- a/qubes/api/admin.py
+++ b/qubes/api/admin.py
@@ -2480,7 +2480,7 @@ class QubesAdminAPI(qubes.api.AbstractQubesAPI):
         wants_arg=True,
         wants_payload=False,
         dest_adminvm=True,
-        scope="local",
+        scope="global",
         read=True,
     )
     async def backup_info(self):
--- a/qubes/ext/admin.py
+++ b/qubes/ext/admin.py
@@ -88,7 +88,10 @@ class AdminExtension(qubes.ext.Extension):
 
     # TODO create that tag here (need to figure out how to pass mgmtvm name)
 
-    @qubes.ext.handler("admin-permission:admin.vm.List")
+    @qubes.ext.handler(
+        "admin-permission:admin.vm.List",
+        "admin-permission:admin.backup.Info",
+    )
     def admin_vm_list(self, vm, event, arg, **kwargs):
         """When called with target 'dom0' (aka "get full list"), exclude domains
         that the caller don't have permission to list
--- a/qubes/backup.py
+++ b/qubes/backup.py
@@ -482,7 +482,7 @@ class Backup:
         )
         return files_to_backup
 
-    def get_backup_summary(self):
+    def get_backup_summary(self, filtered_domains):
         summary = ""
 
         fields_to_display = [
@@ -568,14 +568,13 @@ class Backup:
         summary += "\n"
 
         vms_not_for_backup = [
-            vm.name for vm in self.app.domains if vm not in self.vms_for_backup
+            vm.name for vm in filtered_domains if vm not in self.vms_for_backup
         ]
         summary += (
             "VMs not selected for backup:\n - "
             + "\n - ".join(sorted(vms_not_for_backup))
             + "\n"
         )
-
         return summary
 
     async def _prepare_backup_header(self):


ben-grande:
PipelineRetryFailed

ben-grande:
PipelineRetry

qubes/api/admin.py
```diff
@@ -2492,7 +2492,9 @@ async def backup_info(self):
         self.fire_event_for_permission()
 
         backup = await self._load_backup_profile(self.arg, skip_passphrase=True)
-        return backup.get_backup_summary()
+        domains = self.fire_event_for_filter(self.app.domains)
```
marmarek, line 2491: This fires admin-permission event the second time - see fire_event_for_permission above.

marmarek (changes requested):

marmarek:
PipelineRetry

qubes/ext/admin.py
```diff
@@ -105,10 +108,11 @@ def admin_vm_list(self, vm, event, arg, **kwargs):
 
         policy = self.policy_cache.get_policy()
         system_info = qubes.api.internal.SystemInfoCache.get_system_info(vm.app)
+        method = event.split(":", 1)[1]
 
         def filter_vms(dest_vm):
             request = parser.Request(
-                "admin.vm.List",
+                method,
```
marmarek: Hm, I'd say it should remain `admin.vm.List`. Policy for `admin.backup.Info` controls which backup profiles one can use, not which VMs are included - so, at the very least, the argument in the policy wouldn't match.

marmarek (changes requested):

qubes/ext/admin.py
ben-grande: Tested the current patch and your proposal to see how it behaves.

With the current patch, excluded qubes that can be seen by the interface qube are listed in the summary, despite them not being able to call `admin.backup.Info` due to the Qrexec policy not allowing it. It is also possible to exclude qube from being listed in the exclude section:

```
VMs not selected for backup:
- dom0
- sys-gui-vnc
- untrusted
```

```
qvm-prefs work guivm sys-gui-vnc
qvm-prefs untrusted guivm sys-gui-vnc
```
```
admin.backup.Info +test sys-gui-vnc dom0 allow
admin.backup.Info +test sys-gui-vnc untrusted deny
```

```
VMs not selected for backup:
- dom0
- sys-gui-vnc
```

While for a qube that it's `guivm` is `dom0`:
```
admin.backup.Info +test personal dom0 allow
```
```
VMs not selected for backup:
- dom0
```

And if I allow it to target any qube, it shows all qubes in the system:
```
admin.backup.Info +test personal * allow
```
```
VMs not selected for backup:
- dom0
- A
- B
- C
- All qubes in the system
```

Now if I change to target the method `admin.vm.List`, `personal` doesn't list any other qubes in the "not selected section" (including not listing `dom0`, while `sys-gui-vnc` just lists all the qubes it knows about, despite the rule to deny `untrusted` to be listed.

Do you think that excluding qubes from the "not selected" section using `admin.backup.Info` is valuable?

marmarek: > Do you think that excluding qubes from the "not selected" section using `admin.backup.Info` is valuable?

tl;dr: no

The idea of this list is to make it easier to notice qube that you meant to include in the backup, but got excluded. You can do that only if you know such qube exists. So, including only qubes that are allowed to be listed is okay, and even better - otherwise you risk forgetting to setup policy to include listing them also by `admin.backup.Info` and as such defeating the purpose of that list.

marmarek (approved):

PR #876, opened 2026-08-25
Updating behaviour on concurrent access to a pending VM shutdown oper…

### LINKED ISSUE #11076: `qvm-shutdown --wait` succeeds too soon if VM shutdown was already in progress and volumes were stopping
### Qubes OS release

R4.3

### Brief summary

If `qvm-shutdown --wait` is handling a VM that was already in the process of shutting down, its behavior depends on how far that ongoing shutdown had progressed:

- If the VM was still live and going through its in-VM shutdown procedure, `qvm-shutdown --wait` correctly waits for the storage volumes to finish stopping before succeeding.

- But if the VM had already halted and dom0 had begun `storage.stop()` for the VM without finishing it yet (which can take many seconds for highly fragmented volumes), `qvm-shutdown --wait` succeeds immediately, long before the storage volumes have finished stopping.

### Steps to reproduce

1. Add a 30 second delay at the beginning of [`Storage.stop()`](https://github.com/QubesOS/qubes-core-admin/blob/v4.3.50/qubes/storage/__init__.py#L859-L869) in `/usr/lib/python*/site-packages/qubes/storage/__init__.py` in dom0:
 
       self.vm.log.info("Storage.stop(): sleeping")
       await asyncio.sleep(30)
       self.vm.log.info("Storage.stop(): proceeding")

2. Run `systemctl restart qubesd` (which will also be delayed by 30 seconds).

3. Start any VM, keep a dom0 terminal visible running `journalctl -fu qubesd`, and type `qvm-shutdown --wait VMNAME` (without pressing Enter) into another dom0 terminal.

5. Shut down the VM. As soon as `Storage.stop(): sleeping` appears in the journal, press Enter in the second dom0 terminal.

### Expected behavior

`qvm-shutdown --wait` should take at least ~30 seconds to succeed.

### Actual behavior

It immediately exits successfully.

### Additional information

This might occasionally cause problems on system shutdown, because `qubes-core.service` can finish stopping prematurely (allowing the system shutdown to proceed) while some VMs were still in the process of having their volumes stopped:

https://github.com/QubesOS/qubes-core-admin/blob/v4.3.50/linux/systemd/qubes-core.service#L13

ekdama:
Closes QubesOS/qubes-issues#11076

Now, the behaviours that could make a concurrent call (like the Halted check), waiting for a shutdown of the VM through the `qvm-shutdown --wait` command will also test if there is a waiter present in the VM.
The waiter job is to act as a synchronization primitive, when it is finally consumed all the concurrent shutdown request will exit, if the wait flag is enabled.

Is the event firing handled correctly? I only throw the pre-shutdown event when there is not a pending operation.


qubes/vm/qubesvm.py
```diff
@@ -1668,14 +1668,15 @@ async def shutdown(self, force=False, wait=False, timeout=None):
         :raises qubes.exc.QubesVMNotStartedError: \
             when domain is already shut down.
         """
-
-        if self.is_halted():
+        pending = self.is_halted() and not self.__waiter is None
```
marmarek, line 1676: This could use a comment explaining what case it handles (in-progress shutdown that is already at the storage stopping stage). Or if not a comment, at least include it in the commit message.

marmarek (changes requested):
Looks like `kill()` could use a similar change (if you're up for it)

marmarek:
> Is the event firing handled correctly? I only throw the pre-shutdown event when there is not a pending operation.

Yes, it makes sense.

ben-grande:
I think `domain-shutdown` is fired only once. Could you write an integration test that register `domain-shutdown` event handler and check how many times it is called? It should probably be in `qubes/tests/integ/basic.py`.

ekdama:
Sorry, I had some  personal issues this week and had to be AFK. I believe that I can get the requested changes done in the following days.

qubes/vm/qubesvm.py
```diff
@@ -1668,14 +1668,15 @@ async def shutdown(self, force=False, wait=False, timeout=None):
...
+        if self.is_halted() and not pending:
             raise qubes.exc.QubesVMNotStartedError(self)
 
         try:
-            await self.fire_event_async(
```
ben-grande, line 1676: What happens if a second shutdown is called on `domain-pre-shutdown`? Can you write a test that does that? I'd like to test if `pending` on the second shutdown request will be `False`, in case domain is not halted yet.

marmarek:
IMO `domain-pre-shutdown` can be fired multiple times, if qube is still shutting down (this allows retrying if it ignored the first request for some reason). But when it's already cleaning up (storage stop etc), `domain-pre-shutdown` should not be fired anymore.
As for `domain-shutdown`, it should be fired once per actually completed shutdown, regardless how many calls triggered that shutdown.

I think that's the current behavior, but some test for it could be useful.

ekdama:
With the pending condition, the integration test looped firing domain-pre-shutdown events.
I fixed it by checking the __waiter as the first thing in the try block. 

About kill, how should I handle it? I made these small changes, but I don't see a need since it does not have a wait option.

qubes/vm/qubesvm.py
```diff
@@ -1669,25 +1669,32 @@ async def shutdown(self, force=False, wait=False, timeout=None):
...
-                "domain-pre-shutdown", pre_event=True, force=force
-            )
+            """The best option is to fire the domain-pre-shutdown after self.__waiter check and assignment, to avoid race conditions and undesired synchronization behaviour."""
+            if self.__waiter is None:
+                self.__waiter = asyncio.get_running_loop().create_future()
+                await self.fire_event_async(
```
ben-grande: If the `domain-pre-shutdown` fails (is rejected), the `self.__waiter` will continue to be `None`

tests-data/sysfs/sys/bus/pci/devices/0000:00:08.1/secondary_bus_number
```diff
@@ -0,0 +1 @@
+2
```
ben-grande: Some files included in `tests-data` by mistake?

ekdama: Yes, I am correcting this and linting mistakes as of now. This also broke some test. I will eventually squash all commits.

qubes/vm/qubesvm.py
ben-grande, line 1734: One side-effect of calling this function again, is that it doesn't connect to the previous timeout, thus if it is hangs here,  it will timeout later, possibly raising timeout twice.

qubes/vm/qubesvm.py
ekdama: Can I clean `self.__waiter` in the `except:` block? This would avoid creating a `__waiter `in an unsuccesful operation.

Or else would be better creating a try except block wrapping 1681 to 1684?

ben-grande: https://github.com/QubesOS/qubes-core-admin/pull/876#issuecomment-5509660860

Let's actually allow `domain-pre-shutdown` to repeat depending on the stage of shutdown.

ekdama: Weakening the `pending` condition by just checking `self.__waiter` could make `domain-pre-shutdown` repeatable. Could it work?

qubes/vm/qubesvm.py
ekdama: Let's check pending also there?

ben-grande: >             elif not pending:
>            elif is_running:

We actually want to wait for the full shutdown, independent if it is report as running or it is the second call.

ekdama: Yes sorry, I was confused. `is_running` and `not pending` are not aimed to the timeout call but to avoid calling `libvirt_domain.destroy()` or other domain specific code more than once, as I believe is the expected behaviour. 

Then the waiter is acquired, not concerned by the current VM state (except the initial `is_halted` check).

PR #877, opened 2026-08-26
perf(qubes-hcl-report): cache the yaml output until the next reboot

andreasglashauser:
Collecting the data shells out to lspci, dmidecode, xl info and xl dmesg. Callers that only want the yaml output pay for that on every run, for example Qubes Global Config and `tests/dispvm_perf.py`.

  Everything in the yaml output is fixed for the lifetime of a boot, so it is now cached under the user's cache directory keyed on `/proc/sys/kernel/random/boot_id`.

  Only `--yaml-only` is cached. The support files bundle raw `lspci -nnvk` output, which does change within a boot, eg when a PCI device is attached to a qube and bound to pciback. A report collected with an incomplete `xl dmesg` ring buffer is not cached either, since its virtualization fields come out empty. `--no-cache` collects the data again.

QubesOS/qubes-issues#11081

### DIFF
--- a/qvm-tools/qubes-hcl-report
+++ b/qvm-tools/qubes-hcl-report
@@ -115,7 +115,7 @@ if [[ "$YAML_ONLY" = 1 ]]
     fi
 fi
 
-BOOT_ID=$(cat "$BOOT_ID_FILE" 2>/dev/null)
+BOOT_ID=$(cat "$BOOT_ID_FILE")
 
 # Everything in the yaml output is fixed for the lifetime of a boot: the
 # hardware, the firmware version (a flash needs a reboot to take effect), the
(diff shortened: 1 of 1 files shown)


qvm-tools/qubes-hcl-report
```diff
@@ -412,6 +456,29 @@ versions:
 
 if [[ "$YAML_ONLY" == 1 ]]
   then
+    # Do not cache a report with missing data: an incomplete "xl dmesg" ring
+    # buffer empties the virtualization fields, and that would then be served
+    # for the rest of the boot.
+    # a cache write failure must not disturb the output, so errors here are
+    # discarded rather than reported
+    if [[ "$USE_CACHE" = 1 && -n "$BOOT_ID" && "$XL_DMESG_INCOMPLETE" = no ]] &&
+       mkdir -p "$CACHE_DIR" 2>/dev/null
```
marmarek: Regarding "must not disturb the output" - it may make sense for stdout, but stderr is for error reporting and better not hide it (here and in the few other places below).

marmarek (changes requested):

qvm-tools/qubes-hcl-report
```diff
@@ -126,8 +126,10 @@ BOOT_ID=$(cat "$BOOT_ID_FILE" 2>/dev/null)
 # pciback.
 if [[ "$YAML_ONLY" = 1 && "$USE_CACHE" = 1 && -n "$BOOT_ID" ]]
   then
-    if [[ "$(cat "$CACHE_BOOT_ID_FILE" 2>/dev/null)" = "$BOOT_ID" ]] &&
-       [[ -s "$CACHE_FILE" ]]
+    # tested before reading, so that having no cache yet does not report an
```
andreasglashauser, line 129: I added this comment because it was not (immediately) clear for me.

qvm-tools/qubes-hcl-report
andreasglashauser: Fixed.

marmarek (approved):

PR #879, opened 2026-08-26
Revert "Cleanup disposable on failed startup"

marmarek:
This introduces an issue with named disposables being removed - worse
than the original issue. Proper fix turned out to be more complicated,
so revert offending commit for now.

This reverts commit a82f4f2affcf083f7e7ac24414a7464f00d08ed0.

### DIFF
--- a/qubes/vm/dispvm.py
+++ b/qubes/vm/dispvm.py
@@ -1011,7 +1011,7 @@ class DispVM(qubes.vm.qubesvm.QubesVM):
             await super().start(**kwargs)
         except:
             # Cleanup also on failed startup
-            await self.cleanup(force=True)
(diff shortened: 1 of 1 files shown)


marmarek:
> system_tests_dispvm
> 
>     * [unstable] TC_21_DispVM_Preload: [test_012_preload_low_mem_early_startup](https://openqa.qubes-os.org/tests/194296#step/TC_21_DispVM_Preload/3) (failure)
>       `~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^... AssertionError: 0 != 2`

This is about the re-introduced bug that was fixed by the reverted commit. Will need to live with that for now.

PR #880, opened 2026-08-27
qvm-console-dispvm: clarify error when management DispVM has no guivm

### LINKED ISSUE #10894: Provide a better error message when using qvm-console-dispvm with a disposable template where guivm is None
### Qubes OS release

4.3

### Brief summary

If you set the `guivm` property of `default-mgmt-dvm` to `None`, `qvm-console-dispvm` fails to start.

### Steps to reproduce

In dom0:

```console
qvm-prefs default-mgmt-dvm guivm None
qvm-console-dispvm sys-usb
xterm: Xt error: Can't open display: :0
```

### Expected behavior

Provide a better error message, like: `Error: the qube 'default-mgmt-dvm' needs to have a non-empty guivm property`

### Actual behavior

There is something about the display, but it is not clear what's wrong.

### Additional information

Very low priority and good first issue?

mickeyslaven:
qvm-console-dispvm starts qubes.ShowInTerminal without checking whether the selected management DispVM has a GUI domain. This leaves users with an unclear xterm display error.

The fix checks the management DispVM's guivm property first and reports a direct error when it is empty.

Fixes QubesOS/qubes-issues#10894

AI DISCLOSURE: Generative AI was utilized in testing, writing, and assisting in review of this code. I signed and personally reviewed the code myself and take responsibility for its contents.

### DIFF
--- a/qvm-tools/qvm-console-dispvm
+++ b/qvm-tools/qvm-console-dispvm
@@ -42,4 +42,11 @@ DISPVM="$(qvm-prefs -- "$QREXEC_REQUESTED_TARGET" management_dispvm)"
 
 [[ "x$DISPVM" == "x" ]] && { echo "Error: cannot determine management DispVM of domain '$QREXEC_REQUESTED_TARGET'">&2; exit 1; }
 
+GUIVM="$(qvm-prefs -- "$DISPVM" guivm)"
+
+if [[ -z "$GUIVM" ]]; then
+    echo "Error: management DispVM '$DISPVM' does not have a GUI domain" >&2
+    exit 1
+fi
+
 sudo qvm-run -p --localcmd="QREXEC_REQUESTED_TARGET=$QREXEC_REQUESTED_TARGET QREXEC_REMOTE_DOMAIN=dom0 /etc/qubes-rpc/admin.vm.Console" --service --dispvm="$DISPVM" -- qubes.ShowInTerminal


qvm-tools/qvm-console-dispvm
```diff
@@ -42,4 +42,8 @@ DISPVM="$(qvm-prefs -- "$QREXEC_REQUESTED_TARGET" management_dispvm)"
 
 [[ "x$DISPVM" == "x" ]] && { echo "Error: cannot determine management DispVM of domain '$QREXEC_REQUESTED_TARGET'">&2; exit 1; }
 
+GUIVM="$(qvm-prefs -- "$DISPVM" guivm)"
+
+[[ -z "$GUIVM" ]] && { echo "Error: management DispVM '$DISPVM' does not have a GUI domain">&2; exit 1; }
```
marmarek: This script has `set -e`, false condition here will not only skip echo, but will interrupt the script. Use proper `if` instead.

marmarek (changes requested):

qvm-tools/qvm-console-dispvm
mickeyslaven: Thats understood, the fix is now resubmitted for review

marmarek:
Can you squash the two commits together?

mickeyslaven:
Squashed into one signed commit

marmarek (approved):

PR #887, opened 2026-09-10
Make sure color value is stripped when creating label

### LINKED ISSUE #11116: API allowed me to create a label with an Enter in the color

### Qubes OS release

r4.3

### Brief summary

I used external script (https://github.com/alimirjamali/personal-qubesos/blob/main/qubes-label-tt/qubes-label-tt ) to add some more labels. Something dreadful happened, and now my qubes.xml has demons.

### Expected behavior

Create Label API call should gently tap me on the head with rolled-up newspaper and not allow me to create abominations.

### Actual behavior

```
<qubes>
<labels>
...
<label id=label-9 color="0xf472b6&#10;">ux-pink</label>
```

### Additional information

I know this is my fault, but still, I don't think I should be allowed to shoot myself like this. It also makes label coloring go crazy.... but not window frame coloring.

marmarta:
Currently validating function checks stripped value for correctness but then just discards it. Let's use a stripped value to avoid stray characters.
Also removed reference to a ticket that is no longer relevant in the context.

content notice: debugging was assisted by qwen 3.8 large language model; no code in this PR was generated by an llm

fixes QubesOS/qubes-issues#11116


qubes/utils.py
```diff
@@ -538,7 +527,7 @@ def validate_label_name(untrusted_label, creation: bool = False) -> None:
         )
 
 
-def validate_label_value(untrusted_label_value) -> None:
+def validate_label_value(untrusted_label_value) -> str:
```
ben-grande, line 530: Maybe there shouldn't be a `.strip()` below, other checks below would detect that `\n` is not hexdigit, and the length would not match also.

ben-grande (commented):
I see now, on the issue, I did the debugging steps on R4.3 instead of R4.4, cause I thought you were at that version.

I don't think the validation should be responsible for providing the value in most cases, but on this specific function, it is sanitization (decode to ASCII) plus validation, so I think it is fine.

qubes/utils.py
ben-grande: And please provide a test for a color value ending with a newline char.

PR #889, opened 2026-09-11
tests: fix setting xterm title

marmarek:
Disable dynamic titles, so Whonix doesn't override the title back to
user@host.


marmarek:
openQArun TEST=system_tests_basic_vm_qrexec_gui MACHINE=64bit

PR #890, opened 2026-09-11
WIP: Add more colors

marmarta:
Also move existing colors around for better distinctiveness.

Colors:
overview-full


qubes/app.py
```diff
@@ -1711,14 +1411,20 @@ def _release_lock(self):
...
             5: qubes.Label(5, "0x555555", "gray"),
-            6: qubes.Label(6, "0x3465a4", "blue"),
-            7: qubes.Label(7, "0x75507b", "purple"),
+            6: qubes.Label(6, "0x60a5fa", "blue"),
+            7: qubes.Label(7, "0x7e22ce", "purple"),
             8: qubes.Label(8, "0x000000", "black"),
+            9: qubes.Label(9, "0xf472b6", "pink"),
+            10: qubes.Label(10, "0x4338ca", "indigo"),
+            11: qubes.Label(11, "0x2dd4bf", "cyan"),
+            12: qubes.Label(12, "0xbef264", "lime"),
```
ben-grande, line 1425: Can similar colors be together, such as lime and green? I think this would allow client applications to list them close to each other, such as label selection on qube creation. The downside is lacking backwards compatibility, would have to migrate label indexes from the store, based on some criteria to decide when that migration should happen.

marmarta: This would move existing labels, which might annoy some people. It would also influence how labels are sorted in the manager and in dropdowns.

marmarta:
This needs code to migrate existing labels and (worse) migrate user's custom labels to numbers below these. 
Thankfully looks like qubesd uses names not numbers to refer to labels, so we are not completely SOL.

PR #892, opened 2026-09-16
Improve storage_perf.py

crass:
This refactors `storage_perf.py` and add the `--device` option for testing a given block device, either in dom0 or in the VM. There is an optional `--mount` option to mount and do the test on the mount of the block device.

This is backwards compatible with the previous usage, so existing usages should see no change. The one except is in the creation of the name prefix where the '_' separator was replaced with ':' to improve parseability. I doubt anything has been parsing that though.


marmarek:
> I doubt anything has been parsing that though.

There is a bot posting results on github:
https://github.com/QubesOS/openqa-tests-qubesos/blob/main/utils/github_reporting.py#L177
https://github.com/QubesOS/openqa-tests-qubesos/blob/main/utils/lib/openqa_api.py#L334

But looks like it should be okay with that separator change.

tests/storage_perf.py
```diff
@@ -20,11 +20,12 @@
 import argparse
 import dataclasses
 import os
+import sys
 import subprocess
 import tempfile
 
 import qubesadmin
-
+from qubesadmin.tools import qvm_device
```
marmarek, line 28: Please don't, there is absolutely no guarantee _internal_ API of tools won't change.
You can either call it via subprocess, or call underlying `self.vm.devices["block"].attach()` (or whatever you need).

marmarek (changes requested):

